PR #1, opened 2019-01-21
module level fixture added

koaning:
this is the first branch, i am checking if travis works.


koaning:
ah everything is green. that means that i can now, and only now, merge.

Issue #12, opened 2019-02-27
feature request: dtype selector


kayhoogland:
Selecting Pandas columns by their dtype

koaning:
Could you give an example of what the resulting pipeline might look like?

kayhoogland:
I suppose a transformer such as this one is what I am thinking of. 

```
import warnings
from sklearn.base import BaseEstimator, TransformerMixin

class PandasTypeSelector(BaseEstimator, TransformerMixin):
    def __init__(self, dtype):
        self.dtype = dtype

    def fit(self, X, y=None):
        # Save the column names for later checking
        self.type_columns = list(X.select_dtypes(include=[self.dtype]))
        return self

    def transform(self, X):
        assert isinstance(X, pd.DataFrame)

        transformed_frame = X.select_dtypes(include=[self.dtype])

        if set(list(transformed_frame)) != set(self.type_columns):
            warnings.warn(f'Columns of type {self.dtype} were not equal during fit and transform')

        return transformed_frame
```

koaning:
A `ValueError` might be better than a mere assert but this looks good to me. Any idea on how you might want to test this?

kayhoogland:
- Good point, I agree!
- I believe you mean this: https://goo.gl/ZAdbv7. At what point would like to determine the dtype if it will not be in the __init__?
- For testing, a simple test with a small sample dataframe will be enough I assume. We do not have to test the _.select_dtypes_ function of Pandas.

koaning:
There's some basic tests in there that check things like "a transformer should not change the number of rows" and you might consider adding a test for "a transformer should not change the order of the input/output".

PR #44, opened 2019-03-20
PandasTypeSelector for selecting pandas columns based on dtypes

kayhoogland:
Added PandasTypeSelector to select columns in a pandas dataframe based on their dtype.


sklego/preprocessing.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class PandasTypeSelector(BaseEstimator, TransformerMixin):
+    """
+    Select columns in a pandas dataframe based on their dtype
+
+    :param include: types to be included in the dataframe
+    :param exclude: types to be exluded in the dataframe
+    """
+    def __init__(self, include, exclude=[]):
```
koaning: could you turn the empty list into something that is not mutable? 

also, should `include` also not be optional?

sklego/preprocessing.py
```diff
@@ -0,0 +1,48 @@
...
+        self.exclude = exclude
+
+    def fit(self, X, y=None):
+        """
+        Saves the column names for check during transform
+
+        :param X: pandas dataframe to select dtypes for
+        :param y: not used in this class
+        """
+        self._check_X_for_type(X)
```
koaning: maybe an explicit `ValueError` when you select columns that do not appear in the dataframe that is brought in here? 

you probably also want to do this during the `.transform()` phase

kayhoogland: Yes I changed it

sklego/preprocessing.py
kayhoogland: Good point, changed the defaults to None

sklego/preprocessing.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    def __init__(self, include=None, exclude=None):
+        self.include = include
+        self.exclude = exclude
+
+    def fit(self, X, y=None):
+        """
+        Saves the column names for check during transform
+
+        :param X: pandas dataframe to select dtypes for
```
koaning: for -> out of

tests/test_transformers/test_pandastypeselector.py
```diff
@@ -0,0 +1,66 @@
...
+
+    with pytest.raises(ValueError):
+        transformer.fit(X, y)
+
+
+def test_value_error_inequal(random_xy_dataset_regr):
+    transformer = PandasTypeSelector(include=['number'])
+    X, y = random_xy_dataset_regr
+    X = pd.DataFrame(X)
+    if len(list(X)) > 0:
```
koaning: use `X.shape[0]` -> this is always more performant and direct.

sklego/preprocessing.py
kayhoogland: Check, changed

tests/test_transformers/test_pandastypeselector.py
kayhoogland: Agreed, changed in last commit

kayhoogland:
https://github.com/koaning/scikit-lego/issues/12

PR #50, opened 2019-03-20
Added ColumnSelector to preprocessor with doctests and examples

sandervandorsten:


koaning:
could you add a description of what you did?

.gitignore
```diff
@@ -72,6 +72,7 @@ target/
 
 # Jupyter Notebook
 .ipynb_checkpoints
+notebooks/*.ipynb
```
koaning: we might actually *want* to have notebooks here. if you don't want to add yours specifically it is better to not add it in git.

sandervandorsten:
created a clean branch where only the code is added (see [branchname]-reworked

Issue #23, opened 2019-03-05
feature request: ensemble model that takes p(class | model) into account


koaning:
It might be nice to have an variant of the [voting classifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.VotingClassifier.html). One that goes a bit further than mere voting but takes the uncertainty of seperate classifiers into account.

koaning:
700 downloads cant be wrong. 

https://link.springer.com/chapter/10.1007/978-3-319-00651-2_19

RensDimmendaal:
@koaning : How does this proposal differ from setting voting="soft" in [sklearn.ensemble.VotingClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.VotingClassifier.html)?

koaning:
I'll make it slightly clearer what I am thinking about here. 

Consider that we have an ensemble of three models. They each have a confusion matrix and at the time of inference they each come with a prediction for class `B`.

![image](https://user-images.githubusercontent.com/1019791/74513307-8e117f00-4f0a-11ea-8cc9-253fceee5f1e.png)

Do we really trust what model A has to say here? When it predicts class `B` then it is often wrong. 

![image](https://user-images.githubusercontent.com/1019791/74513376-b4371f00-4f0a-11ea-84f0-28247830e9f1.png)

So maybe ... we need a different way of reweighting here. An idea I've had for a while is to try to reweigh the weights using the confusion matrix. There is even something bayesian about it. 

![image](https://user-images.githubusercontent.com/1019791/74513430-d03ac080-4f0a-11ea-97c4-5a3544a34e2c.png)

PR #52, opened 2019-03-20
Implementation of EnsembleModel

anachal:
Vincent will probably tell me that it's spelled "EnsembleModel"


anachal:
Is it good?

notebooks/EnsembleModel.ipynb
```diff
@@ -0,0 +1,207 @@
+{
```
koaning: please remove this notebook for now.

sklego/ensemble_model.py
```diff
@@ -0,0 +1,87 @@
+from sklearn.linear_model import LogisticRegression
+from sklearn.metrics import confusion_matrix
+
+class EnsembleModel():
+    """
+    A classifier that takes the uncertainty p(class | model) of a classifier into account.
+    Based on the requirements of issue #23 of koaning/scikit-lego
+    """
+
+    def __init__(self, X_test, y_test, estimator=LogisticRegression(), alpha=0.5):
```
koaning: So ... 

... I think it might be a good idea to have a seperate validation set on which to base the confusion matrix. However this object needs to primarily fit a `sklearn.pipeline` so the base behavior should be to base the confusion matrix on the training set. 

I also think it might be nicer not to use `LogisticRegression()` as a base input and instead demand that the user defines this explicitly. Explicit > Implicit.

koaning: Also this might be a better order. 

```
def __init__(self, estimator=LogisticRegression(), alpha=0.5, X_test=None, y_test=None)
```

sklego/ensemble_model.py
```diff
@@ -0,0 +1,87 @@
...
+        self.y_test = y_test
+
+    def fit(self, X, y):
+        """
+
+        :param X: array of feature values used for training the classifier
+        :param y: array of target values used for training the classifier
+        :return: trained classifier (estimator)
+        """
+        self.fitted_estimator_ = self.estimator.fit(X, y)
```
koaning: perhaps a different name than `self.fitted_estimator_` maybe ... `self.orig_estimator` 

also you may want to calculate the confusion matrix in the `self.fit` as well

sklego/ensemble_model.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+
+        :param X: array of feature values upon which a prediction is to be made
+        :return: Updated prediction probabilities of each of the target value made by the trained classifier
+        """
+        p = self.predict_proba(X)
+        prob_matrix = self.prob_matrix()
+        return p @ prob_matrix
+
+    def make_prediction(self, X):
```
koaning: this needs to be called `.predict(X)` ... better yet ... implement just `predict_proba` and you get `.predict()` for free.

tests/test_estimators/test_ensemble_model.py
```diff
@@ -0,0 +1,33 @@
+from sklego.ensemble_model import EnsembleModel
+from sklearn import datasets
+from sklearn.model_selection import train_test_split
+
+# Data
```
koaning: this needs to be put into a fixture.

tests/test_estimators/test_ensemble_model.py
```diff
@@ -0,0 +1,33 @@
...
+iris = datasets.load_iris()
+X = iris.data[:, :2]  # we only take the first two features.
+y = iris.target
+X_train, X_test, y_train, y_test = train_test_split(X, y)
+
+# Model
+model = EnsembleModel(X_test = X_test, y_test = y_test)
+model.fit(X_train, y_train)
+
+def test_alpha():
```
koaning: better: test that if `alpha > 1` an error gets thrown.

tests/test_estimators/test_ensemble_model.py
```diff
@@ -0,0 +1,33 @@
...
+    assert model.alpha >= 0
+    assert model.alpha <= 1
+
+
+def test_predict_proba():
```
koaning: in this case there is a module in the testing framework where shapes and other common things are tested for **all** estimators. might be better to put this test there.

sklego/ensemble_model.py
MBrouns: `sklearn` expects that the parameters (`self.estimator` in this case) don't change in `fit` so it's probably best to clone the self.estimator (see for example the `EstimatorTransformer`)

```
 self.estimator_ = clone(self.estimator)
 self.estimator_.fit(X, y)
```

sklego/ensemble_model.py
```diff
@@ -0,0 +1,87 @@
...
+
+        :param X: array of feature values upon which a prediction is to be made
+        :return: prediction of the target value made by the trained classfifier
+        """
+        return self.fitted_estimator_.predict(X)
+
+    def predict_proba(self, X):
+        """
+
+        :param X: array of feature values upon which a prediction is to be made
```
MBrouns: You probably want to check if the estimator is fitted. You can use `check_is_fitted` from `sklearn.utils.validation` like this `check_is_fitted(self, 'estimator_')`

sklego/ensemble_model.py
```diff
@@ -0,0 +1,87 @@
+from sklearn.linear_model import LogisticRegression
+from sklearn.metrics import confusion_matrix
+
+class EnsembleModel():
+    """
+    A classifier that takes the uncertainty p(class | model) of a classifier into account.
+    Based on the requirements of issue #23 of koaning/scikit-lego
```
MBrouns: It might be nice to refer to the original paper here

koaning: It's not *that* great of a paper. But sure.

koaning:
Actually ... come to think of it ... this model is not an emsemble model as much. It's more of a postprocessor. @MBrouns what might be a better submodule? `postprocessor` or `meta`?

MBrouns:
Agreed that it's not an ensemble, more like a `CMReweigher`. To me it feels more like a `meta` than a postprocessor as it actually wraps the estimator

koaning:
This package is bound to have a lot of models in `meta`. Do you forsee any way to maybe split it up?

koaning:
Referencing the issue: https://github.com/koaning/scikit-lego/issues/23

sklego/ensemble_model.py
anachal: For some reason I couldn't manage to get `.predict()` for free, so I defined it. How can I get it like that?

sklego/ensemble_model.py
anachal: Where can I find the paper?

sklego/ensemble_model.py
MBrouns: I think you only get `predict` if you mixin `LinearClassifierMixin` but we probably don't want that in this case so implementing it yourself is fine

sklego/ensemble_model.py
MBrouns: https://link.springer.com/chapter/10.1007/978-3-319-00651-2_19

sklego/ensemble_model.py
```diff
@@ -0,0 +1,104 @@
...
+       :param y: array of target values used for training the classifier
+       :return:
+                - trained classifier (self.orig_estimator)
+                - confusion matrix populated with the probabilities of each real value vs model prediction value
+                  of the target variable (self.confusion_matrix)
+                - updated prediction probabilities of each target value made by the trained classifier
+                  (self.prob_matrix)
+       """
+        self.estimator_ = clone(self.estimator)
+        self.orig_estimator_ = self.estimator_.fit(X, y)
```
MBrouns: You now have `self.estimator_` and `self.orig_estimator_` referring to the same estimator, I would get rid of one of them

koaning:
In terms of a name convention. `EnsembleModel` might be a bit too vague and the added feature here is the addition of `ReweightedEnsemble`?

Issue #47, opened 2019-03-20
Bug: Sphinx not installed as dependency for development


sandervandorsten:
- create `dev` dependencies in `./setup.
- `pip install sphinx`
- `pip install sphinx_rtd_theme`

sandervandorsten:
also; install doctest and include `doctest` in `make docs`

koaning:
ill have a small go at this.

PR #55, opened 2019-03-20
Bug47 installsphinx

sandervandorsten:
Issue resolved: #47 

- See 'files changed' for all updates
- Not sure if the `make develop` command is correctly edited, could you check that?


Makefile
```diff
@@ -8,9 +8,13 @@ install:
 	pip install -e .
 
 develop:
-	python setup.py develop
+	# python setup.py develop
```
koaning: keep `python setup.py develop` at the end.

sandervandorsten:
@koaning resolved conversation, can safely be merged

PR #57, opened 2019-03-20
Added module to add lag columns to a pandas DataFrame.

Menziess:


sklego/pandas.py
```diff
@@ -0,0 +1,95 @@
...
+
+    if isinstance(cols, str):
+        return [cols]
+
+    if hasattr(cols, '__iter__'):
+        return cols
+
+
+def add_lag(X, cols, lags):
+    """
```
koaning: might be good to check if `X` is indeed a pandas dataframe or a numpy array before doing anything and returning an value error if this isnt the case. 

same holds for if the index or the column is missing. having a valuerror from our library is better than waiting for later packages. it's more friendly to uders

koaning: also. params!

koaning:
If you've made the required edits, make sure the original conversations are resolved with a confirmed comment. This way, I can see that you've taken a mental step towards fixing the issues. 

When there's a new commit I can merge it in.

sklego/pandas.py
```diff
@@ -0,0 +1,186 @@
+import numpy as np
+import pandas as pd
+from scipy.ndimage.interpolation import shift
+
+
+def _as_list(val):
```
MBrouns: If you do `_as_list(range(5))` you get a list where the first element is the generator, is that the behaviour you intend to have? If so, I would suggest updating the docstring as it now mentions that it would return the value as a list if it's not iterable, while in truth it returns the value as a list if not already a list

MBrouns: Does this function belong in the `pandas` module or would it make more sense in a utilities module?

koaning: It is a util for pandas. My attemp with the `.pandas` module was to have utils that only have to do with pandas (this might make sense since the rest of the project is for `sklearn`).

Menziess: """
    Helper function, always returns a list.

    :param val: a list or a single value.
    :returns: the input value as a list.

    :Example:

    >>> _as_list('test')
    ['test']

    >>> _as_list(['test1', 'test2'])
    ['test1', 'test2']
    """

sklego/pandas.py
```diff
@@ -0,0 +1,186 @@
...
+
+    # Check the X instance of pandas DataFrame
+    if isinstance(X, pd.core.frame.DataFrame):
+        return _add_lagged_pandas_columns(X, cols, lags)
+
+    # Check the X instance of numpy array
+    if isinstance(X, np.ndarray):
+        return _add_lagged_numpy_columns(X, cols, lags)
+
+    raise ValueError("Please pass a pandas DataFrame")
```
koaning: *OR* a numpy array

sklego/pandas.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def _add_lagged_numpy_columns(X, cols, lags):
```
koaning: if this is a hidden function to the user, why write an example for it?

sklego/pandas.py
koaning: the `as_list` method can actually be kind of useful but feels too general to be placed here. maybe move it to `common.py`. @MBrouns, thoughts?

koaning: might be nicer to make sure we handle tuples here as well.

sklego/pandas.py
```diff
@@ -0,0 +1,186 @@
...
+
+    lags = _as_list(lags)
+
+    if not all(isinstance(x, int) for x in lags):
+        raise ValueError("Must be list of ints")
+
+    return [-x for x in lags]
+
+
+def add_lag(X, cols, lags):
```
koaning: `add_lags` might be a better name.

sklego/pandas.py
Menziess: In my opinion, if you decide in the future to make it public, it's trivial. If a complex public function becomes private, would you remove the examples?

It also documents the intention for those who may have to debug it, and it works as a proof that everything is functioning correctly, as otherwise you'd have to rely on testing multiple cases in the function that calls the hidden function. For example:
- does numpy work with input x?
- does numpy work with input y?
- does numpy fail correctly with a wrong input?
- does pandas work with input x?
- does pandas work with input y?
- does pandas fail correctly with a wrong input?

Would you keep adding tests to 'add_lag' just to make sure the underlying hidden functions are working as intended?

sklego/pandas.py
MBrouns: Yes, I agree with moving it to common

sklego/pandas.py
MBrouns: Nothing wrong with writing tests for it but in this case I would indeed put them in a separate unit test rather than in doctests

Menziess:
The github editor doesn't remove abundant whitespace apparently.

sklego/pandas.py
Menziess: This will now be handled in a more generalized way.

Menziess:
`make check` fails with the error:
    AttributeError: module 'pandas' has no attribute 'core'

My gut tells me that there may be a namespace conflict because the module is named `pandas`.

sklego/pandas.py
Menziess: Okay, so only the description and params sections of the docstring of hidden functions should remain?

sklego/pandas.py
Menziess: I'll modify the function a bit to be more generally usable.

koaning:
urgh. @MBrouns maybe rename the submodule to `pandas_util`? might make sense too.

sklego/pandas.py
```diff
@@ -20,3 +25,126 @@ def wrapper(*args, **kwargs):
...
         return result_df
     return wrapper
+
+
+def add_lags(X, cols, lags):
+    """
+    Appends lag column(s).
+
+    :param X: array-like, shape=(n_columns, n_samples,) training data.
+    :param cols: an iterable of column names or a single column name string.
```
koaning: **or** indices representing columns in an `np.ndarray`

MBrouns:
Yep, especially when we get namespace conflicts on `pandas` otherwise

sklego/pandas_utils.py
```diff
@@ -0,0 +1,152 @@
...
+        func_args_str = ''.join(', {} = {!r}'.format(*item) for item in list(func_args.items())[1:])
+
+        logger.info(
+            f"[ {func.__name__}(df{func_args_str}) ] "
+            f"n_obs={result_df.shape[0]} n_col={result_df.shape[1]} time={time_taken}")
+        return result_df
+    return wrapper
+
+
+def add_lags(X, cols, lags):
```
koaning: two final things (we're getting close tho!) 

1. `add_lags` isn't tested at the moment. a few tests that confirm error messages would be nice. 
2. @MBrouns we might want to add base behavior to allow the `NA` rows to still be kept around.

sklego/pandas_utils.py
```diff
@@ -0,0 +1,152 @@
...
+    ...                    [7, 8, 9]],
+    ...                    columns=['a', 'b', 'c'],
+    ...                    index=[1, 2, 3])
+
+    >>> add_lags(df, 'a', [1]) # doctest: +NORMALIZE_WHITESPACE
+       a  b  c  a1
+    1  1  2  3  4.0
+    2  4  5  6  7.0
+
+    >>> add_lags(df, ['a', 'b'], 2) # doctest: +NORMALIZE_WHITESPACE
```
koaning: 3. @MBrouns what is better base behaviour if we have lags=2 that we then only put in the second lag or when lags=2 we get [1 ... 2]?

MBrouns: I think with `lags=2` we should only give the second lag. My reasoning for this would be that you can then use this to simulate data becoming available at a later moment.

sklego/pandas_utils.py
Menziess: A drop_na flag that is `True` by default?

sklego/pandas_utils.py
Menziess: @koaning it's a nice functionality, but kind of unintuitive. You can produce the same result with `add_lags(df, ['a', 'b'], range(2))` in this implementation.

@MBrouns this result in the current implementation can achieved by passing `[2]`

koaning: let keep this behavior for now at least. it's kind of an R thing to expect this behaviour but i agree this is more explicit.

sklego/pandas_utils.py
koaning: sounds good.

koaning: one that is in and tested ill click the green merge button.

koaning:
Hi @Menziess could you check the one comment about the testing? Once that is dealt with I'll gladly merge this (this thread is getting long).

koaning:
congratulations and thanks :)

Issue #56, opened 2019-03-20
feature request: pandas log_step


dominiquemaria:
decorator that logs how a pandas dataframe is modified;
information:
- number of rows
- number of columns
- time taken

PR #62, opened 2019-03-20
Pandas log step

dominiquemaria:
Helloooooo

Added a decorator to that returns logs on modifications made on pandas dataframes. Information that it returns: 
- function name
- number of rows after modification
- number of cols after modification
- time taken to execute the function 

corresponds to issue #56.


koaning:
grrRRRRReeetings

tests/test_pandas_utils.py
```diff
@@ -0,0 +1,22 @@
+import pandas as pd
+import numpy as np
+import logging
+
+from sklego.pandas_utils import log_step
+
+logging.basicConfig(level=logging.INFO)
+
+
+def test_logging(caplog):
```
koaning: maybe add a fixture for a more general dataframe 

- maybe add a test that also demonstrates that the number of rows actually decrease. 
- i might also enjoy a test that shows the `**kwargs` appear

sklego/pandas_utils.py
```diff
@@ -0,0 +1,18 @@
+import inspect
+from functools import wraps
+import datetime as dt
+import logging
+
+
+def log_step(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        logger = logging.getLogger(inspect.stack()[1].function)
```
koaning: maybe a comment that explains what `inspect.stack()` is for. it doesn't feel totally obvious. @MBrouns it's been suggested that you are involved in this discussion. @dominiquemaria correct?

MBrouns: I usually put my `log_step` in a separate `utils.py` module but prefer the logs to not all be coming from this utils module, that's why I use inspect to get the `__name__` one stack up the stackframe

koaning: @dominiquemaria feel free not to test for all circumstances initially. ill gladly merge this work if you take care of the other two items.

MBrouns: Fair enough! agreed

koaning:
congratz. you can haz os contributorship.

PR #65, opened 2019-03-21
Debug pipeline

CorXccelerated:
A pipeline with a log statement in between the steps of a pipeline. Useful for debugging a pipeline.


sklego/pipeline.py
```diff
@@ -0,0 +1,151 @@
...
+
+    The log statement is added by overwriting the cache method of the memory,
+    such that the function called in the cache is wrapped with a functions that
+    calls the log callback function (`log_callback`).
+
+    This hack breaks will break when:
+    1) The SKlearn pipeline initialization function is changed.
+    2) The memory is used differently in the fit.
+    3) The :class:joblib.memory.Memory changes behaviour of the cache function.
+    4) The :class:joblib.memory.Memory starts using a `_cache` method.
```
koaning: Two things. 

1. Have you checked if `make docs` renders this docstring appropriately? You do not need to send doc changes but because you're not using the `:param :` syntax this might be something to confirm. 
2. Is it possible to add an example to this docstring as well?

notebooks/debug_pipeline.ipynb
```diff
@@ -0,0 +1,317 @@
+{
```
koaning: The notebook is good to have but would it be possible to offer the information that is in here in the documentation as well? If that is work for a later commit that is fine, we can always add it later, but typically the docs should be the central place where to demo code.

sklego/pipeline.py
```diff
@@ -0,0 +1,151 @@
+'''
```
koaning: I think this docstring is added on a module level and since it is possible that we might have other pipelines in the future (like the `BoosterPipeline`) it might be best to describe the module in a more general setting.

koaning:
In particular (about the testing) ... could we make sure that the tests that are provided will check that: 

- it works on old sklearn objects as well as some of our own
- it can accurately log the right stepnames 
- it accurately shows the shape of the input/output 
- it will also still work in a `GridSearchCV` ?

CorXccelerated:
Considering the docs. I adjusted the docs in order for Sphinx to pick it up.

It is also possible to set sphinx to recognize the numpy doc style, which is I think a little more readble then the (default) sphinx style. But considering the rest of the project has this doc style, we will stick with the sphinx style.

CorXccelerated:
Considering test, I copied the tests from sklearn and ran them for this pipeline.

There is one minor difference. In the debug pipeline, the memory is checked on initialisation in the sklearn pipeline it is checked during fit. Some of the tests check if an error is raised when wrong memory is supplied, for the debug pipeline this will happen when creating the memory not when fitting. I adjusted the tests accordingly.

Considering the python motto: `fail early`. And that I do not _really_ see any other purpose of the pipline to be fitted and just for prediction, I think it is ok that the debug pipeline will fail on initialisation when wrong memory is provided. (You could argue that this makes it easier to debug anyway.)

CorXccelerated:
Grid search works on a couple classifiers.

tests/test_debug_pipeline_w_sklearn_test.py
```diff
@@ -0,0 +1,1035 @@
+"""
```
MBrouns: Are these just the tests copy-pasted from sklearn? That's probably a good idea but it might be nicer to actually import the tests from sklearn and monkeypatch the sklearn pipeline object to your debugpipeline so that changes in the sklearn tests automatically will be captured here as well. 

If you also think this is a good idea I'd be open to help you / add it myself

CorXccelerated: They are from sklearn yes, but I had to change one test. (Actually two because `make_debug_pipeline` is not implemented yet, but that is for future worry.) One test fails `test_set_pipeline_step_none` because there is a new property, namely the `log_callback`.

sklego/pipeline.py
CorXccelerated: Both are done.

sklego/pipeline.py
CorXccelerated: Done.

notebooks/debug_pipeline.ipynb
CorXccelerated: Done.

sklego/pipeline.py
```diff
@@ -0,0 +1,289 @@
...
+
+
+import logging
+import datetime as dt
+
+from sklearn.pipeline import Pipeline
+from sklearn.utils.validation import check_memory
+
+
+def default_log_callback(output, execution_time, **kwargs):
```
MBrouns: should this function be private?

MBrouns:
This is some awesome work!

tests/test_debug_pipeline_w_sklearn_test.py
MBrouns: Ah yes that makes sense

sklego/pipeline.py
CorXccelerated: Made it public such that it is visible in the docs.

CorXccelerated:
Tests to check the logs are included. And test if it works on a custom transformer object is included also!

PR #69, opened 2019-03-22
Run model by group

syahdeini:
need feedback captain @koaning.
Do I need to return dictionary?


koaning:
I fear you may have misinterpreted what a `sklearn` model needs to do. This might be better discussed in person. Feel free to ping me.

PR #81, opened 2019-03-28
Let RandomAdder only transform train data

MBrouns:
Implemented a mixin which allows doing different transformations on your train and test set


MBrouns:
Also, I hate the name I currently gave the mixin so I'm definitely open to suggestions there.

sklego/transformers.py
```diff
@@ -2,28 +2,27 @@
 from sklearn.utils import check_array, check_X_y
 from sklearn.utils.validation import FLOAT_DTYPES, check_random_state, check_is_fitted
 
+from sklego.common import TransformerSelectorMixin
 
-class RandomAdder(TransformerMixin, BaseEstimator):
+
```
koaning: do we need to import both our transformer mixin as well as the one from sklearn?

sklego/common.py
```diff
@@ -0,0 +1,85 @@
+import hashlib
+
+import numpy as np
+import pandas as pd
+from sklearn.utils.validation import check_is_fitted, check_array, check_X_y
+
+
+class TransformerSelectorMixin:
```
koaning: so .... `TrainOnlyTransformerMixin` ?

MBrouns: I've also thought about that one and it does sound better but in theory you can use it to transform train and test differently, not just only transform train.
I do agree that transforming train only is probably going to be the main use case so I might as well call it that.

sklego/transformers.py
MBrouns: I let the TransformerMixin not inherit from anything, so currently you need to inherit them both. My thinking here was that you could have other mixins as well and still be flexible with the MRO as a user inheriting from them

koaning: I don't mind that idea but then we need an argument that this will happen in the future. Can you give more examples of Mixins?

MBrouns: Good question, I'm not completely sure. Maybe some kind of LoggerMixin or CachingMixin. Main thing is expectation wise that Mixins typically only inherit from object (same as the sklearn mixins e.g. TransformerMixin) and it would feel odd to let RandomAdder only inherit from the mixin and then let the mixin inherit from BaseEstimator.

Other option would be to just inherit from this Mixin and rename it to something without mixin but regular inheritance doesn't seem like the right solution here

MBrouns: After a night of sleep: I think it makes sense to combine them, can't think of a use case where you would need it any other way

sklego/common.py
```diff
@@ -0,0 +1,86 @@
...
+     - implement `transform_train()`
+
+    They may also implement `transform_test()`. If it is not implemented,
+    `transform_test` will simply return the untransformed dataframe
+    """
+
+    HASHERS = {
+        pd.DataFrame: lambda X: hashlib.sha256(pd.util.hash_pandas_object(X, index=True).values).hexdigest(),
+        np.ndarray: lambda X: hash(X.data.tobytes()),
+        np.memmap: lambda X: hash(X.data.tobytes()),
```
koaning: nice touch.

sklego/transformers.py
```diff
@@ -2,28 +2,27 @@
...
+class RandomAdder(TrainOnlyTransformerMixin, BaseEstimator):
     def __init__(self, noise=1, random_state=None):
         self.noise = noise
         self.random_state = random_state
 
     def fit(self, X, y):
-
+        super().fit(X, y)
```
koaning: i might be missing something but why do we **need** to call `fit` on super here?

MBrouns: in `super().fit()` we compute the hash on `X`

PR #91, opened 2019-03-31
GMM methods - classification *and* outlier detection

koaning:
As far as outlier detection is concerned, this is the current flow: 

```
import numpy as np
import pandas as pd
import plotnine as p9

from sklego.mixture import GMMOutlierDetector

X = np.random.normal(-10, 1, (2000, 2))
mod = GMMOutlierDetector(n_components=1, threshold=0.99).fit(X)

df = pd.DataFrame({"x1": X[:, 0], "x2": X[:, 1],
                   "loglik": mod.gmm.score_samples(X), 
                   "prediction": mod.predict(X).astype(str)})

(p9.ggplot() + p9.geom_point(data=df, mapping=p9.aes("x1", "x2", color="loglik")))
```

image

```
(p9.ggplot() + p9.geom_point(data=df, mapping=p9.aes("x1", "x2", color="prediction")))
```

image

Issue #3, opened 2019-01-23
feature request: loess/lowess regression 


koaning:
I've always wanted to have loess regression in python. R has a cool version of it but it has always been missing. This would be a great model to host here.

koaning:
The easiest way to explain what it needs to do: 

![](https://raw.githubusercontent.com/simplystats/simplystats.github.io/master/_images/loess.gif)

dvanamstel:
DIBS

tsuhina:
Second that

koaning:
```
library(tidyverse)

df <- data_frame(xs=1:120, ys=10*sin(xs/10) + rnorm(120, 0, xs/10)) %>% 
  filter(ifelse((xs < 100) & (xs > 75), FALSE, TRUE))

ggplot() + 
  geom_point(data=df, aes(xs, ys)) + 
  geom_smooth(data=df, aes(xs, ys), span=0.5)
```

![image](https://user-images.githubusercontent.com/1019791/54868719-dde5d580-4d8f-11e9-9020-8b60924940ef.png)

koaning:
```
ggplot() + 
  geom_point(data=df, aes(xs, ys)) + 
  geom_smooth(data=df, aes(xs, ys), span=0.1)
```

![image](https://user-images.githubusercontent.com/1019791/54868725-eccc8800-4d8f-11e9-9362-ccf11d48c718.png)

koaning:
```
ggplot() + 
  geom_point(data=df, aes(xs, ys)) + 
  geom_smooth(data=df, aes(xs, ys), span=1)
```

![image](https://user-images.githubusercontent.com/1019791/54868727-f48c2c80-4d8f-11e9-9ee9-7d560ccd057d.png)

koaning:
http://web.ipac.caltech.edu/staff/fmasci/home/astro_refs/LocalRegressionBook_1999.pdf

dvanamstel:
Nice book, thanks!

I expect to have a few hours next week to finish up what we started.

koaning:
It seems to be in statsmodels. Maybe it is not a right fit.

PR #92, opened 2019-03-31
Loess Regression

dvanamstel:
Still WIP, but we think in an enough mature stage to have a review and actually being a useful tool.

Prio 1 is to add tests, Prio 2 is to clean the _get_knn_symmetric_indices windowing function.

PS. My git notes the file docs/_static/js/modernizr.min.js has changed (not intentional), is this one automatically updated because there are new Docstrings in the package? Left it out of the commits, let me know if we should add.


sklego/loess.py
```diff
@@ -0,0 +1,332 @@
...
+
+import matplotlib.pyplot as plt
+import numpy as np
+from sklearn.base import BaseEstimator
+from sklearn.base import TransformerMixin
+from sklearn.linear_model import LinearRegression
+from sklearn.neighbors import NearestNeighbors
+from sklearn.pipeline import Pipeline
+from sklearn.preprocessing import PolynomialFeatures
+
```
koaning: `Loess` can be part of `sklego.linear`.

sklego/loess.py
```diff
@@ -0,0 +1,332 @@
...
+                 model=None,
+                 n_degree=2,
+                 transformer_type=True,
+                 window_method='fixed',
+                 window_size=2,
+                 step_size=.2,
+                 fraction=.1):
+        """
+
+        Loess Regression. This class implements multiple methods of Loess regression for the single
```
koaning: image

not that i want to make this a *super* important thing, but could you make the width a bit smaller of the docstring here? makes reviewing in smaller screens easier.

sklego/loess.py
```diff
@@ -0,0 +1,332 @@
...
+
+        In transformer mode, the instance will take each data_point and return it's y value changed
+        according to the Loess regression. Hence, no interpolation is done in between missing
+        datapoints.
+
+        In estimator mode, the instance will create a x_focal_base to evalue the y values for. The
+        created x_focal_base is an equally spaced distance vector between x_min and x_man with
+        distance step_size between elements. Hence, interpolation will happen because of the fixed
+        step sizes.
+
```
koaning: might be good to add a reference to the paper in the docstring as well.

sklego/loess.py
```diff
@@ -0,0 +1,332 @@
...
+            x_window = x[self.indices[index]].reshape(-1, 1)
+            y_window = y[self.indices[index]].reshape(-1, 1)
+
+            # TODO: implement iterative weighted fit
+            model = self.model.fit(x_window, y_window)
+            self.y_focal_base = np.append(self.y_focal_base, model.predict(x_focal))
+            self.model_per_window = np.append(self.model_per_window, model)
+
+
+def plot_windows(x, y, x_focal_base, y_focal_base, indices):
```
koaning: if you want to make plots and have this be part of the library -> lets discuss that but at least place it in a submodule for it `sklego.plotting` for example.

sklego/loess.py
```diff
@@ -0,0 +1,332 @@
...
+        ax = plt.axes(xlim=(np.min(x), np.max(x)), ylim=(np.min(y), np.max(y)))
+        ax.scatter(x[indices[index]], y[indices[index]])
+        ax.scatter(x_focal_base[index], y_focal_base[index], c='r')
+        ax.set_title(f'Window for {x_focal_base[index]}')
+        plt.pause(0.01)
+        plt.show()
+
+
+# HELPER FUNCTIONS TO GENERATE RANDOM DATA
+def random_x(minimum_val, maximum_val, size):
```
koaning: this should be part of `sklego.datasets` if we think it is important enough to have.

sklego/loess.py
```diff
@@ -0,0 +1,332 @@
...
+    Generate n random data-points of size between minimum_val and right bounds
+    :param minimum_val: float, minimal value of the generated data
+    :param maximum_val: float, maximum value of the generated data
+    :param size: tuple or list, shape of the desired output
+    :return: x, y
+    """
+    return (maximum_val - minimum_val) * np.random.random(size=size) + minimum_val
+
+
+def generate_noisy_sine_data(noise_std):
```
koaning: this should be part of `sklego.datasets` if we think it is important enough to have.

sklego/loess.py
```diff
@@ -0,0 +1,332 @@
...
+
+            # TODO: implement iterative weighted fit
+            model = self.model.fit(x_window, y_window)
+            self.y_focal_base = np.append(self.y_focal_base, model.predict(x_focal))
+            self.model_per_window = np.append(self.model_per_window, model)
+
+
+def plot_windows(x, y, x_focal_base, y_focal_base, indices):
+    """
+    TODO: Use matplotlib.animation to create video
```
koaning: if we want to make this animation availble it might be better to place this somewhere in the documentation. perhaps add a script to a new `/examples` folder in root?

koaning:
hi django, 

feel free to notify me again when you believe this work is ready for a review. if you still need to add lots of things it might be better to delay the review. i prefer to see a few tests before i start properly reviewing.

in the meantime to answer your questions: 

1. yes testing should happen first
2. cleaning code is always good
3. feel free to ignore anything related to the docs, that's been patched. you may want to update your branch such that it has all the changes that master currently has (your branch is running a bit behind now i think).

koaning:
to make the reviewing a bit easier, feel free to add some screenshots of what your method does. check out [this pr](https://github.com/koaning/scikit-lego/pull/91) to see an example of what i mean.

koaning:
feel free to check if flake8 passes on your local machines.

koaning:
**discussion** 

so ... i've also made a hacky implementation of a loess prediction today. it looks like this: 

```
def make_prediction(x_point):
    np_arr = np.array([x_point])
    wts = np.exp(-np.array([euclidean(x, np_arr) for x in X])/10)
    return LinearRegression().fit(X, y, sample_weight=wts).predict(np_arr.reshape(-1, 1)).reshape(-1)

make_prediction([1, 2])
``` 

it is super hacky. but substantially it is a short hack and it has a few benefits:

- it is easy to maintain
- if a span is given we can first make a subselect on `X` to accomodate a span
- the `wts` param can be used to allow for any kernel 

im havinig a bit of trouble fully understand the code that was submitted. is there a particular reason why the approach listed here won't work?

sklego/loess.py
dvanamstel: Considering:
- Loess can use non-linear models for the local fits
- The local model fitting is non-linear in itself

Are you sure about this?

sklego/loess.py
dvanamstel: I guess the functions itself are not that important, will move them to the development notebook.

dvanamstel:
> im havinig a bit of trouble fully understand the code that was submitted. is there a particular reason why the approach listed here won't work?

TL;DR: The current implementation has a lot of code for creating the animation, this can be moved outside of the actual LoessSmoother. Let's discuss on implementation and then indeed Tomi and I can revise the whole thing before the next review.


---
I don't see any reason why your proposal wouldn't work. Actually, if you look at the last _fit_model_per_window function, it is similar to your code and the core of the whole thing.

All other code around it mostly to create and store the information to create the GIF you posted [here](https://github.com/koaning/scikit-lego/issues/3#issuecomment-468754430).

I do agree it is better to separate the actual loess estimator and creating fancy animations. I also think it is cleaner to use the wts param for windowing instead of the explicit dataset filtering.

If you want to integrate the Smoother in a sklearn pipeline I think two things can be considered:
1. When calling .fit, no actually fitting is done in the LoessSmoother. When .predict is called, under the hood it first fits the model.
2. When calling .fit, it actually creates and fits a model for all the windows in the dataset (the number of windows are defined by the window_size, step_size, etc. parameters). Then when calling predict, it uses the closest model for prediction.

Now that I think about it, I silently assumed and went for the second option. I currently can't think of a reason why the first wouldn't work.

I think it's best to have a discussion next to a whiteboard on what is required/desired, after which we can update stuff accordingly!

sklego/loess.py
koaning: actually ... we need to follow the sklearn api so it should be `sklego.linear_model` 

the local model *is* linear, it might just have polynomial features going in. the *features* are non-linear such that the model can be.

koaning:
it seems like you've made a very small change to `modernizer.js`. if you could revert the change it might make merge-ing a whole lot easier. 

image

PR #95, opened 2019-04-01
Sara

SCarozza:
Added transformer that cap features to a maximum value. The transformer is in preprocessing.py.
Given a pd.Dataframe, a list of columns to cap, an absolute threshold value or a quantile value, the selected columns are capped to the threshold. 
- If the list of columns is none, all columns are capped.
- if the absolute threshold is a list, columns in the list are capped with the corresponding thresholds in the threshold list
- similarly for quantile thresholds

A test is added for this transformer in tests_transformers


sklego/preprocessing.py
```diff
@@ -144,3 +144,52 @@ def _check_X_for_type(X):
         """Checks if input of the Selector is of the required dtype"""
         if not isinstance(X, pd.DataFrame):
             raise TypeError("Provided variable X is not of type pandas.DataFrame")
+            
+            
+            
+
+class ColumnsCapper(TransformerMixin):
```
koaning: should also inherit from `BaseEstimator`

tests/test_transformers/test_capper.py
```diff
@@ -0,0 +1,15 @@
+from sklego.preprocessing import ColumnsCapper
+from sklearn.base import TransformerMixin
+import pandas as pd
+
+
+test_data = pd.DataFrame({'var1':range(101)})
```
koaning: might be better as a fixture.

tests/test_transformers/test_capper.py
```diff
@@ -0,0 +1,15 @@
...
+capper = ColumnsCapper(quantile_thresholds=0.9)
```
koaning: you're bound to write more than one test for this transformer. so it might be better to add this initialisation in the test itself.

tests/test_transformers/test_capper.py
```diff
@@ -0,0 +1,15 @@
...
+
+def test_capper():
+    assert (capper.transform(test_data).iloc[91]['var1'] == 90.0)
```
koaning: maybe more than 1 test is also appropriate :)

tests/test_transformers/test_capper.py
```diff
@@ -0,0 +1,15 @@
+from sklego.preprocessing import ColumnsCapper
```
koaning: there should also be a `test_basics.py` file in the folder where we test estimators. 

since this estimator also has properties we test there it might be nice to add tests there too for the `ColumnsCapper`.

sklego/preprocessing.py
```diff
@@ -144,3 +144,52 @@ def _check_X_for_type(X):
...
+                            list, there is one value for each column to cap
+    absolute_threshold: `float or list of floats` (e.g. 100) absolute value to use to cap values. if list, there is one
+                            value for each column to cap
+    cols_names: `list of column names or None`. if list, cap these columns. if None, caps all columns in dataframe
+    use_quantile: `bool`. if true uses the quantile, otherwise uses the absolute threshold
+    
+    
+    """
+    
+    def __init__(self, quantile_thresholds, cols_names=None, absolute_thresholds=None, use_quantile=True):
```
koaning: after thinking about it... what is a more obvious api? 

```python
ColumnCapper(col_names=["foo", "bar"], threshold=[0.6, 0.1], method="absolute")
```

vs. 

```python
ColumnCapper(col_names=["foo", "bar"], 
                           absolute_thresholds =[0.6, 0.1])
```

@MBrouns do you have an opinion on this? It feels like we might experience this issue a bunch of times in our APIs.

MBrouns: In the second case I kind of expect to be able to set multiple thresholds (one for each type), if that is the case I would go with the second, otherwise with the first

koaning:
I've kept the changes from @SCarozza alive in another PR that might be a good starting point for another hackathon, can be found here: https://github.com/koaning/scikit-lego/pull/98.

PR #96, opened 2019-04-01
Outlier gmm dan std

orchardbirds:
Added a method for stdev outliers. The brief outline is this:
In the fit function:

1. Calculate likelihoods(X) and mean. Let's call this mean mu.
2. Fit a gaussian kernel to fit the likelihoods.
3. Locate the value of X that corresponds to the peak of the kernel.
4. Remove all likelihoods to the right of this peak (i.e. the super likely ones) to make a new group.
5. Call this group new_likelihoods.
6. Calculate standard deviation of new_likelihoods. Let's call this std.

```python
likelihood threshold = mu - (threshold * std)
```


koaning:
thanks!

Issue #77, opened 2019-03-23
[FEATURE] patsy feature selection


koaning:
patsy is sweet this can be a great transformer to get pandas into sklearn.

koaning:
see their [quickstart](https://patsy.readthedocs.io/en/latest/quickstart.html)

A user interface might be:

```
PatsyTransformer("x1 + np.log(x2 + 10)")
PatsyTransformer("y ~ a + a:b + np.log(x)")
```

If a left hand side of the formula is given it should be ignored since this is technically handled by `.fit(X, y)` but everything else could be used to select columns.

koaning:
I'm picking this up this weekend.

PR #103, opened 2019-04-06
patsy transformer

koaning:
This is a simple transformer to allow for a [patsy](https://patsy.readthedocs.io/en/latest/formulas.html#the-formula-language) formula to be used to define items you'd like in a scikit learn pipeline. 

## Features 

```python
tf = PatsyTransformer("a + np.log(a) + b")
tf.fit_transform(df)
```

If `a` is a numeric column and `b` is a discrete (say string) column then this transformer will create a design matrix ready for pipelines that create four columns: 

1. an intercept
2. column `a` as is 
3. the `log` of column `a` 
4. the dummy variables that below to column b. if there are more than 2 classes then multiple columns shall be created. 

If you do not want to create an intercept you can specify so in the formula. 

```python
tf = PatsyTransformer("a + np.log(a) + b - 1)
tf.fit_transform(df)
```

This is one of those features that feels a bit strange unless you are used to patsy or R formulas. But it is reasonably intuitive. 

## Caveats 

This will **only** work on a pandas dataframe which means lot's of SKlearns standard testing tools won't really apply to this object. We need a formula to select the right thing from a dataset.


koaning:
https://github.com/koaning/scikit-lego/issues/77

MBrouns:
This looks nice! A question I have though (mainly because I'm not super familiar with patsy): What happens if there are different categories in your categorical column for train and test? It doesn't look like we're storing any of that info on `self` atm

MBrouns:
Apparently you can get `dmatrix.design_info` and store that on `self` (https://patsy.readthedocs.io/en/latest/stateful-transforms.html#stateful-transforms)

koaning:
The thing about the categories; this is a somewhat general issue in SKLearn. I could add the info to the object in order to help it debug. 

Does sklearn have standard behavior for this? I imagine it may get treated as an NA.

koaning:
Mhm. This is actually kind of interesting behavior. Investigating!

koaning:
@MBrouns I think most issues should now be accounted for. See the test.

MBrouns:
I made some changes in #104

koaning:
yeah that seems like an easier thing to maintain that change. lgtm.

PR #109, opened 2019-04-17
Lo(w)ess regression

tsuhina:
Updates from the previous pull:
- tests have been added
- code has been cleaned up
- docstrings added
- example notebook and .gif file added (for illustrative purposes)
- added weighted regression (None or euclidean)


tests/test_estimators/test_loess.py
```diff
@@ -0,0 +1,120 @@
+import numpy as np
+import pytest
+from sklearn.pipeline import Pipeline
+
+from sklego.linear import LoessRegressor
+
+
+def test_loessregressor_init_normal():
```
koaning: what is this test testing that isn't tested by other tests?

tests/test_estimators/test_loess.py
```diff
@@ -0,0 +1,120 @@
...
+
+def test_loessregressor_in_pipeline():
+    """
+    Check the LoessRegressor doesn't break an sklearn pipeline object in calling fit and predict
+    methods and returns expected values
+    """
+
+    model = LoessRegressor(span=0.5)
+
+    pipeline = Pipeline([('loess', model)])
```
koaning: might be good to make a test where you don't just test the pipeline but also the gridsearch.

sklego/linear.py
```diff
@@ -0,0 +1,140 @@
...
+            y = self.ys[idx_window]
+
+            weights = self._create_weights(x, X)
+            model = LinearRegression().fit(X, y, sample_weight=weights)
+
+            y_pred = np.append(y_pred, model.predict(x.reshape(-1, 1)))
+
+        if with_indices:
+            return y_pred, indices
+        else:
```
koaning: the else seems verbose.

sklego/linear.py
```diff
@@ -0,0 +1,140 @@
...
+            if with_indices:
+                indices.append(idx_window)
+
+            X = self.xs[idx_window].reshape(-1, 1)
+            y = self.ys[idx_window]
+
+            weights = self._create_weights(x, X)
+            model = LinearRegression().fit(X, y, sample_weight=weights)
+
+            y_pred = np.append(y_pred, model.predict(x.reshape(-1, 1)))
```
koaning: it is more performant not to append. instead it might be better to initialise an array with zeros and then fill the values by assignment.

sklego/linear.py
```diff
@@ -0,0 +1,140 @@
...
+
+    """
+
+    def __init__(self, weighting_method: Union[str, None] = None, span: float = .1):
+        super().__init__()
+
+        self._check_init_inputs(weighting_method, span)
+
+        self.weighting_method = weighting_method
+        self.span = span
```
koaning: might be good to have something like `self.allowed_weighting_methods` to confirm which methods are allowed.

sklego/linear.py
```diff
@@ -0,0 +1,140 @@
...
+        :param weighting_method: String id of the weighting method to be used. Should be 'euclidean'
+            or 'equal'.
+        :param span: Float in (0,1]. Sets the fraction of data to use for making local predictions.
+            If 1, the full training data set is used.
+        """
+
+        if not 0 < span <= 1:
+            raise ValueError("Span should be larger than 0 and smaller or equal to 1")
+
+        expected_weighting_methods = ['euclidean', None]
```
koaning: why not add this to `self`?

koaning: also, why is `None` a setting?

sklego/linear.py
```diff
@@ -0,0 +1,140 @@
...
+        the given point x. The size of the returned set of indices is determined
+        by the span setting.
+
+        :param x: data point to find closest points to
+        :return: List of indices of the data points closest to the requested point
+        """
+
+        n_points = int(len(self.xs) * self.span)
+
+        knn = NearestNeighbors(n_neighbors=n_points).fit(self.xs.reshape(-1, 1))
```
koaning: training a `NearestNeighbor` here to make a selection seems like a very expensive way to calculate the distances. why not use `scipy.distance`?

sklego/linear.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(self, weighting_method: Union[str, None] = None, span: float = .1):
+        super().__init__()
+
+        self._check_init_inputs(weighting_method, span)
+
+        self.weighting_method = weighting_method
+        self.span = span
+        self.xs = None
+        self.ys = None
+        self.dim_ = None
```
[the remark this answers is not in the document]
dvanamstel: The value is set in the predict function and used for the "check_is_fitted" function inside the predict function. Using this "check_is_fitted" has been done in order to stay close to how sklearn does things.

tests/test_estimators/test_loess.py
dvanamstel: It was included to -only- test that initialization goes as expected, following the philosophy of testing one thing only per test.

 Using the .fit to actually check it was a bit ugly, modified it to just asserting initializing with a valid span works.

sklego/linear.py
dvanamstel: Check

sklego/linear.py
dvanamstel: Check

sklego/linear.py
```diff
@@ -0,0 +1,140 @@
...
+from sklearn.neighbors import NearestNeighbors
+from sklearn.utils import check_X_y
+from sklearn.utils import check_array
+from sklearn.utils.validation import FLOAT_DTYPES
+from sklearn.utils.validation import check_is_fitted
+
+
+class LoessRegressor(BaseEstimator, RegressorMixin):
+    """
+    Lo(w)ess regressor.
```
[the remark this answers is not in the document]
dvanamstel: Check

sklego/linear.py
dvanamstel: Changed 'None' to 'unweighted' to make it more clear that will make the algorithm not use any weights.

sklego/linear.py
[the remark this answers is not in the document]
koaning: check as in, it is taken care of, or check as in, agreed that this needs to be done?

sklego/linear.py
dvanamstel: Updated to using scope's KDTree as the indices are required.

sklego/linear.py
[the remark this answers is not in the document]
dvanamstel: Both: agreeing and taking care of it.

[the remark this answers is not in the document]
dvanamstel: @koaning : I've updated the code and created a PR in Tomi's repo so that he can have a look as well before we update the PR to sklego.

koaning:
be aware of the merge conflict by the way.

dvanamstel:
@koaning: With the extended absence of Tomi I went ahead and reviewed my own code. Hope you like the changes!

sklego/linear_model.py
```diff
@@ -64,3 +73,126 @@ def predict(self, X):
         X = check_array(X, estimator=self, dtype=FLOAT_DTYPES)
         check_is_fitted(self, ['coefs_'])
         return np.dot(X, self.coefs_)
+
+
+class LoessRegressor(BaseEstimator, RegressorMixin):
+    """
```
koaning: this description can be moved to the `__init__` with an explainer of all the settings

tests/test_estimators/test_loess.py
```diff
@@ -0,0 +1,118 @@
+import numpy as np
+import pytest
+from sklearn.pipeline import Pipeline
+
+from sklego.linear_model import LoessRegressor
+
```
koaning: you might want to . check how we test other methods... we use sklearns testing methods as well as our own.

sklego/linear_model.py
```diff
@@ -64,3 +73,126 @@ def predict(self, X):
...
+        self.available_weighting_methods = ['euclidean', 'unweighted']
+
+        super().__init__()
+        self._check_init_inputs(weighting_method, span)
+        self.weighting_method = weighting_method
+        self.span = span
+        self.xs = None
+        self.ys = None
+        self.dim_ = None
+        self.logger = logging.getLogger(__name__)
```
koaning: are you using the logger anywhere?

sklego/linear_model.py
```diff
@@ -64,3 +73,126 @@ def predict(self, X):
...
+    def fit(self, X: nmp.array, y: nmp.array):
+        """
+        Fit the regressor on the data by storing the data sorted on the inputs.
+
+        :param X: Array-like object of shape (n_samples, 1), input data for the model.
+        :param y: Array-like object of shape (n_samples, 1), target output data for the model
+        :return: The fitted instance
+        """
+
+        self.xs, self.ys = check_X_y(X, y, estimator=self, dtype=FLOAT_DTYPES)
```
koaning: why do we need to save the x/y values that were used during fitting?

koaning:
Hi fellas,

so I might want to have a go at implementing this. This PR has been taking a while and the merge request even has conflicts now. Are you @tsuhina @dvanamstel  still actively working on this? If so, I'll gladly leave it open. Otherwise I might prefer to close this PR and work on it. 

Please let me know, cheers. Vincent.

tsuhina:
Hi Vincent,

I stopped actively working on it since May due to personal reasons. So from my side, please feel free to give it a go yourself. Django is currently on holiday, so it may take some time before he replies.

Cheers, Tomislav

Sent from Mail for Windows 10

Issue #66, opened 2019-03-22
[FEATURE] Feature capping


stephanecollot:
Use case:

In ML some time your features have extreme large values or even infinite value (np.inf), we want to cap those values with a feature transformer.

Parameters:
- feature to cap (and/or feature to no cap?)
- min value
- max value

koaning:
I like the sound of this transformer. It sounds simple enough too. 

1. I wonder about features not to cap though. Probably only a few columns need tweaking. This is my guess.
2. Instead of setting a min/max level per column, might it make sense to cap everything at a certain quantile (this is something that you might more easily do against all columns).

arthurpaulino:
Capping can be a dangerous move for `np.inf` cells. `pandas` dataframes interprets divisions by 0 as `np.inf` or `-np.inf`, depending on the signal of the numerator:
```python
>>> import pandas as pd
>>> df = pd.DataFrame({'numer':[1, -1], 'denom':[0, 0]})
>>> df['div'] = df['numer']/df['denom']
>>> df
   numer  denom  div
0      1      0  inf
1     -1      0 -inf
```
When, in fact, divisions by 0 make no sense and could very much be either `np.inf` or `-np.inf` or anything else (undetermined).

Such ±inf cells are common results of feature engineering processes (e.g.: dividing the loan amount by the income). I usually replace them by `np.nan` and then deal with those from there.

I know that, for instance, [LightGBM](https://lightgbm.readthedocs.io/en/latest/index.html) is capable of dealing with nan cells. And it does it in a way that every attempt of mine to input values on nan cells ends up with lower accuracy. So I just leave it to LightGBM these days. I wonder how it does it, though. What do you guys think?

koaning:
The idea here (and feel free to discuss it) is to allow the user to tell the model "you know what ... for my model, if you're older than 70 ... you might as well be 70". the idea here being that depending on the spread in the dataset this might reduce it. The user needs to have a good reason to do so but this is therefore also the users responsibility. 

It might make sense that the "cap" can also be specified manually.

koaning:
Note that such a transformer needs to work on a pandas dataframe as well as the numpy array.

arthurpaulino:
Interesting idea. What if the user had a way to ask the model about the impact of capping beforehand? Or even better, a capping report, so to speak, which would help the user to choose safe capping thresholds.

[...]

PR #119, opened 2019-04-25
Implemented ColumnCapper and its tests

arthurpaulino:
Issue: https://github.com/koaning/scikit-lego/issues/66


arthurpaulino:
Oops, forgot to check flake8.

arthurpaulino:
Alright, this one was tricky: https://github.com/koaning/scikit-lego/issues/120

sklego/preprocessing.py
```diff
@@ -206,3 +206,179 @@ def _check_X_for_type(X):
...
+    2  3   inf      3.00      13.6
+    3  4  14.0      3.70      13.6
+    >>> capper = ColumnCapper(discard_infs=True) # Discarding infs
+    >>> df[['a', 'b']] = capper.fit_transform(df[['a', 'b']]) # Transforming multiple columns
+    >>> df
+          a     b  a_capped  b_capped
+    0  1.15  11.1      1.15      11.1
+    1  2.00  12.0      2.00      12.0
+    2  3.00   NaN      3.00      13.6
+    3  3.85  13.8      3.70      13.6
```
koaning: i wonder ... why are we not seeing `13.8` at `b_capped` here?

tests/test_preprocessing/test_columncapper.py
```diff
@@ -0,0 +1,108 @@
...
+    estimator_checks.check_methods_subset_invariance,
+    estimator_checks.check_transformer_data_not_an_array,
+    estimator_checks.check_transformer_general,
+]))
+def test_estimator_checks(test_fn):
+    test_fn(ColumnCapper.__name__, ColumnCapper())
+
+
+def test_quantiles():
+    def expect_type_error(min_quantile=0.05, max_quantile=0.95):
```
koaning: never seen this pattern before but i like it.

sklego/preprocessing.py
koaning: I also notice that `np.inf` values are filtered out during fit but `NaN` values are not. Is there a reason for this?

tests/test_preprocessing/test_columncapper.py
arthurpaulino: I was improvising. It was my first time implementing automated tests :B

sklego/preprocessing.py
arthurpaulino: Hello, @koaning 

### On the examples

> i wonder ... why are we not seeing `13.8` at `b_capped` here?

I used different quantiles for the second capping... maybe it was not an intuitive move. Do you think I should use the same capping quantiles or keep them different and point out to the fact that I changed the capping values?

### On *inf* cells

> I also notice that `np.inf` values are filtered out during fit but `NaN` values are not. Is there a reason for this?

Yes. The *inf* part came from [here](https://github.com/koaning/scikit-lego/issues/66#issuecomment-486374288) (typo fixed):

> This ColumnCapper could then take all `inf` values and cast them to the 95th percentile and `-inf` to the 5th.

And from [here](https://github.com/koaning/scikit-lego/issues/66#issue-424303629):

> In ML some time your features have extreme large values or even infinite value (np.inf), we want to cap those values with a feature transformer.

Leaving *inf* cells untouched doesn't make much sense to me since we're attempting to guarantee capping limits. But this is just my point of view...

As an attempt to contemplate what I said [here](https://github.com/koaning/scikit-lego/issues/66#issuecomment-486225626):

> I usually replace them by np.nan and then deal with those from there.

I made it an option with the `discard_infs` parameter and pointed out the importance of using it depending on how those *inf* cells were produced. But its default value is `False`, which makes `ColumnsCapper` simply cap them.

### On *nan* cells

Unfortunately, we can't know what to do with the *nan* cells, as you said [here](https://github.com/koaning/scikit-lego/issues/66#issuecomment-486374288):

> `Na` remains tricky, but thats what valueimputers are for.

So I left those out as a responsibility for the user. Enforcing some artificial value on them with this transformer doesn't make much sense to me. But again, this is just my point of view. What do you think?

arthurpaulino: Hello @koaning!

> i wonder ... why are we not seeing `13.8` at `b_capped` here?

I used different quantiles for the second capping... maybe it was not an intuitive move. Do you think I should use the same capping quantiles or keep them different and point out to the fact that I changed the capping values?

> I also notice that np.inf values are filtered out during fit but NaN values are not. Is there a reason for this?

Yeah, `pandas` ignores *nan* cells when computing the quantiles. But if I leave *inf* cells there, they will pull the quantiles towards them.

```python
>>> import pandas as pd
>>> import numpy as np
>>> df = pd.DataFrame([1, 2, 3, 4, 5, np.inf])
>>> df.quantile([0.05, 0.95])
         0
0.05  1.25
0.95   inf
```

tests/test_preprocessing/test_columncapper.py
MBrouns: good improv, I like it!

sklego/preprocessing.py
MBrouns: That makes sense

arthurpaulino:
@MBrouns, thanks for the detailed reviews. Your ideas were very insightful and I'll make the changes soon.

Oh wow, what a learning experience! :smile:

tests/test_preprocessing/test_columncapper.py
```diff
@@ -0,0 +1,108 @@
...
+
+from sklego.common import flatten
+from sklego.preprocessing import ColumnCapper
+from tests.conftest import nonmeta_checks
+
+# ColumnCapper works with nan/inf cells
+nonmeta_checks_allow_nan_inf = [test for test in nonmeta_checks if test.__name__ != 'check_estimators_nan_inf']
+
+
+@pytest.mark.parametrize("test_fn", flatten([
```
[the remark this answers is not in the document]
arthurpaulino: Wouldn't failing at some test block the PR? Or are you saying that you would manually merge it even with failed tests?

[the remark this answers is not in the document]
MBrouns: Ah yes, I should've been more clear. If possible, it's best to use the lists containing all tests for a specific category of transformer. 

If that isn't possible because on of the tests in that list can't pass, copy the contents of the list to the test parameterization and put a comment in the body to indicate which test you removed and why.

arthurpaulino:
Another detail to consider... I think it would be nice to use the [sklearn.preprocessing.RobustScaler](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.RobustScaler.html) standard to provide quantiles:

```python
RobustScaler(with_centering=True, with_scaling=True,
             quantile_range=(25.0, 75.0), copy=True)
```

They use a tuple instead of two separate parameters.

koaning:
might it be nice to allow the user to select the interpolation method used in pandas? looking at the quantile method is seems like there are subtle differences.

koaning:
One final comment from my side at the moment: it might be nice to add a snippet to the documentation. The examples from the docstring are automatically generated but the adoption rate might be even better if there's a separate article attached to our current [preprocessors docs](https://scikit-lego.readthedocs.io/en/latest/preprocessing.html).

arthurpaulino:
> might it be nice to allow the user to select the interpolation method used in pandas? looking at the quantile method is seems like there are subtle differences.
> 
> maybe add a `method=` param?

Great idea. And it's also available on the [numpy API](https://docs.scipy.org/doc/numpy/reference/generated/numpy.quantile.html), which I am planning to use to avoid those conversions to `pandas.DataFrame`, as @MBrouns said.

But isn't `method` too generic? For those who are familiar with pandas/numpy APIs, `interpolation` would be more likely to ring a bell.

> One final comment from my side at the moment: it might be nice to add a snippet to the documentation. The examples from the docstring are automatically generated but the adoption rate might be even better if there's a separate article attached to our current [preprocessors docs](https://scikit-lego.readthedocs.io/en/latest/preprocessing.html).

Good idea. I am a fan of more precise documentation like [sklearn's User Guide](https://scikit-learn.org/stable/user_guide.html).

koaning:
@arthurpaulino agreed -> interpolation > method

arthurpaulino:
@koaning I'm afraid I won't have enough time to write a more in-depth explanation of `ColumnCapper` on the [preprocessors docs](https://scikit-lego.readthedocs.io/en/latest/preprocessing.html) right now but I think it would be nice to have an explicit explanation of each interpolation method in the future.

MBrouns:
Good to hear the min-max thing worked! Was the problem that you want to ignore the NaN values but the numpy method doesn't? if so, there is a [`nanquantile`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.nanquantile.html#numpy.nanquantile) function. Otherwise keeping it in pandas is also fine.

koaning:
I'll update to master branch and merge! 

I also wish to take a moment to point out that this is amazing ... 

... a mere two months ago I was discussing with @MBrouns that it might be a good idea to have an open source library as part of our training curriculum. It seemed awesome to teach people how to contribute but we also noticed that colleagues were writing the same transformers/estimators over and over again at clients. 

The reception of the package has been good and there are companies in the Netherlands that have shown interest in contributing. Then we noticed we saw a pull request from somebody who lives on the other side of the planet who we never met. 

What a time to be alive. 

Obrigado!

arthurpaulino:
@koaning @MBrouns 
Thanks for the opportunity. I am learning a LOT!

arthurpaulino:
@MBrouns Another thing that was holding me back was the *inf* filter. I thought `pandas.DataFrames` did something smarter behind the scenes but my tests confirmed otherwise. Let me show you.

```python
>>> arr = np.array([[1, 2], [3, np.inf]])
>>> arr
array([[ 1.,  2.],
       [ 3., inf]])
>>> arr[arr<np.inf]
array([1., 2., 3.])
```

The mask flattens the `numpy.array`, which is undesirable. It does not happen with `pandas.DataFrame`:

```python
>>> df = pd.DataFrame(arr)
>>> df
     0    1
0  1.0  2.0
1  3.0  inf
>>> df[df<np.inf]
     0    1
0  1.0  2.0
1  3.0  NaN
```

Using `np.nanquantile` works there just like using the `quantile` method. But would it be possible to create an efficient *inf* filter? I thought `pandas` did not copy the array to perform the *inf* filter, but I was wrong:

```python
>>> df_ = df[df<np.inf]
>>> df_
     0    1
0  1.0  2.0
1  3.0  NaN
>>> arr[0,0] = 10  # this will change df. will it change df_, though?
>>> df
      0    1
0  10.0  2.0
1   3.0  inf
>>> df_
     0    1
0  1.0  2.0
1  3.0  NaN
```

It means that, instead of instantiating a `pandas.DataFrame` object, it would be possible to simply make a manual copy of the `numpy.array` and use the `np.putmask` to filter its *inf* cells. And then use the `np.nanquantile` to compute the quantiles even with *nan* cells. Copying the data is something that `pandas` already does, as shown above. The advantage of this approach is to avoid the conversion overhead.

I will make a PR soon to optimize the `fit` method without the need for `pandas.DataFrame` methods :)

koaning:
@arthurpaulino ill have a look at you merge request after pydata amsterdam. both @MBrouns and myself are in the committee. I did however, push your changes to the current release (0.1.6). so congrats, your changes are live :)

PR #123, opened 2019-04-26
Added orthogonal transformer (Closes #80)

### LINKED ISSUE #80: [FEATURE] orthogonal projections
a feature transformer that makes all features linearly independant of eachother. gram-schmidt process.

pim-hoeven:
Closes #80 

Implemented a transformer that makes X orthogonal using a [QR decomposition](https://en.wikipedia.org/wiki/QR_decomposition).

This method is faster in most cases but always as fast as sklearn PCA.


sklego/preprocessing.py
```diff
@@ -206,3 +206,55 @@ def _check_X_for_type(X):
...
+
+    def __init__(self, normalize=False):
+        self.normalize = normalize
+
+    def fit(self, X, y=None):
+        """
+        Store the inverse of R of the QR decomposition of X, which can be used to calculate the orthogonal projection
+        of X. If normalization is required, also stores a vector with normalization terms
+        """
+        input_is_df = isinstance(X, pd.DataFrame)
```
MBrouns: sklearn has a `check_array` function that will convert the dataframe for you and also performs some checks. Also, do we have to convert to a numpy array? I kinda expect all the code you've written to work agnostically of whether it's a dataframe or an array anyway.

sklego/preprocessing.py
```diff
@@ -206,3 +206,55 @@ def _check_X_for_type(X):
...
+
+        if input_is_df:
+            X = np.array(X)
+
+        # Q, R such that X = Q*R, with Q orthogonal, from which follows Q = X*inv(R)
+        Q, R = np.linalg.qr(X)
+        self.inv_r_ = np.linalg.inv(R)
+
+        if self.normalize:
```
MBrouns: I think it would be nice to always have this `normalization_vector_` but if `self.normalize` is False, just instantiate it as a vector of ones. That saves two if statements in the `transform`

tests/test_preprocessing/test_orthogonal_transformer.py
```diff
@@ -0,0 +1,84 @@
...
+@pytest.fixture
+def sample_matrix():
+    np.random.seed(1313)
+    return np.random.normal(size=(50, 10))
+
+
+@pytest.fixture
+def sample_df(sample_matrix):
+    return pd.DataFrame(sample_matrix)
+
```
MBrouns: We have some default tests to see if the estimator confirms to the sklearn api

```
@pytest.mark.parametrize("test_fn", flatten([
    nonmeta_checks,
    general_checks,
    transformer_checks
]))
def test_estimator_checks(test_fn):
    test_fn(OrthogonalTransformer.__name__, OrthogonalTransformer())
```

It would be nice to add those here as well

koaning:
+1 for the PR!

tests/test_preprocessing/test_orthogonal_transformer.py
pim-hoeven: Added those in the latest commit

sklego/preprocessing.py
pim-hoeven: I agree, changed in the latest commit

sklego/preprocessing.py
pim-hoeven: In the latest commit, I have added the check_array for converting to np array. Do we want to return a `pd.DataFrame` if that is the supplied type of `X`?

pim-hoeven:
@MBrouns is resolving requests done by requester?

koaning:
i think it is not done during the pydata conference.

sklego/preprocessing.py
MBrouns: That would indeed be preferred. The easiest would probably be to just call `check_array` but not reassign it to X

MBrouns:
Apologies, indeed the conference had most of my attention as of late. I've resolved all but the one comment

sklego/preprocessing.py
pim-hoeven: Matrix multiplication always results in a numpy array, so if the goal is that a df is returned if a df is the input, then conversion is necessary. (In `fit`, matrixmultiplication is always done). 

Do you want me to add this conversion?

MBrouns: `fit` shouldn't change the input X anyway right?

koaning:
I've recently stumbled on a use-case for this. @MBrouns can we merge it?

MBrouns:
Aye!

PR #124, opened 2019-04-27
setup.py fix

koaning:
this should contain the datasets fix ... i think :)


koaning:
@MBrouns i think this should also fix things

arthurpaulino:
We're still unable to do `sklego.__version__`, though.

scikit-learn defines the version on the [`__init__.py` file](https://github.com/scikit-learn/scikit-learn/blob/7b136e92acf49d46251479b75c88cba632de1937/sklearn/__init__.py#L47), then [imports the `sklearn` module](https://github.com/scikit-learn/scikit-learn/blob/7b136e92acf49d46251479b75c88cba632de1937/setup.py#L40) on the `setup.py` file and sets the version of the package there.

koaning:
@arthurpaulino fair comment. @MBrouns i recall a potential downside of this approach but i can't recall. any objections to this change?

koaning:
@MBrouns i've added the version change.

MBrouns:
It's been a while but something I could imagine with this approach is that your setup.py tries to import `sklego` while it is not yet installed, causing problems when first installing the package. I think it works in this case because we're not doing imports in `__init__.py` (yet), but otherwise I would expect installation errors

The way I've done it in the past (which does feel dirty though):

make a `version.py` with `__version__ = '1.0.6'`
in `__init__.py` add from `sklego.version import __version__`

in setup.py:
```
exec(open('sklego/version.py').read())
setup(
    ...
    version=__version__,
    ...
```

koaning:
Considering that our project is more like a box of tools than an integrated tool I imagine the approach in this PR is fine. In a library like [brent](https://github.com/koaning/brent) this might certainly be different.

MBrouns:
Yes, and also it's still changeable if it does happen to give problems later down the road

PR #125, opened 2019-04-28
 A series of small fixes that I noticed

arthurpaulino:
1. Removed internal/empty modules from docs page
2. Removed nonexistent modules from docs page
    They were raising exceptions when building the docs
3. Fixed a broken link in the end of the [Pasty Formulas](https://scikit-lego.readthedocs.io/en/latest/preprocessing.html#patsy-formulas) documentation
4. Added `.abc` to the `Iterable` call
    Calling `collections.Iterable` was triggering a deprecation warning on `common.flatten`
5. Fixed `tet_meta` directory name (is now `test_meta`)


sklego/common.py
```diff
@@ -125,7 +125,7 @@ def flatten(nested_iterable):
     ['test1', 'test2']
     """
     for el in nested_iterable:
-        if isinstance(el, collections.Iterable) and not isinstance(el, (str, bytes)):
```
koaning: yeah i remember seeing a deprication warning for this.

koaning:
these are indeed welcome changes. thanks!

PR #128, opened 2019-04-29
Optimizing ColumnCapper's fit method

arthurpaulino:
Motivated by https://github.com/koaning/scikit-lego/pull/119#issuecomment-487604935.

I guess this PR closes the issue https://github.com/koaning/scikit-lego/issues/66.

### DIFF
--- a/sklego/preprocessing.py
+++ b/sklego/preprocessing.py
@@ -311,7 +311,8 @@ class ColumnCapper(TransformerMixin, BaseEstimator):
             raise ValueError("ColumnCapper cannot fit columns containing only inf/nan values")
 
         q = [quantile_limit/100 for quantile_limit in self.quantile_range]
-        self.quantiles_ = np.nanquantile(a=X, q=q, axis=0, interpolation=self.interpolation)
+        self.quantiles_ = np.nanquantile(a=X, q=q, axis=0, overwrite_input=True,
+                                         interpolation=self.interpolation)
 
         # Saving the number of columns to ensure coherence between fit and transform inputs
         self.n_columns_ = X.shape[1]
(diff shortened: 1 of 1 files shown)


sklego/preprocessing.py
```diff
@@ -298,30 +298,38 @@ def fit(self, X, y=None):
...
+        if infs_mask.any():
+            if isinstance(X, pd.DataFrame):
+                # Copying X is not enough in this case. We need to go deeper.
+                X = X.values.copy()
+            else:
+                X = X.copy()
+            copied = True
+            # Now we're free to edit X
+            np.putmask(X, infs_mask, np.nan)
+        else:
```
koaning: this is a bit ... nit picky. but technically you do not need the `else`. line 317 could be moved just before the if statement, no?

koaning:
Mhm, there also seems to be a unit-test failure but this might be caused by the master branch. Will try to fix it.

sklego/preprocessing.py
MBrouns: Yes! Also, I'm curious about how much memory the `overwrite_input` saves. If it's not a lot, I might do away with the flag entirely. If it is a lot it might be good to copy `X` regardless whether there are infs or not

arthurpaulino: I'd guess it can save at maximum as much memory as `X` consumes. For the sake of simplicity, I've made it copy `X` regardless.

arthurpaulino:
@koaning @MBrouns I'm not quite understanding what happened. Do I need to fix something on my PR?

koaning:
scikit-learn made a new release which is making some parts of the tests fail.

arthurpaulino:
Oh, I see. Thank you :)

MBrouns:
I'm fine with merging this one! We'll need to take a look at the debug pipeline at a later stage

koaning:
once again @arthurpaulino : muito brigado

PR #134, opened 2019-05-14
sklearn update fix

koaning:


koaning:
this one was actually sensible. 

but there's something else going on on master.

koaning:
image

readme change caused a red dot. might be a version thing.

koaning:
ah. scikit-learn recently upgraded. that seems to cause the error ... 

... we might need to test against multiple versions :(

koaning:
@CorXccelerated just a heads up, this seems to affect the debug pipeline.

koaning:
so ... to me it seems like the testing for the debug pipeline is due for a bit of a revamp. the tests are super long. it does seem like some of the standard error messages out of scikit learn have changed.

koaning:
@MBrouns @CorXccelerated i don't suppose tomorrow y'all are going to be hacking on this :) ?

Issue #32, opened 2019-03-16
feature request: GMM Naive Bayes


koaning:
nuff said.

koaning:
https://github.com/koaning/scikit-lego/pull/135

koaning:
got merged.

PR #135, opened 2019-05-14
Machine Learning Feature: NaiveGMMBayes

koaning:


koaning:
https://github.com/koaning/scikit-lego/issues/32

koaning:
@MBrouns could you check if you agree with the testing/GMMNaiveBayes?

tests/test_estimators/test_gmm_naive_bayes.py
```diff
@@ -0,0 +1,49 @@
...
+from sklego.common import flatten
+from sklego.naive_bayes import GaussianMixtureNB
+from sklego.testing import check_shape_remains_same_classifier
+from tests.conftest import nonmeta_checks, general_checks
+
+
+@pytest.mark.parametrize("test_fn", flatten([
+    nonmeta_checks,
+    general_checks,
+    estimator_checks.check_classifier_data_not_an_array,
```
MBrouns: I guess you're not using all `estimator_checks`? It might be good to add a comment about which ones not and why.

koaning: I've put in all `classifier_checks` and ... it passes.

PR #141, opened 2019-05-22
[WIP] debug pipeline use log message

JCZuurmond:
To have a more seamless integration with the new sklearn pipeline. These adjusted try to mimic the same behaviour as much as possible.

We should discuss if we want this. What is logged is slightly different than before (see notebook).

WIP: update docstring example.


koaning:
@JCZuurmond can I delete this PR? It seems a bit old by now.

JCZuurmond:
Yes, you can delete this PR. 

The intention is to use 'print' instead of 'logging'. Print is the convention in sklearn, but I was not happy with my rewrite. And I did not spend time on it anymore.

PR #146, opened 2019-05-31
Decay estimator

koaning:
Made a `DecayEstimator` model.


koaning:
@MBrouns tests have been added. Will work on an example for the docs (one from the pydata talk) if you've got no more comments on it.

sklego/meta.py
```diff
@@ -117,3 +117,60 @@ def predict(self, X):
...
+    def fit(self, X, y):
+        """
+        Fit the data after adapting the same weight.
+
+        :param X: array-like, shape=(n_columns, n_samples,) training data.
+        :param y: array-like, shape=(n_samples,) training data.
+        :return: Returns an instance of self.
+        """
+        X, y = check_X_y(X, y, estimator=self, dtype=FLOAT_DTYPES)
+        self.weights_ = np.cumprod(np.ones(X.shape[0]) * self.decay)[::-1]
```
MBrouns: I think we don't need `weights_` as an attribute right? We don't use it in any other method

sklego/meta.py
```diff
@@ -117,3 +117,60 @@ def predict(self, X):
...
+
+    The DecayEstimator will use exponential decay to weight the parameters.
+
+    w_{t-1} = decay * w_{t}
+
+    """
+
+    def __init__(self, model, decay: float = 0.999):
+        self.model = model
+        self.decay = decay
```
MBrouns: Maybe it's premature but I could imagine that passing a decay function to class rather than a single linear decay weight could add some more flexibility. Something like:

```
def linear_decay(decay):
   return lambda distance: ...

class GeneralDecay():
    def __init__(self, model, decay_func):
        ....

LinearDecay = lambda model, decay_weight: GeneralDecay(model, linear_decay(decay_weight)) 
```

sklego/meta.py
koaning: i added it more as a convenience. say' for plotting purposes. this is especially useful when we have support for multiple decay functions.

sklego/meta.py
koaning: I agree on this but this requires a bit of thought. We could internally allow for multiple variants of decay but perhaps this is better implemented as a string (this might make gridsearch results a bit nicer). This might be nice to skip in this PR because I like the idea of adding a decay function as an exercise for one of our students. 

A more involved implementation (one that is interesting) involves a convenient method to assign decay not based on time/order but rather on things like: 

- is it a holiday? 
- is this customer important? 
- any other property that is in the data 

For this latter part I do not see a nice implementation.

koaning: A more involved implementation (one that is interesting) involves a convenient method to assign decay not based on time/order but rather on things like: 

- is it a holiday? 
- is this customer important? 
- any other property that is in the data 

For this latter part I do not see a nice implementation yet though.

sklego/meta.py
MBrouns: that's a fair point!

sklego/meta.py
MBrouns: Adding a decay function is indeed a nice exercise but the framework should already allow for it beforehand I would say. I'm open to skipping it for this PR though

koaning: I've added the option for `decay_func`. It currently does nothing but it gives space. 

```
    def __init__(self, model, decay: float = 0.999, decay_func="exponential"):
        self.model = model
        self.decay = decay
        self.func = decay_func
```

Issue #29, opened 2019-03-10
feature request: RBF Features


koaning:
this is like the repeating RBF features except that this ... won't repeat. it will simply span the entire space of a variable.

Issue #20, opened 2019-03-01
feature request: RBFRepeater


koaning:
feature generation that can be used for timeseries. trick from the london talk.

MaxHalford:
Hey @koaning,

I stumbled upon your talk a few days ago and really enjoyed many of your talking points. I was curious about the RBF kernel trick so I decided to implement it in an online learning library me and some friends are working on. From what I understand The idea is simply to computing the distance between, say, a month and all the 12 months of the year using a RBF. This way September is closer to August than it is to March, which isn't taken into account if one simply one-hot encodes the month. Is this correct? If you're interested I coded it at the end of this [notebook](https://creme-ml.github.io/notebooks/Building%20a%20simple%20time%20series%20model.html).

koaning:
1. That cream stuff sounds cool beans. I'll give it a spin. Also: PyData Amsterdam has a CFP open at the moment. I'm still in the committee and that cream library sounds like something we'd love to host. 
2. The goal here is to make an `sklearn` compatible transformer that is general. Your example is good but our goal is to be able to supply a date column and a number of RBFs you'd like per year. There's going to be a sprint this wednesday so I'll keep this thread up to date.

MaxHalford:
1. Sounds great! I've started making some slides (written in English) for an upcoming of the data science Meetup back here in Toulouse, so maybe I can reuse them.
2. Okay good to know, I just wanted to make sure the maths were right. Indeed I think that having a transformer to extract date features would be nice because it could then pipeline into a `RBFTransformer`.

Good stuff!

koaning:
Question about creme: most of the learning that occurs, is that just a small SGD step that occurs per datapoint or is there something more happening? SKlearn has some [passive agressive](http://koaning.io/posts/passive-agressive-algorithms/) things [api here](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.PassiveAggressiveClassifier.html), but creme is not doing that atm? 

I like the idea of doing a rolling mean on an intercept by the way.

MaxHalford:
[...]

PR #147, opened 2019-05-31
[wip] add repeating and spanning basis functions

RensDimmendaal:
I'm building preprocessing functions to address:
#20 
#29 

Open Issues:
1. RepeatingBasisFeatures currently accepts only one feature. This is problematic for most tests.

### DIFF
--- a/sklego/preprocessing.py
+++ b/sklego/preprocessing.py
@@ -377,3 +377,106 @@ class ColumnCapper(TransformerMixin, BaseEstimator):
         allowed_interpolations = ('linear', 'lower', 'higher', 'midpoint', 'nearest')
         if interpolation not in allowed_interpolations:
             raise ValueError("Available interpolation methods: {}".format(', '.join(allowed_interpolations)))
+
+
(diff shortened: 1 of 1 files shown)


kayhoogland:
Any updates on this @RensDimmendaal? Would love to use it!

RensDimmendaal:
Far point. I have left this to sit for too long.  It is unclear to me how to run the standard tests this because transformer only accepts one column. @koaning mentioned the idea of wrapping the transformer with a column selector for the tests. I'll try that in the next 7 days.

MBrouns:
Wrapping the thing in a columntransformer is probably a reasonable default anyway! Basically something like this

```
class RepeatingBasisFunction:
    def __init__(column, *args, **kwargs):
        return ColumnTransformer([
            ('repeatingbasis', _RepeatingBasisFunction(*args, **kwargs), column)
        ], remainder='passthrough')

class _RepeatingBasisFunction:
    pass # Your current implementation
```

RensDimmendaal:
Continued in a different branch #171

PR #150, opened 2019-06-10
Datasets: Make Simple Timeseries

koaning:
i needed a dataset for the `Decay` part of the docs and i figured adding a good timeseries generator might make a good addition. 

this is work in progress at the moment. it turns out to be rather tricky to get consistent seasons in.


koaning:
@MBrouns I've created a small helper function that can generate time-series data. Should be useful to demonstrate the benefit of a `DecayEstimator` as well as the `RepeatingBasisFeatures`.

MBrouns:
Nice thing to have!

 At some point it could be cool if we could add some kind of fluent api where we can add different seasonalities:

```
(TimeSeries(start_date=..., end_date=..., freq=...)
    .add_trend(start_date=, end_date, ...)
    .add_trend()
    .add_seasonality(period=n)
)
```

But that's definitely a more long term thing

PR #152, opened 2019-06-14
Information filter

koaning:


koaning:
using nb sphinx for docs 

image

sklego/preprocessing.py
```diff
@@ -420,3 +420,68 @@ def transform(self, X):
         X = check_array(X, estimator=self)
 
         return X @ self.inv_R_ / self.normalization_vector_
+
+
+def scaler_(vec, unto):
```
MBrouns: You actually already compute the projection here, right? Should this function be called `scalar_projection`? and the `project_` then `vector_projection`?

koaning: it'd be a minor rename only, but sure, it's more explicit that way.

sklego/preprocessing.py
```diff
@@ -420,3 +420,68 @@ def transform(self, X):
...
+    return vec.dot(unto)/unto.dot(unto)
+
+
+def project_(vec, unto):
+    return scaler_(vec, unto) * unto
+
+
+class InformationFilter(BaseEstimator, TransformerMixin):
```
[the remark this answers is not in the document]
koaning: ah yes.

koaning:
@MBrouns could you send a screenshot? I'm not sure where the "how it works" is supposed to be.

koaning:
image

found and fixed it

sklego/preprocessing.py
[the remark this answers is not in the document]
koaning: done.

sklego/preprocessing.py
koaning: done.

PR #153, opened 2019-06-18
P percent rule

MBrouns:
Implements the p_percent fairness metric described in https://arxiv.org/pdf/1507.05259.pdf

Also made a start on implementing the actual classifier they're discussing but that should probably go in a separate PR

### DIFF
--- a/sklego/metrics.py
+++ b/sklego/metrics.py
@@ -18,7 +18,7 @@ def correlation_score(column):
     return correlation_metric
 
 
-def neg_p_percent(column, positive_target=1):
+def p_percent_score(column, positive_target=1):
     r"""
     The p_percent score calculates the ratio between the probability of a positive outcome
     given the sensitive attribute (column) being true and the same probability given the
@@ -35,7 +35,7 @@ def neg_p_percent(column, positive_target=1):
     :param column: Name of the column (when X is a dataframe)
     or the index of the column (when X is a numpy array).
     :param positive_target: The name of the class which is associated with a positive outcome
-    :return: Negative p percent score (in gridsearch, larger is better and we want to typically punish correlation).
+    :return: p percent score
     """
     def impl(estimator, X, y_true=None):
         """Remember: X is the thing going *in* to your pipeline."""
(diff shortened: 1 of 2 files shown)


sklego/metrics.py
```diff
@@ -16,3 +16,41 @@ def correlation_metric(estimator, X, y_true=None):
...
+        sensitive_col = X[:, column] if isinstance(X, np.ndarray) else X[column]
+
+        if not np.all((sensitive_col == 0) | (sensitive_col == 1)):
+            raise ValueError(f'neg_p_percent only supports binary indicator columns for `column`. '
+                             f'Found values {np.unique(sensitive_col)}')
+
+        y_hat = estimator.predict(X)
+        y_given_z1 = y_hat[sensitive_col == 1]
+        y_given_z0 = y_hat[sensitive_col == 0]
+        p_y1_z1 = np.sum(y_given_z1 == positive_target) / (len(y_given_z1) or 1)
```
koaning: why not `np.mean` here?

tests/test_metrics/test_neg_p_percent.py
```diff
@@ -0,0 +1,59 @@
...
+def test_neg_p_percent_pandas(sensitive_classification_dataset):
+    X, y = sensitive_classification_dataset
+    mod_unfair = LogisticRegression().fit(X, y)
+    assert neg_p_percent("x2")(mod_unfair, X) == 0
+
+    mod_fair = make_pipeline(
+        ColumnSelector('x1'),
+        LogisticRegression(),
+    ).fit(X, y)
+    assert neg_p_percent("x2")(mod_fair, X) == -0.9
```
koaning: just to check if i understand. the score for the fair model is ... worse?

MBrouns: yes because lower is better in grid search right?

sklego/metrics.py
MBrouns: good one, will change it

tests/test_metrics/test_neg_p_percent.py
koaning: i thought higher is better? from the [docs](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html) 

image

MBrouns: Thanks for catching that one! Inverted the whole thing. Thats what you get for building features without wifi to check the sklearn docs 🙂

koaning: Might be one for the `import this` feature. 

> Off by one errors are hard to catch, 
> Flipping the sign can be even harder (and worse).

Issue #132, opened 2019-05-09
[FEATURE] Thresholder


koaning:
This is a meta model that can take a classifier but is "more strict" when it comes to picking a class. Usually in binary classification the cut-off point for a class is 0.5 but I can imagine that sometimes you'd like the model to be more strict like move the cut-off to 0.7. 

I'm wondering if we want this object to be able to handle multiple classes but something like this might be sweet.

koaning:
At least for two dimensions it should be a pretty sweet feature.

koaning:
It might make sense here to make a `WontPredictMixin` for this since we probably also want to have a similar behavior once we detect outliers. See [this argument](http://koaning.io/posts/high-on-probability-low-on-certainty/).

PR #154, opened 2019-06-19
Thresholder

koaning:
does what it says.

### DIFF
--- a/sklego/meta.py
+++ b/sklego/meta.py
@@ -130,7 +130,6 @@ class DecayEstimator(BaseEstimator):
     The DecayEstimator will use exponential decay to weight the parameters.
 
     w_{t-1} = decay * w_{t}
-
     """
 
     def __init__(self, model, decay: float = 0.999, decay_func="exponential"):
--- a/sklego/meta.py
+++ b/sklego/meta.py
@@ -163,7 +163,7 @@ class DecayEstimator(BaseEstimator):
 
     def predict(self, X):
         """
-        Predict new data by making random guesses.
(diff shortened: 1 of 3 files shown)


koaning:
https://github.com/koaning/scikit-lego/issues/132

sklego/meta.py
```diff
@@ -175,3 +174,54 @@ def predict(self, X):
...
+        """
+        Fit the data.
+
+        :param X: array-like, shape=(n_columns, n_samples,) training data.
+        :param y: array-like, shape=(n_samples,) training data.
+        :return: Returns an instance of self.
+        """
+        X, y = check_X_y(X, y, estimator=self, dtype=FLOAT_DTYPES)
+        self.estimator_ = clone(self.model)
+        if not self._is_classifier():
```
MBrouns: shouldn't this be `if not self.estimator_._is_classifier()`?

Also, we should probably check whether the `estimator` has `predict_proba`. Not all classifiers necessarily have that.

Remember James' trick?

```
class ProbabilisticClassifierMeta(type):
    def __instancecheck__(self, other):
        return hasattr(other, 'predict_proba')

class ProbabilisticClassifier(metaclass=ProbabilisticClassifierMeta):
    pass
    
class SomeModel:
    def predict_proba(self):
        return 1
```

Then in the `fit` you can do:
```
isinstance(estimator, ProbabilisticClassifier)
```
I think it could be neat

koaning: I've added `sklego.base`.

PR #156, opened 2019-06-20
FairClassifier

MBrouns:
todo:

- [x] fix tests comparing with logisticregression
- [x] show that decreasing threshold increases p_percent_score
- [x] fix for when y is object array instead of number
- [x] Multiclass?

Source:
    - M. Zafar et al. (2017), Fairness Constraints: Mechanisms for Fair Classification


sklego/linear_model.py
```diff
@@ -64,3 +68,126 @@ def predict(self, X):
...
+        column_or_1d(y)
+        label_encoder = LabelEncoder().fit(y)
+        y = label_encoder.transform(y)
+        self.classes_ = label_encoder.classes_
+
+        if len(self.classes_) > 2:
+            raise ValueError(f"This solver needs samples of exactly 2 classes"
+                             f" in the data, but the data contains {len(self.classes_)}: {self.classes_}")
+
+        self._fit(sensitive, X, y)
```
koaning: not sure if I like `_fit` as a method name. isn't `_solve` maybe a bit nicer?

sklego/linear_model.py
```diff
@@ -64,3 +68,126 @@ def predict(self, X):
...
+        n_obs, n_features = X.shape
+        theta = cp.Variable(n_features)
+        y_hat = X @ theta
+
+        log_likelihood = cp.sum(
+            cp.multiply(y, y_hat) -
+            cp.log_sum_exp(cp.hstack([np.zeros((n_obs, 1)), cp.reshape(y_hat, (n_obs, 1))]), axis=1)
+        )
+        if self.penalty == 'l1':
+            log_likelihood - cp.sum((1 / self.C) * cp.norm(theta[1:]))
```
koaning: shouldn't this be `-=` instead of `-`?

MBrouns: Yes! that's why my regularization tests didn't work, thanks

sklego/linear_model.py
MBrouns: Agreed

sklego/linear_model.py
koaning: makes sense :)

koaning:
I only have comments about documentation but this is best discussed in person. The `InformatitionFilter` commit contains a fairness part of the documentation and it might be a good idea to add this classifier to that example. This can be done later though since that PR would first need to get merged. 

See https://github.com/koaning/scikit-lego/pull/152

Issue #158, opened 2019-06-27
[BUG] Grouped Estimator does not accept string values in group. 


koaning:
Fix please it is breaking code at a supermarket. 

```python
from sklego.meta import GroupedEstimator
from sklego.datasets import load_chicken
from sklearn.linear_model import LinearRegression

df = load_chicken(give_pandas=True) 
df['diet'] = ['omgomgomg' + s for s in df['diet'].astype(str)] 

X = df[['time', 'diet']]
y = df['weight']
GroupedEstimator(LinearRegression(), groups=['diet']).fit(X, y).predict(X)
```

Probably the grouped column needs to be removed from the internal `check_array` checker.

pim-hoeven:
Please see proposed solution in PR #159

PR #159, opened 2019-07-03
Bugfix for #158

pim-hoeven:
Fixed the bug caused when using the grouped_estimator with a string column as grouping variable. 

**Solution:** Try the checks without adjustments, if that fails: remove the grouping column from the array or dataframe.


koaning:
i wonder ... when would we want to have the column in the group? it should be a constant value. could it not be removed in all cases?

pim-hoeven:
1. My pleasure!
2. @MBrouns can I get a sticker?
3. That was my first approach, but then some other tests failed, for example in checking for `check_estimators_nan_inf` and `check_fit2d_predict1d` (cannot drop a column in that case). Therefore I adopted this EAFP approach. Do you agree?

koaning:
ah yes. those are tricky. 

i see two paths going forward and maybe a check with @MBrouns might be good. 

1. We can make an estimator that *contains* a `GroupEstimator` in a `sklearn.pipeline.Pipeline`. The idea is to have a preprocessing step that *adds* the column that we want to remove. This might cause the tests to pass ... this option is OK, but a bit meh. 
2. Alternatively we might also ignore those two tests in particular. Especially for `meta` estimators we cannot have the same rules as for a base estimator in sklearn. I would be fine with dropping those two tests in favour of, what I feel, is better behaviour for the model that is being implemented. This option feels best for me at the moment. 

The sklearn tests are guidelines that are easy to automate but they aren't devine gospel. @pim-hoeven feel free to question my thoughts here if you have a better alternative.

pim-hoeven:
I think the second approach makes more sense for two reasons:
1. It's easier to implement and keeps the `GroupEstimator` cleaner and less complicated
2. Especially the `check_fit2d_predict1d` makes no sense in this context: if we don't have a grouping column then why use a `GroupEstimator`?

Regarding the `check_estimators_nan_inf` we might want to build in checks on the grouping columns as well, specifically that it does not contain missing values. 

So, (@MBrouns) my proposed solution would be:
- Include a `__check_group_columns` in the estimator that checks whether the grouping column(s) exist and are valid
- Do the `check_X_y` on the non-grouping columns
- Remove the `check_fit2d_predict1d`

Do you agree?

koaning:
Sounds great to me.

sklego/meta.py
```diff
@@ -54,6 +54,47 @@ def __init__(self, estimator, groups, use_fallback=True):
...
+
+        if len(diff) > 0:
+            raise KeyError(f'{diff} not in columns of X ({x_cols})')
+
+    @staticmethod
+    def __check_missing_and_inf(X):
+        """Check that all elements of X are non-missing and finite"""
+        if np.any(pd.isnull(X)):
+            raise ValueError("X has NaN values")
+        try:
```
koaning: i wonder ... why do we need a try/except here?

sklego/meta.py
```diff
@@ -54,6 +54,47 @@ def __init__(self, estimator, groups, use_fallback=True):
...
+            if np.any(np.isinf(X)):
+                raise ValueError("X has infinite values")
+        except TypeError:  # X cannot be converted to numeric, so checking infinites does not make sense
+            pass
+
+    def __validate(self, X, y=None):
+        try:
+            X_data = X.drop(columns=self.groups, inplace=False)
+        except AttributeError:  # np.array
```
koaning: it feels cleaner to test this using `isinstance(X, pd.DataFrame)` would you not agree?

sklego/meta.py
```diff
@@ -54,6 +54,47 @@ def __init__(self, estimator, groups, use_fallback=True):
...
+            X_data = np.delete(X, self.groups, axis=1)
+
+        if y is not None:
```
koaning: I might add a comment here to explain that you do this because we validate during both fitting as well as prediction.

tests/test_meta/test_grouped_model.py
```diff
@@ -16,9 +16,7 @@
     estimator_checks.check_estimators_nan_inf,
     estimator_checks.check_estimators_overwrite_params,
     estimator_checks.check_estimators_pickle,
-    estimator_checks.check_fit2d_predict1d,
     estimator_checks.check_fit2d_1sample,
-    estimator_checks.check_fit1d,
```
koaning: if you remove checks thats perfectly fine, but please add a comment indicating why it has been removed. this makes it easier to maintain.

sklego/meta.py
```diff
@@ -54,6 +54,47 @@ def __init__(self, estimator, groups, use_fallback=True):
         self.groups = groups
         self.use_fallback = use_fallback
 
+    def __check_group_cols_exist(self, X):
+        try:
```
koaning: checking the type of `X` is probably better to do via `isinstance` instead waiting for an `AttributeError`.

koaning:
1. Great work! This is an improvement. 
2. I did have some comments on this. Mainly around the checking if something is a `pd.DataFrame` object or not. My view suggests that asserting this via `isinstance` is a better habbit. Feel free to challenge me on this. 
3. There are some places where adding comments makes it easier to understand *why* a certain decision is made.

pim-hoeven:
1. Thanks, and thank you for the comments!
2. I do not have a strong view on this personally. I used to do these checks using `isinstance`, but read somewhere that the try-except approach is preferred (because, for example, we might in the feature want to pass something that behaves like a DataFrame but does not necessarily pass the isinstance check).  See for example the answers to [this so post](https://stackoverflow.com/questions/21024913/python-isinstance-vs-hasattr-vs-try-except-what-is-better). Still, I'm not very strongly opinionated on this point (and agree that isinstance is more readable), so if this argument does not convince you than I'll gladly change the try-excepts to isinstance checks.
3. Agreed, I will add some comments

sklego/meta.py
pim-hoeven: The check for infinities is only relevant in case of numeric variables. If the check raises `TypeErrror`, the values are not numeric (and cannot be converted) and so checking for infinities does not make sense. Moved the comment at the exception to the try for more clarity.

sklego/meta.py
MBrouns: I agree that in this case the intent is clearer with an isinstance check.

sklego/meta.py
```diff
@@ -54,6 +54,52 @@ def __init__(self, estimator, groups, use_fallback=True):
...
+            x_cols = set(X.columns)
+        except AttributeError:
+            try:
+                ncols = X.shape[1]
+            except IndexError:
```
MBrouns: is `X` ever only 1d?

sklego/meta.py
```diff
@@ -54,6 +54,52 @@ def __init__(self, estimator, groups, use_fallback=True):
...
+
+        # Check whether grouping columns exist
+        diff = set(as_list(self.groups)) - x_cols
+
+        if len(diff) > 0:
+            raise KeyError(f'{diff} not in columns of X ({x_cols})')
+
+    @staticmethod
+    def __check_missing_and_inf(X):
+        """Check that all elements of X are non-missing and finite"""
```
MBrouns: Doesn't `check_array` (https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_array.html) allow forcing finites etc?

sklego/meta.py
koaning: I guess it could be. groupby-mean as a model?

sklego/meta.py
pim-hoeven: Yes, but that does not work well for string values, which we need for the grouping column.

koaning:
No worries about the delay, it is all volunteer work.

koaning:
There is something interesting happening in the ci pipeline though. It runs fine locally? 

```
________ ERROR collecting tests/test_meta/test_estimatortransformer.py _________
ImportError while importing test module '/home/travis/build/koaning/scikit-lego/tests/test_meta/test_estimatortransformer.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_meta/test_estimatortransformer.py:2: in <module>
    from pandas.tests.extension.numpy_.test_numpy_nested import np
E   ModuleNotFoundError: No module named 'pandas.tests.extension.numpy_'
___________ ERROR collecting tests/test_meta/test_outlier_remover.py ___________
ImportError while importing test module '/home/travis/build/koaning/scikit-lego/tests/test_meta/test_outlier_remover.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_meta/test_outlier_remover.py:2: in <module>
    from pandas.tests.extension.numpy_.test_numpy_nested import np
E   ModuleNotFoundError: No module named 'pandas.tests.extension.numpy_'
```

koaning:
urgh. it seems unrelated to your work, i think pandas has a new version or something. 

image

koaning:
yep. it's pandas. 

image

pim-hoeven:
No I didn't have these locally, but only ran the relevant tests. Is the pandas issue something I can / should change?

koaning:
i haven't looked at your code yet but i'll prolly make a fix tomorrow/day after. once that is merged to master we can try again. :)

koaning:
@pim-hoeven do check the conversations. it feels like a comment here and there might still be missing.

koaning:
I am fixing the failing tests here; https://github.com/koaning/scikit-lego/pull/167

koaning:
Also, just to check @pim-hoeven, you've got a sticker right?

pim-hoeven:
- Added some more comments to code
- Added sticker to laptop

Issue #160, opened 2019-07-03
[BUG] kmeanfold does xpass?


koaning:
@kayhoogland @MBrouns did you folks also see this in your test results? 

```
tests/test_meta/test_thresholder.py .................                                                                                                   [ 49%]
tests/test_metrics/test_correlation_score.py ..                                                                                                         [ 49%]
tests/test_metrics/test_neg_p_percent.py ...                                                                                                            [ 50%]
tests/test_model_selection/test_klusterfold.py ........X........X........X........X........X........X........X........X........X........X........X..... [ 57%]
...X........X........X........X........X........X........X........X........X........X........X........X........X........X........X........X             [ 67%]
tests/test_model_selection/test_timegapsplit.py ....                                                                                                    [ 67%]
tests/test_pandas_utils/test_pandas_utils.py ......                                                                                                     [ 68%]
```

koaning:
at the end it seems to `xpass`? any idea what that is?

```
=================================================== 1383 passed, 27 xpassed, 638 warnings in 77.32 seconds ====================================================
```

kayhoogland:
The xpass is the last test in test_klusterfold. For cluster methods that predict one cluster it should raise a ValueError.

kayhoogland:
They should actually be xfail. See my new pull request: https://github.com/koaning/scikit-lego/pull/161

MBrouns:
I didn't look at the detailed test report, only checked whether the CI was green. I'm a bit surprised that an xpass actually results in the test suite as a whole passing. It might be preferable to use `with pytest.raises(ValueError):` instead in that case.

MBrouns:
The other option would be to add `strict=True` to the xfail decorator

koaning:
I prefer `with pytest.raises(ValueError):` mainly because it is consistent with the rest of the API. 

Also, I know that I am not the most senior pytest user. But I never heard of `xpass` before today. I can imagine introducing it can be very confusing for other users.

kayhoogland:
I agree. See my updated pull request.

PR #161, opened 2019-07-04
Bugfix Xpass -> Xfail

kayhoogland:
Bugfix for issue: https://github.com/koaning/scikit-lego/issues/160


koaning:
I might prefer `with pytest.raises(ValueError):` mainly because it is consistent with the rest of the API. 

Also, I know that I am not the most senior pytest user. But I never heard of `xpass` before today. I can imagine introducing it can be very confusing for other users.

Issue #157, opened 2019-06-21
[FEATURE] ColumnDropper


koaning:
It might be easier to sometimes explain what you *don't* want.

PR #166, opened 2019-07-24
Feature/column dropper

kayhoogland:
#157


koaning:
Fixing failing tests here: https://github.com/koaning/scikit-lego/pull/167

koaning:
I am OK with merge-ing this but @MBrouns and I should talk about how to abstract away this numpy/dataframe thing. It feels like this is something we should start building in `common`.

PR #167, opened 2019-07-25
fixes are in

koaning:
It seems that we were using pandas extensions (a hidden one) instead of plain numpy. Can't remember why. Changed the code and the tests seem to run fine now.


koaning:
Once this is fixed I can have another look at the other PRs.

PR #169, opened 2019-07-25
Shrinkage estimator

pim-hoeven:
Extension of the `GroupedEstimator`. We would like to be able to estimate models based on hierarchical groups. 

Example: 
Hierarchical groups can be continent, country, city. Our model for the city is then a combination of a model for the city itself and a model for the country it is in, which in turn is a combination of a model for the country itself and a model for the continent it is on.

Shrinkage, or the weight of the previous model, can be (for example) a constant fraction, or depending on both group sizes, or depending on the performance of the models, etc.


sklego/meta.py
```diff
@@ -100,7 +152,69 @@ def __remove_groups_from_x(self, X):
...
+        """Validate the shrinkage function"""
+        group_lengths = [10, 5, 2]
+        expected_shape = np.array(group_lengths).shape
+        try:
+            result = self.shrinkage_function_(group_lengths)
+            if not isinstance(result, np.ndarray):
+                raise ValueError(f"shrinkage_function({group_lengths}) should return an np.ndarray")
+            if result.shape != expected_shape:
+                raise ValueError(f"shrinkage_function({group_lengths}).shape should be {expected_shape}")
+        except Exception as e:
```
JCZuurmond: Try not to use general exceptions. If catching multiple exceptions do them separate or as group (in a tuple).

sklego/meta.py
```diff
@@ -100,7 +152,69 @@ def __remove_groups_from_x(self, X):
...
+
+        return grouped_estimations
+
+    def __check_shrinkage_func(self):
+        """Validate the shrinkage function"""
+        group_lengths = [10, 5, 2]
+        expected_shape = np.array(group_lengths).shape
+        try:
+            result = self.shrinkage_function_(group_lengths)
+            if not isinstance(result, np.ndarray):
```
JCZuurmond: Does not make sense to raise an Error and catch it directly. The following makes more sense (in my opinon)

```      
try:
    result = self.shrinkage_function_(group_lengths)
except Exception:         # specify the exception: ValueError, etc
     raise ValueError(f"Caught an exception while checking the shrinkage function.") from e
else:
    if not isinstance(result, np.ndarray):
        raise ValueError(f"shrinkage_function({group_lengths}) should return an np.ndarray")
    if result.shape != expected_shape:
         raise ValueError(f"shrinkage_function({group_lengths}).shape should be {expected_shape}")
```
```

File: sklego/meta.py [hunk 3]
```diff
@@ -100,7 +152,69 @@ def __remove_groups_from_x(self, X):
...
+                raise ValueError(f"shrinkage_function({group_lengths}) should return an np.ndarray")
+            if result.shape != expected_shape:
+                raise ValueError(f"shrinkage_function({group_lengths}).shape should be {expected_shape}")
+        except Exception as e:
+            raise ValueError(f"Caught an exception while checking the shrinkage function: {str(e)}")
```
JCZuurmond: `raise ValueError("some text") from e` will give you more information in the traceback.

https://stackoverflow.com/questions/57345684/how-to-catch-an-my-exception-raise-in-python-and-bounce-it-again

sklego/meta.py
```diff
@@ -100,7 +152,69 @@ def __remove_groups_from_x(self, X):
...
+
+    def __set_shrinkage_function(self):
```
JCZuurmond: I think a Python mantra is to "fail fast". Maybe makes sense to do these checks during the init (if `self.shrinkage` is given).

Consider:

```
shrink_options = {
   "constant": constant_shrinkage,
   "relative": relative_shrinkage,
   "min_n_obs": min_n_obs_shrinkage,
}
if isnstance(self.shrinkage_function, str) and self.shrinkage_function not in shrink_options:
   raise ValueError('The specified ....') 

self.shrinkage_function = shrink_options.get(self.shrinkage_function, self.shrinkage_function)
```

Maybe add another check to see if `self.shrinkage_function` is callable when not a `str`

sklego/meta.py
```diff
@@ -100,7 +152,69 @@ def __remove_groups_from_x(self, X):
...
+                "relative": relative_shrinkage,
+                "min_n_obs": min_n_obs_shrinkage,
+            }
+            try:
+                self.shrinkage_function_ = shrink_options.get(self.shrinkage_function)
+            except AttributeError:
+                raise ValueError(f"The specified shrink function {self.shrinkage_function} is not valid, "
+                                 f"choose from {list(shrink_options.keys())}")
+        else:
+            self.shrinkage_function_ = self.shrinkage_function
```
JCZuurmond: why the variables with trailing underscores?  Trailing underscores are used when variables clash with python built-in functions. e.g. `filter_`.

sklego/meta.py
```diff
@@ -114,62 +228,129 @@ def fit(self, X, y):
...
-                           .drop_duplicates()
-                           .loc[lambda d: d['new'] == 1]
-                           .itertuples())
-            raise ValueError(f"found a group(s) {culprits} in `.predict` that was not in `.fit`")
+        if not self.shrinkage:
+            return self.__predict_group(X, group_colnames=self.group_colnames_)
+        else:
+            return self.__predict_shrinkage_groups(X)
+
+    def __expanding_list(self, list_to_extent: List[str], return_type=list) -> Union[List[list], List[tuple]]:
```
JCZuurmond: Separate function or `@staticmethod`?

sklego/meta.py
MBrouns: It's also sklearn convention for attributes that are learned in `fit`

sklego/meta.py
```diff
@@ -100,7 +152,69 @@ def __remove_groups_from_x(self, X):
...
+            }
+        except KeyError:
+            raise KeyError("The indices of y and X should match")
+
+        return grouped_estimations
+
+    def __check_shrinkage_func(self):
+        """Validate the shrinkage function"""
+        group_lengths = [10, 5, 2]
+        expected_shape = np.array(group_lengths).shape
```
JCZuurmond: I am in doubt if this check makes sense here. This is basically a test. 

The default shrinkage functions should be tested in the tests. And it might be fair to expect the user to supply a valid shrinkage function themselves.

sklego/meta.py
```diff
@@ -114,62 +228,129 @@ def fit(self, X, y):
...
+                self.estimators_.update(
+                    self.__fit_grouped_estimator(X, y, self.value_colnames_, level_colnames)
+                )
+        else:
+            self.estimators_ = self.__fit_grouped_estimator(X, y, self.value_colnames_, self.group_colnames_)
+
+        self.groups_ = as_list(self.estimators_.keys())
+
+        if self.shrinkage:
+            self.complete_groups_ = [grp for grp in self.groups_ if len(as_list(grp)) == len(self.group_colnames_)]
```
JCZuurmond: maybe pass `complete_groups_` to `complete_groups_ ` instead of creating an attribute?

JCZuurmond: now I had to do a search to find out where this attribute is created

sklego/meta.py
JCZuurmond: ah! Did not know that. 

Then the question is the `shrinkage_function` "learned" during fit?

JCZuurmond:
@pim-hoeven cool feature to add!

I was curious about what you created. And added comments in the process of understanding your code. Feel free to decided yourself which comments are useful and which are not.

sklego/meta.py
JCZuurmond: Or only check the function when it is given by the user.

sklego/meta.py
```diff
@@ -38,24 +41,70 @@ def transform(self, X):
...
+    :param **kwargs: keyword arguments to the shrinkage function
     """
-
-    def __init__(self, estimator, groups, use_fallback=True):
+    def __init__(self, estimator, groups, use_fallback=True, shrinkage=False, shrinkage_function="constant", **kwargs):
         self.estimator = estimator
         self.groups = groups
-        self.use_fallback = use_fallback
+        self.use_fallback = use_fallback  # Do we need this in case of shrinkage?
+        self.shrinkage = shrinkage
```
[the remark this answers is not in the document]
koaning: fair enough.

sklego/meta.py
koaning: It might be preferable to move such a function to `sklego.common`.

sklego/meta.py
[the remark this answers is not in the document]
pim-hoeven: Good idea

sklego/meta.py
```diff
@@ -38,24 +41,70 @@ def transform(self, X):
...
+    The augmented prediction for each level is the weighted average between its prediction and the augmented
+    prediction for its parent.
+
+    Let $\hat{y}_i$ be the prediction at level $i$, with $i=0$ being the root, than the augmented prediction
+    $\hat{y}_i^* = \alpha \hat{y}_i + (1 - \alpha) \hat{y}_{i-1}^*$, with $\hat{y}_0^* = \hat{y}_0$.
+
+
+    """
+    return np.array(
+        [alpha ** (len(group_sizes) - 1)]
```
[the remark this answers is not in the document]
pim-hoeven: The line below is also multiplied by `(1 - \alpha)`, this one is not.

sklego/meta.py
pim-hoeven: I'll move it to `sklego.common`

sklego/meta.py
pim-hoeven: I moved the initialisation of the `shrinkage_function` to the init because one of Cor's other remarks, so also removed the trailing underscore

sklego/meta.py
pim-hoeven: Included this, good idea!

sklego/meta.py
pim-hoeven: The thing with this one is that we really don't know what goes wrong in the user specified shrinkage function, so we want to catch all of these errors

pim-hoeven: Although I do agree that `try ... except ... else` is an improvement!

sklego/meta.py
```diff
@@ -100,7 +152,69 @@ def __remove_groups_from_x(self, X):
...
+    def __fit_grouped_estimator(self, X, y, value_columns, group_columns):
+        group_indices = X.groupby(group_columns).indices
+
+        try:
+            grouped_estimations = {
+                group: clone(self.estimator).fit(X.loc[indices, value_columns], y.loc[indices])
+                for group, indices in group_indices.items()
+            }
+        except KeyError:
+            raise KeyError("The indices of y and X should match")
```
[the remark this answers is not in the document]
pim-hoeven: This check was used before I added `reset_index` a couple of lines above, so I'll remove it as the indices will always match

sklego/meta.py
```diff
@@ -38,24 +41,70 @@ def transform(self, X):
...
-
-    def __init__(self, estimator, groups, use_fallback=True):
+    def __init__(self, estimator, groups, use_fallback=True, shrinkage=False, shrinkage_function="constant", **kwargs):
         self.estimator = estimator
         self.groups = groups
-        self.use_fallback = use_fallback
+        self.use_fallback = use_fallback  # Do we need this in case of shrinkage?
+        self.shrinkage = shrinkage
+        self.shrinkage_function = shrinkage_function  # Default is constant
+        self.kwargs = kwargs
```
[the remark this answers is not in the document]
pim-hoeven: I renamed it to `shrinkage_kwargs` for clarity. Although I liked learning about the `partial`, it is a bit cumbersome to use in gridsearch

sklego/meta.py
pim-hoeven: I moved it to `__init__` but moved it back to `fit` on suggestion of Matthijs because that works better with `sklearn.clone` which is used later on.

pim-hoeven:
@koaning @MBrouns I made some changes based on @CorXccelerated's comments and I'm fairly satisfied with the progress. I removed the WIP, looking forward to your comments :)

koaning:
@pim-hoeven I'll have a look but could you resolve all the conversations that can be considered to be resolved?

sklego/meta.py
```diff
@@ -38,34 +38,113 @@ def transform(self, X):
...
+    :param **shrinkage_kwargs: keyword arguments to the shrinkage function
     """
-
-    def __init__(self, estimator, groups, use_fallback=True):
+    def __init__(self, estimator, groups, value_columns=None, use_fallback=True, shrinkage=None, **shrinkage_kwargs):
         self.estimator = estimator
         self.groups = groups
-        self.use_fallback = use_fallback
+        self.value_columns = value_columns
+        self.use_fallback = use_fallback  # Do we need this in case of shrinkage?
```
koaning: I would be fine with mentioning this gets ignored in the docstring in case of shrinkage. Maybe worth to throw a warning if a value is passed incorrectly.

sklego/meta.py
```diff
@@ -38,34 +38,113 @@ def transform(self, X):
...
+                "min_n_obs": min_n_obs_shrinkage,
+            }
+            try:
+                self.shrinkage_function = shrink_options.get(self.shrinkage)
+            except AttributeError:
+                raise ValueError(f"The specified shrinkage function {self.shrinkage} is not valid, "
+                                 f"choose from {list(shrink_options.keys())} or supply a callable.")
+        elif callable(self.shrinkage):
+            self.__check_shrinkage_func()
+            self.shrinkage_function = self.shrinkage_function
```
koaning: what does this line do exactly?

koaning: should this not be

```
self.shrinkage_function = self.shrinkage
````
?

Comment (koaning) [reply sklego/meta.py]: it seems worth a test to see if this does not go wrong.

File: tests/test_meta/test_grouped_model.py [hunk 15]
```diff
@@ -103,5 +105,122 @@ def test_chickweigt_string_groups():
     y = df['weight']
 
     # This should NOT raise errors
-    GroupedEstimator(LinearRegression(), groups=['diet']).fit(X, y).predict(X)
-    GroupedEstimator(LinearRegression(), groups=1).fit(X_np, y).predict(X_np)
+    GroupedEstimator(LinearRegression(), groups=['diet'], shrinkage=None).fit(X, y).predict(X)
```
koaning: since default value is `shrinkage=None` ... is there a reason to explicitly add it?

sklego/meta.py
koaning: yep. added test. there's a problem here.

sklego/meta.py
pim-hoeven: The thing with this one is that we really don't know what goes wrong in the user specified shrinkage function, so we want to catch all of these errors

sklego/meta.py
pim-hoeven: I chose the latter option

tests/test_meta/test_grouped_model.py
pim-hoeven: No, removed it

sklego/meta.py
pim-hoeven: Sharp, must have happened while refactoring. This test was already included, but caught another error. Your test failed for another reason: when only specifying one column, shrinkage is not possible. Your comment led to two improvements:
- Added the option to shrink to a global model
- Improved tests to look for specific errors, not just a `ValueError`

sklego/meta.py
pim-hoeven: I made an argument `use_global_model` that is `use_fallback` in case of a regular `GroupedEstimator` and adds a shrinked model for the entire input in case of shrinkage.

PR #171, opened 2019-07-30
Add Repeating Basis Functions

RensDimmendaal:
I worked on this a while ago in PR #147, but I started from a fresh branch because I decided to limit the scope only to repeating basis functions (#20), excluding the spanning basis functions (#29).

Feedback adapted so far:
- Write a transformer when X contains only one column
- Write a wrapper which selects one column, when X has more than one column
- Added basic tests
- Added docstring

To Do:
- Write documentation

Could someone review the code? 
I'll work on the documentation in the meantime as well.

Tagging people involved in PR #147: @kayhoogland @koaning @MBrouns


sklego/preprocessing.py
```diff
@@ -609,3 +610,133 @@ def _check_X_for_type(X):
         """Checks if input of the Selector is of the required dtype"""
         if not isinstance(X, pd.DataFrame):
             raise TypeError("Provided variable X is not of type pandas.DataFrame")
+
+class RepeatingBasisFunction(TransformerMixin, BaseEstimator):
+    """
+    This is a transformer for features with some form of circularity. 
+    E.g. for days of the week you might face the problem that, conceptually, day 7 is as 
+    close to day 6 as it is to day 1. While numerically their distance is different.
+    This transformer solves that problem.
```
koaning: `solve` might be too strong a word, maybe `remedy`?

sklego/preprocessing.py
```diff
@@ -609,3 +610,133 @@ def _check_X_for_type(X):
...
+    The transformer selects a column and transforms it with a give number of repeating (radial) basis functions.
```
koaning: "with a give" -> spelling

sklego/preprocessing.py
```diff
@@ -609,3 +610,133 @@ def _check_X_for_type(X):
...
+    :param remainder: By default, only the specified column is transformed, and the 
+        non-specified columns are dropped. (default of ``'drop'``). By specifying 
+        ``remainder='passthrough'``, all remaining columns will be automatically passed 
+        through. This subset of columns is concatenated with the output of the transformer.
+
+    :type n_periods: int, default=12
+    :param n_periods: number of basis functions to create, i.e., the number of columns that
+        will exit the transformer.
+
+    :type floor: float or 'min', default='min'
```
koaning: I am a bit unsure what you mean with the floor/ceil parameters here. The RBF spans between 0-1 no? 

image

koaning:
I've taken a few minutes to have a glance, will try to find some more time to look at the code later.

sklego/preprocessing.py
RensDimmendaal: Thanks. Two actionable points of feedback come out of this:
1. Ensure the curves always range between 0 and 1. Now this is note the case, but that would be neater.
2. Clearly label what ceil/floor do. What they do is scale the input data. Imagine you want to make repeating basis functions on a column with hour of the day. But your training data does not include data points between midnight and 3am. You might still want to have your repeating basis functions be split on the 0-24 range rather than the 3-24 range. Explicitly saying what the minimum and maximum of your input range fixes this.

Will take them on board next week

RensDimmendaal: I've renamed floor and ceil to one parameter `input_range` which accepts a tuple. I hope it now better reflects what it does: it captures at what values the input data starts over again. E.g. for days of the week you'd provide (1,7)

RensDimmendaal:
* Adopted the docstring suggestions
* Rename the floor/ceil parameters to input_range to clarify what it does
* Added documentation as a notebook

Not sure why travis fails, seems to me like the issue is with other notebooks which i have not touched [link](https://travis-ci.com/koaning/scikit-lego/jobs/224227006).

koaning:
So there's a bug in a part of the code that has to do with the `grouped-estimator`. If that part of the library is causing this then I'll still gladly merge the work here.

koaning:
image

but this seems to be formatting on your side?

koaning:
But this you may safely ignore, this is something that I am picking up with @pim-hoeven. 

image

RensDimmendaal:
Thanks. My bad on the formatting issue. Fixed it now.

image

If that other issue can be ignored then I think it is ready.

koaning:
Great. I'll have a look at the end of the day.

koaning:
image

Great that you took the effort of moving the preprocessing documentation into a notebook but is there a reason why only your code is evaluated?

koaning:
I am playing with the notebook and i see something odd. 

image

image


When remainder should be dropping ... it isn't?

koaning:
I am having worries about the remainder behaviour. To me it feels like the base behaviour should be that it always drops the column. In my mind it would be used similar to this: 

image

You would have a feature union split the steps and pass the correct column to the model. I think having passthrough behaviour can still be valid but I would argue that it goes against a base behaviour. 

@MBrouns thoughts?

RensDimmendaal:
**Docs**:
Had to do a slight rewrite of the code snippet you screenshotted above, but I've now converted all snippets in the preprocessing docs to working code.

**Regarding faulty "drop" behavior**:
Shame on me, that's a bug. I've fixed it and added a test for it now as well.

**Default behavior "drop"/"passthrough"**:
That's fair. It makes more sense to have "drop" as the default. That's also the default of the ColumnTransformer which I use to accomplish the behavior.

MBrouns:
regarding the drop / passthrough I think we should go for the principal of least surprise and to me it feels like the transformer is most like the one hot encoder. The OHE keeps all the non-categorical columns and simply adds the encoded columns so for me it would make most sense to have the same behaviour here. 

On the other hand, the `categorical_features` attribute of the OHE is getting deprecated and users are told to use a `ColumnTransformer` instead, so I don't know..

koaning:
Oh yeah. I forgot to click the merge button. @RensDimmendaal well done!

RensDimmendaal:
Whoo! :-)

PR #173, opened 2019-08-06
Update the badges

Fokko:
Make the links clickable as well.

I couldn't update the AppVeyor one since it is on the settings page on which only you have access: You can see a badge image URL and sample markup on the Badges tab of project settings. Badge URLs for both private and public projects contain a security token.


koaning:
Ah. Never noticed that. Fair enough. 

Somethings strange is happening on the travis side though. Will investigate. 

image

koaning:
Mhm. 

https://travis-ci.org/koaning/scikit-lego
https://travis-ci.com/koaning/scikit-lego

This is a bit suspicious, but just in case, @Fokko could you make sure the link points to the dot com version?

Fokko:
@koaning that is some weird stuff 🤔

Fokko:
@koaning done!

PR #174, opened 2019-08-22
Improve .gitpod.yml

jankeromnes:
Hi @koaning!

Many thanks for filing https://github.com/gitpod-io/gitpod/issues/743. I'm looking into why `pytest` is being slow in your repo, and while I was here I took the liberty to tweak your `.gitpod.yml` a little:

- It seems that Gitpod doesn't support the `jupyter/scipy-notebook` Docker image (this could be filed as another bug, but I thought that maybe using that image wasn't necessary)
- I've scripted your README.md's setup instructions so that Gitpod runs them automatically when starting a workspace

I'm now running `pytest` and will try to benchmark what's being slow there. 🙂

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,1 +1,3 @@
-image: jupyter/scipy-notebook
+tasks:
+  - init: pip3 install -e ".[dev]"
+    command: python3 setup.py develop


koaning:
Ah. I may have missed that from the docs. Why can't I use that container?

PR #177, opened 2019-08-22
Update pythonpackage.yml

koaning:
this is an attempt at github actions to see if it makes the workflow easier


koaning:
so I am noticing that github actions is quite a bit faster than travis/appveyor/circleci. this is nice.

PR #181, opened 2019-08-23
Update readme.md

koaning:


koaning:
image

appveyor/travis integration is a bit odd. i still see this even when it is turned off.

koaning:
Fixed!

PR #183, opened 2019-08-25
small doc fix

koaning:
The meta documentation still had an error in it. This fixes that.


koaning:
Oh man. A merge on complete button would be epic.

Issue #163, opened 2019-07-11
[BUG] p-percent divide by zero warnings.


koaning:
There's a lot of warnings when I use the p-percent score in a grid search with the fair classifier. 

![image](https://user-images.githubusercontent.com/1019791/61027244-f79a3f80-a3b5-11e9-87c9-d97a1d8ab428.png)

I think the warning is good, but seeing a kilometer of red that you gotta scroll through in a notebook is less ideal and might frighten people who are new. Can we maybe apply some smoothing to prevent this error?

kayhoogland:
https://github.com/koaning/scikit-lego/pull/185#pullrequestreview-280886896

MBrouns:
fixed in #195

Issue #106, opened 2019-04-10
[FEATURE] import this


koaning:
We're bound to learn lessons that can be combined into a "Zen of Scikit-Lego". Just like projects like sympy and evol we might want to add our own. If there are general zen lessons learned while making this tool: let's up 'em here. Some day in the future somebody might gain enlightenment by typing. 

```
from sklego import this
```

koaning:
if you don't know the requirements you don't know if they're met, 
if you haven't gotten to where you're going, you aren't there yet.

koaning:
you cannot understand a problem by avoiding it, 
but sometimes the problems go away if you ignore them.

koaning:
you can add lot's of features and consider every extension,
but the path towards ruin is paved with good intensions.

arthurpaulino:
[the unknown may be ignored,
but infinity is ever present.](https://github.com/koaning/scikit-lego/pull/119#discussion_r278881928)

koaning:
> defer responsibility to the end user
> this might just empower them

koaning:
https://github.com/koaning/scikit-lego/pull/153

> Off by one errors are hard to catch,
> Flipping the sign can be even harder (and worse).

koaning:
First version of this: https://github.com/koaning/scikit-lego/pull/239.

Menziess:
Correlation does not imply causation.

koaning:
There's a PR here which will add it to the docs. https://github.com/koaning/scikit-lego/pull/409

I might also say that the poem feels complete. People can still add to it. But no longer needed to have this issue list.

PR #185, opened 2019-08-28
Bugfix/p percent divide by zero

kayhoogland:
I discussed https://github.com/koaning/scikit-lego/issues/163 with @MBrouns and we agreed that a wall of red warnings does not look nice but it should be something that the users should control themselves with for example `warnings.filterwarnings('once')`.

However, if a division by zero occurs, the p_percent_score should be 0. If you never predict a positive target for one of the subgroups the model is not fair by definition. In this pull request, I implemented functionality to warn the user about this.

### DIFF
--- a/sklego/metrics.py
+++ b/sklego/metrics.py
@@ -1,6 +1,7 @@
 import numpy as np
 import warnings
 
+
 def correlation_score(column):
     """
     The correlation score can score how well the estimator predictions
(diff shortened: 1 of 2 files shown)


MBrouns:
Apparently this is intentional: https://github.community/t5/GitHub-API-Development-and/Run-a-GitHub-action-on-pull-request-for-PR-opened-from-a-forked/m-p/15746/highlight/true#M415

It basically says that the action will be run on the fork, if that has actions enabled, otherwise it won't be run.

koaning:
That's ... janky.

koaning:
It's suggested that this will get fixed once github actions are out of beta. Mhmm ... this is a bit inconvenient. @kayhoogland could you turn on actions?

koaning:
In regards to the warning behaviour: I agree. 

I agree so much in fact, that I have argued in the past to be poetic about it: https://github.com/koaning/scikit-lego/issues/106#issuecomment-500716350.

koaning:
Fixed it! 

It was a github setting after all.

koaning:
dunno why this is taking so long tho.

image

koaning:
Yeah nevermind something weird is up here. I even clicked "approve" and nothing is moving.

kayhoogland:
Should I open the pull request again?

MBrouns:
please do!

PR #197, opened 2019-09-20
confusion balancer - and more!

koaning:
A few changes were made: 

1. I added the meta-estimator that attempts to cause a balance. I can demonstrate that it does something but I have not created a full example yet of it working super well. My proposal is to add it but not to write documentation for it yet. This is a building block for the voting classifier that keeps model mistakes in mind. 
2. I also added two datasets and made the api adhere to the scikit learn standard. We now have the `abalone` dataset as well as the `heroes` dataset. For the horde!


sklego/meta.py
```diff
@@ -474,3 +476,65 @@ def predict_proba(self, X):
...
+    def fit(self, X, y):
+        """
+        Fit the data.
+
+        :param X: array-like, shape=(n_columns, n_samples,) training data.
+        :param y: array-like, shape=(n_samples,) training data.
+        :return: Returns an instance of self.
+        """
+        X, y = check_X_y(X, y, estimator=self.model, dtype=FLOAT_DTYPES)
+        if not self._is_classifier():
```
MBrouns: We have a `ProbabilisticClassifier` in `base` which does the check you need here so you can use `isinstance(self.model, base.ProbabilisticClassifier)` i think

koaning: yeah i noticed that later on, should be fixed now.

PR #200, opened 2019-09-20
allow setting the log level on the pandas log_step utility

MBrouns:


koaning:
once flake agrees you can merge. i agree with the change.

koaning:
I should say, should there perhaps be documentation for this?

MBrouns:
Agreed on the docs part, I added a notebook and will merge if everything is green!

PR #206, opened 2019-09-22
Equal opportunity Classifier

MBrouns:
We already had our FairClassifier which accounted for Demographic Parity with regards to sensitive attributes, but of course no metric is perfect. If we for example use this in a loan approval situation, the demographic parity only looks at loans given, and not at the rate at which loans are paid back. That might result in a lower percentage of qualified people who are given loans in one population than in another. Another way of measuring fairness could be therefore to measure __equal opportunity__ instead. This constraint would boil down to:

$$P(\hat{y}=1 | z=0, y=1)=P(\hat{y}=1 | z=1, y=1)$$

and be turned into a metric as such:

$$\text{equality of opportunity} = \min \left(\frac{P(\hat{y}=1 | z=1, y=1)}{P(\hat{y}=1 | z=0, y=1)}, \frac{P(\hat{y}=1 | z=0, y=1)}{P(\hat{y}=1 | z=1, y=1)}\right)$$****

This PR deprecates the original `FairClassifier` and renames it to `DemographicParityClassifier`. Furthermore, it adds an `equal_opportunity_score` metric and `EqualOpportunityClassifier` model


sklego/linear_model.py
```diff
@@ -231,4 +189,150 @@ def _add_intercept(self, X):
...
+
+        multiclass_meta = {"ovr": OneVsRestClassifier, "ovo": OneVsOneClassifier}[
+            multi_class
+        ]
+        return multiclass_meta(
+            _DemographicParityClassifer(*args, **kwargs), n_jobs=n_jobs
+        )
+
+
+@deprecated(
```
koaning: the fairclassifier is depricated right, not the demographic parity one?

sklego/linear_model.py
```diff
@@ -231,4 +189,150 @@ def _add_intercept(self, X):
...
+
+    @classmethod
+    def _get_param_names(cls):
+        return sorted(
+            super(_DemographicParityClassifer, cls)._get_param_names()
+            + _FairClassifier._get_param_names()
+        )
+
+
+class EqualOpportunityClassifier(BaseEstimator, LinearClassifierMixin):
```
koaning: nice!

tests/test_estimators/test_demographic_parity.py
```diff
@@ -4,7 +4,7 @@
 from sklearn.linear_model import LogisticRegression
 
 from sklego.common import flatten
-from sklego.linear_model import FairClassifier
+from sklego.linear_model import DemographicParityClassifier
```
koaning: does it make sense to confirm that we have a deprecation warning of the deprecated model?

doc/fairness.ipynb
```diff
@@ -115,11 +115,162 @@
    "source": [
```
koaning: image

the latex does not seem to render on the page? might be a github rendering issue but worthwhile to check.

sklego/linear_model.py
MBrouns: Thanks for catching that!

tests/test_estimators/test_demographic_parity.py
MBrouns: Yes, will add that

doc/fairness.ipynb
MBrouns: it seems to be a github issue as when I `make docs` it renders fine

MBrouns:
So in contrast with the DemographicParityClassifier, this approach wasn't published so there is no formal benchmark run anywhere. It might be worth doing that before merging.

tests/test_estimators/test_demographic_parity.py
```diff
@@ -0,0 +1,113 @@
...
+
+    sensitive_cols = [0]
+    X_without_sens = np.delete(X, sensitive_cols, axis=1)
+    lr = LogisticRegression(penalty="none", solver="lbfgs")
+    fair = DemographicParityClassifier(
+        covariance_threshold=None, sensitive_cols=sensitive_cols, penalty="none"
+    )
+    try:
+        fair.fit(X, y)
+    except SolverError:
```
koaning: is there a good reason for this? something tells me that there should be no `SolverError` for the `random_xy_dataset_clf` datasets

koaning:
@MBrouns whats the status on this? it is likely that I'll do a round of merges tomorrow, want this to go in?

tests/test_estimators/test_demographic_parity.py
MBrouns: Absolutely right! I removed it

Issue #201, opened 2019-09-20
[FEATURE] adding a prior ... it might be a good meta


koaning:
Sometimes it makes sense to add a prior to your classifier. If you know that your dataset is unbalanced for example.

koaning:
1. I think `ClassifierPrior` might be slightly better as a name but I am open to suggestions. 
2. This sounds good.
3/4. Why would you need to store/use the confusion matrix? 

This sounds a lot like the [confusion balancer](https://scikit-lego.readthedocs.io/en/latest/meta.html#Confusion-Balancer) that I was working on Friday. The impression I got from your idea is that you don't check the confusion matrix but instead use the `prior` parameter that is passed apon initialising the `ClassifierPrior`. This is a very valid idea when your train data is expected to not represent reality. 

Was this not the original idea or did I misunderstand the idea?

jsamoocha:
About the confusion matrix: assume a binary classifier that predicts _Y_ (having classes _a_ and _b_). Say the inner classifier predicts _Y_hat_ = _a_. If we're interested in P(Y=a|Y_hat=a), applying Bayes rule, we need to know P(Y_hat=a|Y=a) (i.e. the True Positive Rate if "a" is the "positive" prediction) and P(Y_hat=a) (i.e. all "a" predictions / all predictions). Don't you need the confusion matrix to compute that likelihood and marginal?

koaning:
It is a shame that github does not render mathjax, but here goes. 

![image](https://user-images.githubusercontent.com/1019791/65637976-78e09600-dfe5-11e9-8f5e-3f2c1442804b.png)

This does make a bit of sense actually. I originally thinking you just wanted to reweigh the classes by saying one of them is more likely. You don't need a confusion matrix for that. Would also be a nice feature.

jsamoocha:
Ok, let me do both options, then. This means the classifier gets an additional param `evidence` that can have values `predict_proba`, `confusion_matrix`, or `both`.

In case of `predict_proba`, the prediction result will be a balance(*) between the inner estimator's `predict_proba()` result and the prior probabilities.

In case of `confusion_matrix`, the result will use the inner estimator's `predict()`, confusion matrix and Bayes Rule as described above.

In case of `both`, we'll need to find a probabilistically sound balance between the 2 approaches (WIP).

(*) As for balance, an additional parameter `alpha` could be added, similar to your approach in the confusion balancer?

For the class name, how about `SubjectiveClassifier`?

[...]

PR #213, opened 2019-10-02
subjective classification

jsamoocha:
Work-in-progress on a subjective meta classifier, that uses a prior class distribution to compute posterior probabilities of predicted classes (#201 ).


sklego/meta.py
```diff
@@ -537,3 +537,33 @@ def predict(self, X):
...
+
+class SubjectiveClassifier(BaseEstimator, ClassifierMixin, MetaEstimatorMixin):
+    def __init__(self, estimator, prior):
+        self.estimator = estimator
+        self.prior = prior
+
+    def fit(self, X, y):
+        X, y = check_X_y(X, y, estimator=self.estimator, dtype=FLOAT_DTYPES)
+        if not isinstance(self.estimator, ClassifierMixin):
+            raise ValueError("The SubjectiveClassifier meta model only works on classifcation models")
```
koaning: spelling

koaning:
Is there a reason why you did not implement `predict_proba()`?

koaning:
I am also a little bit confused by the implementation. The way I interpreted it, the estimated probabilty (y_hat in the formula below), is given by `self.estimator.predict(X)`. The prior is p(y_a) and the confusion matrix determines p(y_hat_a | y_a).  



Did I get this idea wrong?

jsamoocha:
> I am wondering if it is easier to have the internals be pandas dataframes ... it might be better if you keep numpy arrays because then you can multiply matrices with `@`.

There doesn't seem to be a need for matrix multiplication in this case. The `DataFrame` allows easier lookups for non-numeric classes using `index` and `columns`.

> Is there a reason why you did not implement `predict_proba()`?

TODO - this is all WIP. Just wanted to quickly share the core principle to avoid any confusion. 


> I am also a little bit confused by the implementation. The way I interpreted it, the estimated probabilty (y_hat in the formula below), is given by `self.estimator.predict(X)`. The prior is p(y_a) and the confusion matrix determines p(y_hat_a | y_a).
> 
> 
> 
> Did I get this idea wrong?

For a binary classifier, if you expand the denominator p(y_hat_a), it becomes p(y_hat_a | y_a)*p(y_a) + p(y_hat_a | y_b)*p(y_b). The `_evidence()` method computes this denominator for the general case with _n_ classes.

koaning:
check that errors go to .fit 
use self.classes_ in favor of self.estimator.classes_
check classes attr in first test -> classes_
add standard tests
docstring

koaning:
Merged as an experimental feature. Won't add documentation for it yet.

koaning:
@jsamoocha thanks! I think this is an interesting idea to keep exploring. I've added it to the repo for now. Let's ping back later. I will also close all related issues.

PR #215, opened 2019-10-11
`definitely` typo - Sorry if it's very minor

amrrs:


MBrouns:
Thanks for the PR! Don't worry about it being a minor change, it's good to fix typos like this :)

There seems to be something with the GH actions that aren't working as expected. I'll try to fix these and merge afterwards

amrrs:
Thank you 😊

amrrs:
@MBrouns just to let you know that I had referenced your function in my recent presentation at pycon https://speakerdeck.com/amrrs/machine-learning-bias-4a462ffe-e38e-4f56-9fcd-ce883d748f12?slide=32

koaning:
@amrrs thanks for that! also ... @MBrouns it looks like we might need to think about stuff like deprecation now

PR #223, opened 2019-10-18
[wip] this is how the repo looks with black.

MBrouns:


koaning:
not bad, but i daresay we should first merge AllTheThings[tm] because black will cause a merge conflict.

MBrouns:
Agreed. I do think it improved the state overall though so I'm open to doing this!

koaning:
coolio.

koaning:
We might need to expand the "contributing" section as well if we want to add a pre-comit hook.

PR #230, opened 2019-10-25
fix projection image in fairness docs

### LINKED ISSUE #217: [DOCS] fairness img broken
projection image is broken in docs. 

image

MBrouns:
fixes #217

### DIFF
--- a/doc/fairness.ipynb
+++ b/doc/fairness.ipynb
@@ -274,7 +274,7 @@
     "\n",
     "The `InformationFilter` uses a variant of the [gram smidt process](https://en.wikipedia.org/wiki/Gram%E2%80%93Schmidt_process) to filter information out of the dataset. We can make it visual in two dimensions;\n",
     "\n",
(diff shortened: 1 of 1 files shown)


doc/fairness.ipynb
```diff
@@ -274,7 +274,7 @@
     "\n",
     "The `InformationFilter` uses a variant of the [gram smidt process](https://en.wikipedia.org/wiki/Gram%E2%80%93Schmidt_process) to filter information out of the dataset. We can make it visual in two dimensions;\n",
     "\n",
-    "![proj_img](_static/projections.png)\n",
+    "![proj«img](_static/projections.png)\n",
```
koaning: interesting. any context on why this character causes the fix?

MBrouns: I think the character you're seeing is githubs' way of indicating a character removal. I only removed the underscore from the alt text

PR #231, opened 2019-10-25
Add conda installation instruction

maxibor:
Conda install intruction now that scikit-lego is available through conda-forge.
Might want to wait for the conda package to be available (few min-hours maybe) to merge this PR and close #211 .

PR #234, opened 2019-10-28
Conda

MBrouns:


MBrouns:
This branch only exists so that #231 will have a proper build status and we can merge that

PR #239, opened 2019-10-29
v1 of this

koaning:
this is v1 ... but also to test appveyor.


sklego/this.py
```diff
@@ -0,0 +1,14 @@
+poem = """
+defer responsibility to the end user,
+this might just empower them.
+all the features and every extension,
+the path towards ruin is paved with good intensions.
```
MBrouns: Small typo -> intentions

sklego/this.py
```diff
@@ -0,0 +1,21 @@
+poem = """
+roses are red, violets are blue,
```
MBrouns: love it!

PR #244, opened 2019-12-02
implement get_feature_names on pandas transformers

MBrouns:
implementing `get_feature_names` allows for easier interoperability with libraries such as ELI5. For now, I implemented it only on the transformers which work on Pandas objects although I think it should work on the others as well

### DIFF
--- a/sklego/preprocessing.py
+++ b/sklego/preprocessing.py
@@ -174,10 +174,6 @@ class ColumnSelector(BaseEstimator, TransformerMixin):
     """
 
     def __init__(self, columns: list):
-        # if the columns parameter is not a list, make it into a list
-        if not isinstance(columns, list):
-            columns = [columns]
-
         self.columns = columns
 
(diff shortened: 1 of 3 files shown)


sklego/preprocessing.py
```diff
@@ -200,6 +201,9 @@ def transform(self, X):
             return X[self.columns]
         return X
 
+    def get_feature_names(self):
```
koaning: seems fair, but i wonder ... any reason why this cannot be a property?

MBrouns: I agree that a property would be prettier, but scikit-learn does it like this in several other places as well (https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.TfidfVectorizer.html#sklearn.feature_extraction.text.TfidfVectorizer.get_feature_names)

koaning: Fair enough. :)

Issue #249, opened 2020-01-03
Code Correction and enhancement in some of the preprocessing classes


greatsharma:
I suggests the following changes to preprocessing classes -

1.  Originally In transform method of PandasTypeSelector, 
	_check_X_for_type(X) is called after checking dtypes. By this 
	if we call transform method on numpy array then it raise
        `AttributeError: 'numpy.ndarray' object has no attribute dtypes`
	
	but _check_X_for_type(X) should be called before checking dtypes,
	and then calling transform on numpy array raises
       `TypeError: Provided variable X is not of type pandas.DataFrame`

	which is the suitable and correct error message to user.

2. In class ColumnSelector - 
	- The following if statement should be moved from fit to init method
                `if not isinstance(self.columns, list):
                        self.columns = [self.columns] `
	
	 this is done for two reasons first to make the code consisent with 
	 other similar classes(in this case ColumnDroper class). And second
	 because this checking is suitable during class instantiation and
	 has nothing specific to fit method.

 	-  call _check_X_for_type(X) in the transform method, because this
	   class is dedicated to DF only and checking X type in fit is not
	   sufficient because if we fit a DF but try to transform an array
	   then it throws,
		`IndexError: only integers, slices (`:`), ellipsis (`...`),
                    numpy.newaxis (`None`) and integer or boolean
                    arrays are valid indices`

	  but the more suitable and correct error would be,
		`TypeError: Provided variable X is not of type pandas.DataFrame`

3. In class OrthogonalTransformer - 
	let's take a df with columns ['a', 'b', 'c', 'd']

	- Originally if we fit df[['a', 'b']] but transform df[['c', 'd']]
	  or df[['b', 'a']] then it just runs the code, but the returned 
	  results will be incorrect because we fit and transform 2 DFs with
	  different columns. But we need not to be worried in case of numpy
	  arrays as there is nothing like column names.

	Such transformation should be handled with raising following error, 
	`KeyError: Columns were not equal during fit and transform`

4. In class ColumnDroper -
	- In transform method call check_is_fitted otherwise
         get_feature_names() will throw attribute error if we call it on
	  object which is transformed without fitting.

	- In transform method call check_X_for_type() for the same reason
	  I explained for ColumnSelector class above.

koaning:
1. Happy new year!
2. Thanks, you really seem to put effort in here.
3. Could you split these points into separate issues? Might make it easier to keep the discussion clean.

PR #250, opened 2020-01-03
PR resolving issue 249

greatsharma:
In this PR I resolved the issue #249 and also added test for ColumnSelector class


sklego/preprocessing.py
```diff
@@ -100,13 +100,13 @@ def transform(self, X):
         :param X: pandas dataframe to select dtypes for
         """
         check_is_fitted(self, ['type_columns_', 'X_dtypes_', 'feature_names_'])
+        self._check_X_for_type(X)
```
greatsharma: Originally In transform method of PandasTypeSelector, _check_X_for_type(X) is called after checking dtypes. By this if we call transform method on numpy array then it raise
`AttributeError: 'numpy.ndarray' object has no attribute dtypes`

but _check_X_for_type(X) should be called before checking dtypes,
and then calling transform on numpy array raises
`TypeError: Provided variable X is not of type pandas.DataFrame`

which is the suitable and correct error message to user.

sklego/preprocessing.py
```diff
@@ -174,6 +174,10 @@ class ColumnSelector(BaseEstimator, TransformerMixin):
     """
 
     def __init__(self, columns: list):
+        # if the columns parameter is not a list, make it into a list
```
greatsharma: The following if statement should be moved from fit to init method
   ` if not isinstance(self.columns, list):`
    `     self.columns = [self.columns]`

this is done for two reasons first to make the code consisent with
other similar classes(in this case ColumnDroper class). And second
because this checking is suitable during class instantiation and
has nothing specific to fit method.

sklego/preprocessing.py
```diff
@@ -197,6 +198,7 @@ def transform(self, X):
         :param X: ``pd.DataFrame`` on which we apply the column selection
         :returns: ``pd.DataFrame`` with only the selected columns
         """
+        self._check_X_for_type(X)
```
greatsharma: call _check_X_for_type(X) in the transform method, because this class is dedicated to DF only and checking X type in fit is not sufficient because if we fit a DF but try to transform an array then it throws,
    `IndexError: only integers, slices (:), ellipsis (...), numpy.newaxis (None) and integer 
     or boolean arrays are valid indices`

but the more suitable and correct error would be,
`TypeError: Provided variable X is not of type pandas.DataFrame`

sklego/preprocessing.py
```diff
@@ -421,6 +426,10 @@ def fit(self, X, y=None):
 
     def transform(self, X):
         """Transforms X using the fitted inverse of R. Normalizes the result if required"""
+        if type(X) is pd.core.frame.DataFrame:
```
greatsharma: handling bad transformation of OrthogobalTransformer mentioned in #249

sklego/preprocessing.py
```diff
@@ -604,6 +613,8 @@ def transform(self, X):
         :param X: ``pd.DataFrame`` on which we apply the column selection
         :returns: ``pd.DataFrame`` with only the selected columns
         """
+        check_is_fitted(self, ['feature_names_'])
```
greatsharma: to avoid error on calling get_feature_name on non fitted object

sklego/preprocessing.py
```diff
@@ -604,6 +613,8 @@ def transform(self, X):
...
+        self._check_X_for_type(X)
```
greatsharma: handling transformation on numpy array

sklego/preprocessing.py
greatsharma: Originally  _check_X_for_type(X) is called after checking dtypes. By this if we call transform method on numpy array then it raise
`AttributeError: 'numpy.ndarray' object has no attribute dtypes`

but _check_X_for_type(X) should be called before checking dtypes, and then calling transform on numpy array raises,
`TypeError: Provided variable X is not of type pandas.DataFrame`

which is the suitable and correct error message to user.

sklego/preprocessing.py
```diff
@@ -174,6 +174,10 @@ class ColumnSelector(BaseEstimator, TransformerMixin):
     """
 
     def __init__(self, columns: list):
+        # if the columns parameter is not a list, make it into a list
+        if not isinstance(columns, list):
+            columns = [columns]
+
```
greatsharma: The following if statement is moved from fit to init method. This is done for two reasons first to make the code consisent with other similar classes(in this case ColumnDroper class). And second because this checking is suitable during class instantiation and has nothing specific to fit method.

sklego/preprocessing.py
greatsharma: call _check_X_for_type(X) in the transform method, because this class is dedicated to DF only and checking X type in fit is not sufficient because if we fit a DF but try to transform an array then it throws,
    `IndexError: only integers, slices (:), ellipsis (...), numpy.newaxis (None) and integer 
     or boolean arrays are valid indices`

but the more suitable and correct error would be,
`TypeError: Provided variable X is not of type pandas.DataFrame`

sklego/preprocessing.py
```diff
@@ -421,6 +426,10 @@ def fit(self, X, y=None):
 
     def transform(self, X):
         """Transforms X using the fitted inverse of R. Normalizes the result if required"""
+        if type(X) is pd.core.frame.DataFrame:
+            if not self.columns_ == list(X.columns):
+                raise KeyError('Columns were not equal during fit and transform')
+
```
greatsharma: handling bad transformation of OrthogobalTransformer mentioned in #249

sklego/preprocessing.py
greatsharma: to avoid error on calling get_feature_name on non fitted object

sklego/preprocessing.py
greatsharma: handling transformation on numpy array

sklego/preprocessing.py
MBrouns: Normally I would agree but this isn't how scikit-learn works. When an estimator is cloned the attributes are set directly, so this overrides whatever happens in the `__init__`. Therefore the convention is to do all parameter validation and modification inside the fit method

greatsharma: > Normally I would agree but this isn't how scikit-learn works. When an estimator is cloned the attributes are set directly, so this overrides whatever happens in the `__init__`. Therefore the convention is to do all parameter validation and modification inside the fit method

In that case we need to do the same thing in ColumnDroper class so to make the code consistent otherwise readers get confused why the same thing is done in 2 different ways (as I got when I first saw it), this may miss-lead some of them.

In either case we should change one of the class and make the code same for same functionality

MBrouns: Agreed, if `ColumnDropper` does it the other way around atm that is a bug.

sklego/preprocessing.py
MBrouns: This seems like something that we should either do on **all** transformers, or on None. I don't think adding this on all transformers is the way to go.

koaning: In particular ... the end user is responsible for doing sensible things with our library. If you're gonna fit on a dataset that does not resemble what you're gonna predict on then this is something that we cannot prevent either way.

tests/test_preprocessing/test_columnselector.py
```diff
@@ -0,0 +1,59 @@
...
+def test_select_one(df):
+    result_df = ColumnSelector(['e']).fit_transform(df)
+    expected_df = pd.DataFrame({
+        "e": [0, 1, 0, 1, 0, 1]})
+
+    assert_frame_equal(result_df, expected_df)
+
+
+def test_select_none(df):
+    result_df = ColumnSelector([]).fit_transform(df)
```
MBrouns: Shouldnt this just be a `ValueError`? I can't imagine a use-case atm and the next step in the pipeline will give an error, while the cause is in this step.

MBrouns:
Thanks for the pull request! There's some good changes in it

MBrouns:
Also, just out of curiosity: How did you come across `scikit-lego`?

greatsharma:
> Also, just out of curiosity: How did you come across `scikit-lego`?

I watched @koaning talk Untitled12.ipynb at pydata  and get to know about sklego. But the fun fact is before even knowing such thing exists I already created some custom transformer built over sklearn and when I visited this repo I found that coincidentally some of your class names are same as mine so I think this is the right place to contribute because whatever you are solving is the daily suffering and pain of all data scientists. And this library is really great and painkiller.

sklego/preprocessing.py
greatsharma: Hmm, I agreed.

tests/test_preprocessing/test_columnselector.py
greatsharma: In that case we need to throw the same error in ColumnDroper if someone drops all columns from the dataframe.
Either throw `ValueError` in both cases or in none.

greatsharma:
> I just sat next to @MBrouns reviewing this. First off, there's lot of good ideas in here. Thanks for the review! There are a few details worth discussing further however. We've resolved everything we deem as merge-able please reply on the open topics.
> 
> In the future however; it might be better to split the PR an the issue up into separate concerns. That makes the reviewing (and therefore merging) a lot easier for us, the maintainers.

OK, I will take care of that

tests/test_preprocessing/test_columnselector.py
greatsharma: @MBrouns should I throw VlaueError in both cases?

MBrouns: I would say to add something along these lines to fit for the columnselector:

```python
if len(self.columns) == 0:
    raise ValueError("expected `columns` to be at least of length 1, found length of 0 instead")
```

and for the columndropper:

```python
if len(self.feature_names_) == 0:
   raise ValueError(f"using the ColumnDropper for dropping {self.columns} would result in an empty output dataset ")
```

greatsharma: @MBrouns I added this check undef _check_column_length(), please review

greatsharma: Please review this so that I can move ahead in this PR if in case anything else is to be done

koaning: Don't worry, we will have a look but it is fair to point out that we're all volunteers here and haste makes waste. Also ... it's dinnertime here and I will cook for my household in a moment so this might need to wait until later.

greatsharma: I will surely take care of that from now :)

sklego/preprocessing.py
```diff
@@ -403,6 +405,9 @@ def fit(self, X, y=None):
         Store the inverse of R of the QR decomposition of X, which can be used to calculate the orthogonal projection
         of X. If normalization is required, also stores a vector with normalization terms
         """
+        if type(X) is pd.core.frame.DataFrame:
```
koaning: preferable: `isinstance(X, pd.DataFrame)`

sklego/preprocessing.py
koaning: here too `isinstance(X, pd.DataFrame)`

sklego/preprocessing.py
koaning: This bit is still in the codebase. 

```
            if not self.columns_ == list(X.columns):
                raise KeyError('Columns were not equal during fit and transform')
```

tests/test_preprocessing/test_columnselector.py
```diff
@@ -0,0 +1,59 @@
...
+    assert_frame_equal(result_df, df)
```
koaning: shouldn't this give a ValueError now?

greatsharma: I already changed it these images are pointing to the previous one.

sklego/preprocessing.py
greatsharma: I removed these things. These images are pointing to previous one

sklego/preprocessing.py
greatsharma: These are no more in code

sklego/preprocessing.py
```diff
@@ -174,6 +175,10 @@ class ColumnSelector(BaseEstimator, TransformerMixin):
     """
 
     def __init__(self, columns: list):
+        # if the columns parameter is not a list, make it into a list
+        if not isinstance(columns, list):
```
koaning: this is a detail. but wouldn't this be a usecase for the `as_list` function we have in `sklego.common`?

koaning:
I've posted my final detail. @MBrouns unless you see something I missed, feel free to click the green button.

sklego/preprocessing.py
koaning: This is my only issue now.

greatsharma: @koaning I just readed `as_list` function and I agree with you. I did the changes you can now review them.

MBrouns:
Considering the amount of commits in this PR + some back and forth on different implementations, it might be nice to do a rebase and squash some commits before merging this. @greatsharma have you rebased before, if not I'll walk you through it

greatsharma:
@MBrouns all I know about git rebase is, 
`git checkout my_branch_1`
`git rebase master`
`git checkout master`
`git merge my_branch_1`

But I never worked on this before. If I am wrong or missing something or should have known something more about `rebase` then I would really love if you help me in this.

koaning:
github  does squash too. will merge.

koaning:
For the future; please split up into separate PRs. For now; thanks for the PR :)

PR #255, opened 2020-01-03
added the arrests dataset

koaning:
I've added only the arrests dataset since the other one is a tad bit too large.


sklego/datasets.py
```diff
@@ -4,6 +4,47 @@
...
+    - Personal communication from Michael Friendly, York University.
+
+    The documentation page of the dataset from the package can be viewed here:
+    http://vincentarelbundock.github.io/Rdatasets/doc/carData/Arrests.html
+    """
+    filepath = resource_filename("sklego", os.path.join("data", "arrests.csv"))
+    df = pd.read_csv(filepath)
+    if give_pandas:
+        return df
+    if give_pandas:
```
MBrouns: the `if give_pandas` check seems duplicated?

koaning: ah yeah. did some copy/pasting here.

koaning:
I think in the future we can consider arrow but I believe today it is an external dep. I'll play with gzip and see if that helps. In the future another option might be that we consider the datasets as a seperate dependency. Something like this might be good; 

```
pip install sklego[data]
```

koaning:
Just zip seems ... fine? 

```
drwxr-xr-x  10 user  staff   340B Jan  4 17:52 .
-rw-r--r--   1 user  staff    58K Jan  4 17:52 abalone.zip
-rw-r--r--   1 user  staff    18K Jan  4 17:52 arrests.zip
-rw-r--r--   1 user  staff   2.1K Jan  4 17:52 chickweight.zip
-rw-r--r--   1 user  staff   1.3K Jan  4 17:52 heroes.zip
drwxr-xr-x  20 user  staff   680B Jan  4 17:50 ..
-rw-r--r--   1 user  staff   167K Jan  4 17:48 arrests.csv
-rwxr-xr-x   1 user  staff   187K Jan  4 17:47 abalone.csv
-rw-r--r--   1 user  staff   6.2K Jan  4 17:47 chickweight.csv
-rw-r--r--   1 user  staff   3.1K Jan  4 17:47 heroes.csv
```

PR #256, opened 2020-01-03
added some shields... fancy!

koaning:
it might be overkill, but it seems nice that this is all tracked.


koaning:
This is the rich diff. 

image

MBrouns:
LGTM!

PR #259, opened 2020-01-04
Remove redundant if condition

greatsharma:
Sorry, I am not creating an issue first because it's just oneliner.

In predict() of RandomRegressor, the following if always evaluates to False hence is redundant,

    if self.strategy not in self.allowed_strategies:
            raise ValueError(f"strategy {self.strategy} is not in {self.allowed_strategies}")

This is because it has been already checked in fit method and their is no way to predict without fitting.


koaning:
Yep. Nice one. Thanks! 

The test failure was due to a connection mishap on travis' side. I'll run it again and merge if it is green.

greatsharma:
> Yep. Nice one. Thanks!
> 
> The test failure was due to a connection mishap on travis' side. I'll run it again and merge if it is green.

Can we add some other baseline models like BaselineDropnaRow which takes the data and an estimator and then first transform the data by removing all nan rows and then fitting the estimator passed as parameter. And if in case every row contains a nan value then we can even raise error. And similarly for BaselineDropnaColumn

koaning:
> Can we add some other baseline models like BaselineDropnaRow which takes the data and an estimator and then first transform the data by removing all nan rows and then fitting the estimator passed as parameter. And if in case every row contains a nan value then we can even raise error. And similarly for BaselineDropnaColumn 

Please add an issue for this and if possible, write a simple proposal.

PR #267, opened 2020-01-16
move list coersions out of init methods

MBrouns:


koaning:
Are we sure these are the only instances? Either way, this is a good change.

PR #273, opened 2020-01-24
Adding Verbosity to GMM Classifiers

arose13:
The bayesian classifiers have verbosity the mle version does not.

Perhaps also adding `**gmm_kwargs` would permanently solve this problem.

### DIFF
--- a/sklego/mixture.py
+++ b/sklego/mixture.py
@@ -225,6 +225,8 @@ class GMMOutlierDetector(OutlierMixin, BaseEstimator):
         precisions_init=None,
         random_state=None,
         warm_start=False,
+        verbose=0,
+        verbose_interval=10
     ):
         self.threshold = threshold
         self.method = method
@@ -242,6 +244,8 @@ class GMMOutlierDetector(OutlierMixin, BaseEstimator):
         self.precisions_init = precisions_init
         self.random_state = random_state
         self.warm_start = warm_start
+        self.verbose = verbose
+        self.verbose_interval = verbose_interval
 
     def fit(self, X: np.array, y=None) -> "GMMOutlierDetector":
         """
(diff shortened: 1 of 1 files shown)


koaning:
some of the `**gmm_kwargs` tricks won't fly in scikit-learn's testing framework so I'd prefer to not go that far. adding verbose mode seems perfectly fine tho.

MBrouns:
varargs or varkwargs in the `__init__` are indeed not allowed in `scikit-learn` as the `clone` implementation (used in gridsearch for example) uses function definition inspection of the `__init__` to figure out what parameters need to be copied over to new instances. If you use varargs or varkwargs, you can't name the individual attributes anymore and cloning will probably result in some parameters not being set correctly

koaning:
CI also seems to be failing. Internally it looks as if an `int` accidentally got converted to a `tuple`? 

```
GaussianMixture(covariance_type='full', init_params='kmeans', max_iter=100,

                means_init=None, n_compone...ne, reg_covar=1e-06, tol=0.001, verbose=(0,),

                verbose_interval=10, warm_start=False, weights_init=None)
```

@arose13 do you also get this result locally?

arose13:
@koaning I didn't get that error but I'm running Python 3.7 so maybe that's why. I'm creating a virtual environment now and rerunning the tests to see if I can get it to break.

PS: @koaning @MBrouns didn't know about that cloning stuff. Interesting I'll say.

koaning:
found the error. note the comma at the end; 

```
self.verbose = verbose,
```

koaning:
The code formatting is complaining now. If you run this locally, what happens? 

```
make develop
make check
make black
```

arose13:
I've only been running the tests 1 by 1. It looks like black has failed but I'll run those and see what's up

koaning:
LGTM.

@arose13 thanks for the PR! 

One thing worth mentioning; we just did a release last week so it might take a bit before this is on pypi. If this is a change you need sooner rather than later know that you can pip install via git too.

arose13:
@koaning understood and thanks for the merge

PR #275, opened 2020-01-26
interval-encoder

koaning:
This is the first variant of the interval encoder.


koaning:
I’ve added three things here. 

- a first implementation of non-constrained interval encoder 
- light documentation on it 
- while i was added, i added missing documentation for the RBF features

koaning:
I did just realise that by merging this we might need to do a release because of the documentation.

tests/test_preprocessing/test_interval_encoder.py
```diff
@@ -0,0 +1,43 @@
...
+            estimator_checks.check_pipeline_consistency,
+            estimator_checks.check_estimators_nan_inf,
+            estimator_checks.check_estimators_overwrite_params,
+            estimator_checks.check_estimator_sparse_data,
+            estimator_checks.check_estimators_pickle,
+        ]
+    ),
+)
+def test_estimator_checks(test_fn):
+    test_fn(IntervalEncoder.__name__, IntervalEncoder(n_chunks=2))
```
MBrouns: Should we also test for `n_chunks=1` to better cover the extremities?

sklego/preprocessing.py
```diff
@@ -9,6 +9,94 @@
...
+    :param n_chunks: the number of cuts that makes the interval
+    :param method: the interpolation method used
+    :param span: a hyperparameter for the interpolation method, if the
+    method is `normal` it resembles the width of the radial basis
+    function used to weigh the points
+    """
+
+    def __init__(self, n_chunks=10, span=1, method="normal"):
+        allowed_methods = ["average", "normal"]
+        if method not in allowed_methods:
```
MBrouns: This should go to fit

sklego/preprocessing.py
```diff
@@ -9,6 +9,94 @@
...
+            raise ValueError(f"`method` must be in {allowed_methods}, got `{method}`")
+        self.span = span
+        self.method = method
+        self.n_chunks = n_chunks
+
+    def fit(self, X, y):
+        """Fits the estimator"""
+        # these two matrices will have shape (columns, quantiles)
+        # quantiles indicate where the interval split occurs
```
MBrouns: Do we need additional non-negativity validation on span and n_chunks?

sklego/preprocessing.py
```diff
@@ -9,6 +9,94 @@
 from sklego.common import TrainOnlyTransformerMixin, as_list
 
 
+def _mk_average(xs, ys, intervals, span=1, method="average"):
+    """
+    asdafa
```
MBrouns: This probably shouldn't be there

tests/test_preprocessing/test_interval_encoder.py
koaning: It will result in the mean ... but that makes it a good test. Noted!

sklego/preprocessing.py
koaning: oops. i need to read somebody's blogpost.

sklego/preprocessing.py
koaning: strange ... the spellcheck should've picked it up.

sklego/preprocessing.py
koaning: That feature is not built yet. Once there, yes.

tests/test_preprocessing/test_interval_encoder.py
koaning: added.

koaning:
I've now also added some extra tests that check that if you give silly input that it will throw a `ValueError` during `.fit()`.

PR #278, opened 2020-01-26
Lowess

koaning:
I like the implementation as far as API goes. 

image


koaning:
But the performance is really slow. Note the effect of the span parameter. 

image

MBrouns:
The implementation looks good! There is a minor inconsistency in the Wording of the parameter validation exceptions though

koaning:
Could you rephrase what you mean with "Wording of the paramater validation exceptions"?

MBrouns:
Ah sure, 

```
if self.span is not None:
            if not 0 <= self.span <= 1:
                raise ValueError(f"we must have 0 <= span <= 1, got: {self.span}")
        if self.sigma < 0:
            raise ValueError(f"sigma must be >= 0, got: {self.sigma}")
```

The errors are phrased slightly differently.

koaning:
I completely forgot about this branch. May wrap it up tonight.

MBrouns:
this looks almost finished right? Just need to add tests?

koaning:
And docs. But this is something I'd love to pick up today.

koaning:
The docs will get fancy gifs. 

image

koaning:
I think this is it. Once we merge this we'll also need to push to pypi though because the docs are going to list a new method.

PR #283, opened 2020-02-06
Bayesian Kernel Density Classifier

arose13:
just opening the pull request.

Ignore the `if __name__ == 'main':` block that won't be there in the end


sklego/neighbors.py
```diff
@@ -0,0 +1,93 @@
+import numpy as np
+
+from sklearn.base import BaseEstimator, ClassifierMixin
+from sklearn.neighbors import KernelDensity
+from sklearn.utils import check_X_y
+from sklearn.utils.multiclass import unique_labels
+from sklearn.utils.validation import check_is_fitted, check_array, FLOAT_DTYPES
+
+
+# noinspection PyPep8Naming,PyAttributeOutsideInit
```
koaning: what is this comment for?

sklego/neighbors.py
```diff
@@ -0,0 +1,93 @@
...
+class BayesianKernelDensityClassifier(BaseEstimator, ClassifierMixin):
+    """
+    Bayesian Classifier that uses Kernel Density Estimations to generate the joint distribution
+    Parameters:
+        - bandwidth: float
+        - kernel: for scikit learn KernelDensity
+    """
+    def __init__(self, bandwidth=0.2, kernel='gaussian'):
+        # TODO (2/6/2020) add all other sklearn kwargs
```
koaning: yes ... definitely would like to see that :)

sklego/neighbors.py
```diff
@@ -0,0 +1,93 @@
...
+        self.bandwidth = bandwidth
+        self.kernel = kernel
+
+    def fit(self, X: np.ndarray, y: np.ndarray):
+        # Checks
```
koaning: We appreciate comments, but in this case the comments feel over-used. The code should explain **what** is happening while the comment should explain **why** it is happening in a certain way.

sklego/neighbors.py
```diff
@@ -0,0 +1,93 @@
...
+        X = np.expand_dims(X, 1) if X.ndim == 1 else X  # type: np.ndarray
+
+        # Training
+        self.classes_ = unique_labels(y)
+        self.models_, self.priors_logp_ = {}, {}
+        for target_label in self.classes_:
+            selector = y == target_label
+            x_subset, y_subset = X[selector], y[selector]
+
+            # Joint distribution
```
koaning: the comment `setting joint distribution` feels fine here

sklego/neighbors.py
```diff
@@ -0,0 +1,93 @@
...
+            [self.models_[target_label].score_samples(X) for target_label in self.classes_]
+        ).T
+
+        log_likelihood_and_prior = np.exp(log_likelihood + log_prior)
+        posterior = log_likelihood_and_prior / log_likelihood_and_prior.sum(axis=1, keepdims=True)
+        return posterior
+
+    def predict(self, X):
+        # Checks
+        self._check_is_fitted()
```
koaning: this is already done in `.predict_proba`. do we need to do this twice?

sklego/neighbors.py
```diff
@@ -0,0 +1,93 @@
...
+        X = check_array(X, estimator=self, dtype=FLOAT_DTYPES)
+
+        # Inference
+        return self.classes_[np.argmax(self.predict_proba(X), 1)]
+
+    def _check_is_fitted(self):
```
koaning: might be fine not to have this as a method. single line seems fine.

sklego/neighbors.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    def __init__(self, bandwidth=0.2, kernel='gaussian'):
+        # TODO (2/6/2020) add all other sklearn kwargs
+        self.bandwidth = bandwidth
+        self.kernel = kernel
+
+    def fit(self, X: np.ndarray, y: np.ndarray):
+        # Checks
+        X, y = check_X_y(X, y, estimator=self, dtype=FLOAT_DTYPES)
+        X = np.expand_dims(X, 1) if X.ndim == 1 else X  # type: np.ndarray
```
koaning: no need to have the comment there that explains the type

sklego/neighbors.py
```diff
@@ -0,0 +1,93 @@
...
+        X = check_array(X, estimator=self, dtype=FLOAT_DTYPES)
+
+        # Inference
+        return self.classes_[np.argmax(self.predict_proba(X), 1)]
+
+    def _check_is_fitted(self):
+        check_is_fitted(self, ['classes_', 'models_', 'priors_logp_'])
+
+
+if __name__ == '__main__':
```
koaning: This might be good for the documentation. It might be cleaner to handle some of the parts here in pandas though.

koaning:
It also seems like the style flags are being raised. 

Could you run? 

```
make flake
```

Also ... could you run: 

```
black --check sklego tests setup.py
```

We use the style guides of both `flake8` and `black`.

koaning:
@MBrouns do we want to have this in `neighbors` or maybe someplace else?

sklego/neighbors.py
MBrouns: I think `check_X_y` already has an `ensure_2d` option which does this

MBrouns:
`neighbors`makes sense I think

sklego/neighbors.py
arose13: I wasn't going to until I saw it in `GMMClassifier` and `BayesianGMMClassifier`. I definitely have no problem removing it.

MBrouns:
Also, if you incorporate @koaning's changes, the PR looks good I think!

sklego/neighbors.py
arose13: it suppresses automatic lint checkers in PyCharm since `sklearn` uses a few unpythonic practices. It won't be there in the final commit.

sklego/neighbors.py
arose13: kk sick

sklego/neighbors.py
arose13: Done son

sklego/neighbors.py
arose13: whoops, this was only for type hinting

sklego/neighbors.py
arose13: I suppose I wanted to avoid duplicating code. Are you sure you want this function removed?

MBrouns: It turns out you actually don't need to specify the attributes you're checking for in `check_is_fitted`. That means the line can be replaced with just `check_is_fitted(self)`. The main argument here against having it in a separate method that the cognitive burden of the duplication is probably lower than the cognitive burden of the layer of indirection introduced by the function. I know what `sklearn.utils.check_is_fitted` does, not necessarily what `self._check_is_fitted` does and how it is exactly different from the sklearn version

sklego/neighbors.py
arose13: good idea

koaning: We assume the most recent version of pandas though, since there's been a recent upgrade it might be good to ensure that.

sklego/neighbors.py
```diff
@@ -0,0 +1,101 @@
...
+        self.breath_first = breath_first
+        self.leaf_size = leaf_size
+        self.metric_params = metric_params
+
+    def fit(self, X: np.ndarray, y: np.ndarray):
+        X, y = check_X_y(X, y, estimator=self, dtype=FLOAT_DTYPES)
+
+        self.classes_ = unique_labels(y)
+        self.models_, self.priors_logp_ = {}, {}
+        for target_label in self.classes_:
```
MBrouns: Maybe not necessarily in this PR, but this seems easily parallelizable with something like joblib.

arose13: Very good point, I think that's a good idea. I just confirmed that sklearn's kernel density estimator doesn't use any parallel components which is good. In most cases, it will merely half the runtime but if you have a lot of classes it could be a great benefit.

tests/test_estimators/test_neighbor_classifier.py
```diff
@@ -0,0 +1,42 @@
...
+
+def test_trivial_classification():
+    # Two linearly separable mvn should have a 100% prediction accuracy
+    x = np.concatenate(
+        [np.random.normal(-10, 1, (100, 2)), np.random.normal(10, 1, (100, 2))]
+    )
+    y = np.concatenate([np.zeros(100), np.ones(100)])
+
+    model = BayesianKernelDensityClassifier().fit(x, y)
+    assert (
```
koaning: no need to add a string explaining the assert statement. the function name is enough here ...

koaning: it'll also fit on a single line

koaning:
It would be nice if you could add a small snippet with a picture for the documentation.

sklego/neighbors.py
```diff
@@ -0,0 +1,125 @@
...
+            where classes are ordered as they are in self.classes_.
+        """
+        check_is_fitted(self)
+        X = check_array(X, estimator=self, dtype=FLOAT_DTYPES)
+
+        log_prior = np.array(
+            [self.priors_logp_[target_label] for target_label in self.classes_]
+        )
+
+        log_likelihood = Parallel(n_jobs=min(cpu_count(), len(self.classes_)))(
```
MBrouns: `n_jobs` should probably be a hyperparameter that defaults to `min(cpu_count(), len(self.classes_)` but is still user-settable

koaning:
Once the `nbjobs` comment is in and once all the non-required code is gone ... we can merge and deploy a new version. Technically the documentation is something that can wait.

sklego/neighbors.py
arose13: fair

arose13:
> Once the `nbjobs` comment is in and once all the non-required code is gone ... we can merge and deploy a new version. Technically the documentation is something that can wait.

Btw, to run in parallel, I've used sklearn's joblib so that I wouldn't have to create an extra project requirement. The joblib submodule in sklearn, however, is deprecated and going to be removed in sklearn version 0.24

Just saying that as a note

MBrouns:
Is sklearn going to remove joblib as a dependency or is only the direct import deprecated? If it’s the latter I don’t have objections to adding it as a project requirement. What do you think @koaning

koaning:
the whole [persistance thing](https://scikit-learn.org/stable/modules/model_persistence.html#persistence-example) is gonna be reel hard without `joblib` for pickling … so if `sklearn` is dropping it … we prolly gotta think about having it … 

either way … this is something i probably want to do some reading on … it is outside of the scope of this PR

sklego/neighbors.py
```diff
@@ -0,0 +1,122 @@
...
+        self.kernel = kernel
+        self.algorithm = algorithm
+        self.metric = metric
+        self.atol = atol
+        self.rtol = rtol
+        self.breath_first = breath_first
+        self.leaf_size = leaf_size
+        self.metric_params = metric_params
+
+        if n_jobs is not None:
```
koaning: @MBrouns I think this needs to be tested in `.fit()` no?

tests/test_estimators/test_neighbor_classifier.py
```diff
@@ -0,0 +1,50 @@
...
+            estimator_checks.check_classifiers_train,
+            estimator_checks.check_supervised_y_2d,
+            estimator_checks.check_supervised_y_no_nan,
+            estimator_checks.check_estimators_unfitted,
+            check_shape_remains_same_classifier,
+        ]
+    ),
+)
+def test_estimator_checks(test_fn):
+    test_fn(BayesianKernelDensityClassifier.__name__, BayesianKernelDensityClassifier())
```
koaning: Could you add a `njobs=2` as well … just in case?

tests/test_estimators/test_neighbor_classifier.py
```diff
@@ -0,0 +1,50 @@
...
+    model = BayesianKernelDensityClassifier().fit(x, y)
+    assert (model.predict(x) == y).all()
+
+
+def test_n_jobs_param():
+    BayesianKernelDensityClassifier()  # Default params
+    BayesianKernelDensityClassifier(n_jobs=None)  # None specified
+
+    with pytest.raises(ValueError):
+        BayesianKernelDensityClassifier(n_jobs=1.23)  # No floats
```
koaning: I think the failure should present itself in `.fit()`.

sklego/neighbors.py
MBrouns: Yep!

MBrouns: Also, I just checked some other sklearn estimators that have n_jobs, and it seems like`None` actually means only 1 job, -1 means all possible cores

arose13: Thanks for doing all this research for me give me a sec to make those changes

arose13: I just found that (at least for RandomForests) no checks are done on the `n_jobs` argument at all. It seems that all negative ints use the cores

The only thing that appears to break the code is floats. Since using a float raises a really obscure error (I assume with how job lists get sliced?) I think floats should be the only thing I check for.

Ugly error with floats looks like this.
```python
Traceback (most recent call last):
  File "C:\Users\antho\Miniconda3\lib\site-packages\joblib\parallel.py", line 797, in dispatch_one_batch
    tasks = self._ready_batches.get(block=False)
  File "C:\Users\antho\Miniconda3\lib\queue.py", line 167, in get
    raise Empty
_queue.Empty

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:/Users/antho/.PyCharm2019.3/config/scratches/scratch_2.py", line 25, in <module>
    RandomForestClassifier(5000, n_jobs=1.23).fit(x, y)
  File "C:\Users\antho\Miniconda3\lib\site-packages\sklearn\ensemble\_forest.py", line 383, in fit
    for i, t in enumerate(trees))
  File "C:\Users\antho\Miniconda3\lib\site-packages\joblib\parallel.py", line 1004, in __call__
    if self.dispatch_one_batch(iterator):
  File "C:\Users\antho\Miniconda3\lib\site-packages\joblib\parallel.py", line 808, in dispatch_one_batch
    islice = list(itertools.islice(iterator, big_batch_size))
ValueError: Stop argument for islice() must be None or an integer: 0 <= x <= sys.maxsize.
```

MBrouns: I'm actually ok with ignoring floats. We can expect some level of competency for our users I would say. @koaning?

sklego/neighbors.py
```diff
@@ -0,0 +1,127 @@
...
+
+        log_prior = np.array(
+            [self.priors_logp_[target_label] for target_label in self.classes_]
+        )
+
+        if self.n_jobs is None:
+            n_jobs = 1
+        elif self.n_jobs == -1:
+            n_jobs = min(cpu_count(), len(self.classes_))
+        else:
```
MBrouns: Well, n_jobs could be set to 2 right?

koaning:
I'll do some small fixes on my end because it is simple. Should be merged in a few minutes. (No need for the back and forth)

koaning:
@arose13 I had a go at rewriting the tests as well. 

```
import numpy as np

import pytest

from sklego.neighbors import BayesianKernelDensityClassifier
from sklego.testing import check_shape_remains_same_classifier
from tests.conftest import nonmeta_checks, general_checks, estimator_checks


@pytest.mark.parametrize(
    "test_fn",
    flatten(
        [
            nonmeta_checks,
            general_checks,
            estimator_checks.check_classifier_data_not_an_array,
            estimator_checks.check_classifiers_one_label,
            estimator_checks.check_classifiers_classes,
            estimator_checks.check_classifiers_train,
            estimator_checks.check_supervised_y_2d,
            estimator_checks.check_supervised_y_no_nan,
            estimator_checks.check_estimators_unfitted,
            check_shape_remains_same_classifier,
        ]
    ),
)
def test_estimator_checks(test_fn):
    test_fn(BayesianKernelDensityClassifier.__name__, BayesianKernelDensityClassifier())
    test_fn(BayesianKernelDensityClassifier.__name__, BayesianKernelDensityClassifier(n_jobs=2))



@pytest.fixture
def trivial_example():
    x = np.concatenate(
        [np.random.normal(-10, 1, (100, 2)), np.random.normal(10, 1, (100, 2))]
    )
    y = np.concatenate([np.zeros(100), np.ones(100)])
    return x, y

def test_trivial_classification(trivial_example):
    x, y = trivial_example
    model = BayesianKernelDensityClassifier().fit(x, y)
    assert (model.predict(x) == y).all()


def test_n_jobs_param(trivial_example):
    x, y = trivial_example

    BayesianKernelDensityClassifier()  # Default params
    BayesianKernelDensityClassifier(n_jobs=None).fit(x, y)  # None specified
    BayesianKernelDensityClassifier(n_jobs=-1).fit(x, y)  # None specified
    BayesianKernelDensityClassifier(n_jobs=2).fit(x, y)  # 2 Jobs
    BayesianKernelDensityClassifier(n_jobs=1).fit(x, y)  # 2 Jobs

@pytest.mark.parametrize("jobs", [1.2, 0, -2])
def test_n_jobs_param_fail(trivial_example, jobs):
    x, y = trivial_example
    with pytest.raises(ValueError):
        BayesianKernelDensityClassifier(n_jobs=jobs).fit(x, y)
```

Do you agree with these changes? Adding a `fixture` makes for less repeating code.

arose13:
@koaning Damn that's some fancy ****. I'll create that make sure it's passing and stuff

arose13:
Something unexpected is going on on my end. When I run the following I get the value error I expect...

```python
from sklearn.datasets import make_moons
x, y = make_moons(1000, noise=10)
kde = BayesianKernelDensityClassifier(n_jobs=0).fit(x, y)
```

```python
ValueError: n_jobs == 0 in Parallel has no meaning
```

However, when I run this test....

```python
@pytest.mark.parametrize('n_jobs', [0, 1.23])
def test_n_jobs_params_fails(simple_dataset, n_jobs):
    x, y = simple_dataset
    with pytest.raises(ValueError):
        BayesianKernelDensityClassifier(n_jobs=n_jobs).fit(x, y)
```

I the test fails because setting `n_jobs` to `0` doesn't throw any error. 

```python
simple_dataset = (array([[-11.79409212, -11.59653365],
       [-10.68976232,  -8.81104612],
       [ -7.26670806,  -8.57035241],
      ...., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.,
       1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]))
n_jobs = 0

    @pytest.mark.parametrize('n_jobs', [0, 1.23])
    def test_n_jobs_params_fails(simple_dataset, n_jobs):
        x, y = simple_dataset
        with pytest.raises(ValueError):
>           BayesianKernelDensityClassifier(n_jobs=n_jobs).fit(x, y)
E           Failed: DID NOT RAISE <class 'ValueError'>

tests\test_estimators\test_neighbor_classifier.py:58: Failed
```

Have you guys ever seen something like this before?

arose13:
> So it seems this implementation implements parallel ... for `.predict()` **not** `.fit()`.
> 
> This seems like a bad idea. I think during predict (at least in production) the overhead of spawning a pool of resources will cause the predictions to be slower than they need to be. Allowing it during `fit` makes perfect sense, but I would not have it at inference.
> 
> @MBrouns got an opinion on this?
> 
> Again, if it is easier to just push without parallel that is also fine by me.

Sorry, I'm only now reading this. The model has no fitting time and parallelizing it won't speed it up. The real computation occurs on predict. But maybe `n_jobs` default to None, therefore only using 1 CPU deals with this already?

MBrouns:
the `KernelDensity`s fit method fits a KDtree or something similar right?

arose13:
@MBrouns you are correct monsieur. This means that nothing is really done when `.fit()` is called (besides memorizing the dataset which is often just pointing at it) this is why it's O(1), only when `.predict()` is called is the KDTree traversed and this computationally expensive step.

koaning:
My Proposal: shall we drop parallel for this release? We can always add it later but it feels like we need to have a discussion about it, which is currently blocking this PR as well as a release I'd love to get out.

arose13:
I also figured that it didn't make sense testing for kernels since sklearn already does so I don't do anything interesting with them

koaning:
I will merge this for now and make a release. 

I'll add an issue on the documentation for now.

koaning:
@arose13 thanks for the PR!

arose13:
@koaning twas cool

koaning:
It should be live now too. Newest version on pypi: v0.4.0

PR #284, opened 2020-02-07
Monotonic Intervals 

koaning:
Work in progress for the interval preprocessor ... monotonic shapes are incomming.


MBrouns:
dearme, readme, 

vincent, the betas . only need to increase/decrease.

koaning:
### it's beautiful

image

koaning:
## with smoothin 

image

I've also decided to do a cleanup in this PR. preprocessing really needs to be split up now.

koaning:
Here's the main changes that I've made; 

- I added a fixture in `conftest.py`  such that we also have a lightweight dataset of different sizes. this saved about 4 minutes of testing in just the `test_interval_encoder.py` 
- I added new tests 
- I split up `preprocessing.py` into a submodule for a better overview. The tests are still green. 
- I refactored the old code for the interval encoder a bit to make it easier to allow of other averages 
- I updated the documentation

sklego/preprocessing/intervalencoder.py
```diff
@@ -0,0 +1,144 @@
...
+class IntervalEncoder(TransformerMixin, BaseEstimator):
+    """
+    The interval encoder bends features in `X` with regards to`y`.
+    We take each column in X separately and smooth it towards `y` using
+    the strategy that is defined in `method`.
+    Note that this allows us to make certain features strictly monotonic
+    in your machine learning model if you follow this with an appropriate
+    model.
+    :param n_chunks: the number of cuts that makes the interval
+    :param method: the interpolation method used, must be in ["normal", "average"], default: "normal"
```
MBrouns: `increasing` and `decreasing` are also valid

sklego/preprocessing/intervalencoder.py
```diff
@@ -0,0 +1,144 @@
...
+        # these two matrices will have shape (columns, quantiles)
+        # quantiles indicate where the interval split occurs
+        X, y = check_X_y(X, y, estimator=self)
+        self.quantiles_ = np.zeros((X.shape[1], self.n_chunks))
+        # heights indicate what heights these intervals will have
+        self.heights_ = np.zeros((X.shape[1], self.n_chunks))
+        self.num_cols_ = X.shape[1]
+
+        average_func = (
+            _mk_average
```
MBrouns: The dispatching here feels a bit double. First we're dispatching to `_mk_average` and `_mk_monotonic_average`, but inside those methods we're dispatching again based on `method`. How about something like this:

```
average_func = {
   'normal': _mk_normal,
   'average': _mk_average,
   'increasing': partial(_mk_monotonic, constraint_cmp=operator.ge),
   'decreasing':  partial(_mk_monotonic,  constraint_cmp=operator.le),
}[self.method]

koaning: i was wondering about this. the downside of having two functions `_mk_normal` and `_mk_average` is that you'll end up with two functions with *a lot* or repitition, doing against the DRY principle. 

by having the code run this way you'll get `*_average` functions that all accept a `method` param that feels somewhat general.

sklego/preprocessing/intervalencoder.py
koaning: noted

sklego/preprocessing/intervalencoder.py
koaning: maybe one other thing; there is going to be one extra function added for the `updown` and `downup` relation. it might be preferable to delay any refactoring until it is clear how that function is supposed to work.

MBrouns: Yep, seems like a good idea!

koaning:
Then I think we're green?

Also ... when the density thing is in ... new version? Since I am also updating the docs ... I prefer to merge once we also do a new version.

MBrouns:
New version sounds good!

PR #286, opened 2020-02-09
h2 tag

koaning:
it's a good thing we've added sources, but it should be an h2 tag instead of h1. it now renders like this; 

image


MBrouns:
Good catch, didn't know it worked like that

PR #301, opened 2020-02-28
Remove duplicate if give_pandas: return df lines

JCZuurmond:
Minor change update:

```
if give_pandas:
    return df
```

Was there twice. I removed one of them.


koaning:
Agreed.

PR #302, opened 2020-02-28
add return_type argument to PatsyTransformer

RoyalTS:
Passing `return_type='dataframe'` makes PatsyTransformer return a pandas DataFrame, which has the advantage of being immediately digestible by other Transformers that rely on the presence of column names.


koaning:
Hey @RoyalTS, this looks like a sensible change. I like the two tests. LGTM! Thanks!

PR #310, opened 2020-03-31
Addded Hall of Fame to the docs under contribution

Pverheijen:


koaning:
Could you add a screenshot of what this looks like? 

Also ... do we need to set both `..image` as well as `:target:` here?

koaning:
Since we've got 22 contributors ... I may also be fine with just listing everyone there. Is that also possible?

koaning:
Otherwise top 20 might be preferable.

MBrouns:
> Could you add a screenshot of what this looks like?
> 
> Also ... do we need to set both `..image` as well as `:target:` here?

the target makes sure the picture acts as a link to the contributors github profile

Pverheijen:
> Could you add a screenshot of what this looks like?

Screenshot from 2020-03-31 13-30-47

This is a screenshot of how it currently looks. 

--

In terms of listing all of the contributors or the top20. It's not something configurable in the hall-of-fame version that sourcerer provides without self-hosting. We could change the configuration but then we would have to self host it, or we could build something else ourselves.

Pverheijen:
This is what it looks like when we try to show 8 contributors instead of 7. 
Little bit sad that it doesn't work out of the box like that. 

Screenshot from 2020-03-31 13-42-02

koaning:
Then seven is good enough for me. 

*green*

PR #311, opened 2020-03-31
Implemented make_debug_pipeline in sklego.pipeline and fixed a bug

Pverheijen:
function make_debug_pipeline added to sklego.pipeline
fixed an example in the docs for debug_pipeline, last step should be a passthrough


koaning:
> last step should be a passthrough

Could you clarify why? 

Also, could you add a test to this PR? 

I wonder what happens in an edge case where there are two StandardScalers after eachother. Will these two steps have the same name?

MBrouns:
> I wonder what happens in an edge case where there are two StandardScalers after eachother. Will these two steps have the same name?

sklearn adds a trailing unique id number to it

Pverheijen:
The mistake was mine actually;

I've tried using named steps with the make_debug_pipeline which resulted in the following error code: 

TypeError: Last step of Pipeline should implement fit or be the string 'passthrough'. '[('add_1', Adder(value=1)), ('add_10', Adder(value=10)), ('add_100', Adder(value=100)), ('add_1000', Adder(value=1000))]' (type <class 'list'>) doesn't

This was caused by calling:
debugpipeline = make_debug_pipeline(steps)

What I should have used instead was: 
debugpipeline = make_debug_pipeline(*[s[1] for s in steps])

Pverheijen:
> > last step should be a passthrough
> 
> Could you clarify why?
> 
> Also, could you add a test to this PR? I see the docstest, but having one/two explicitly in tests might be good too.
> 
> I wonder what happens in an edge case where there are two StandardScalers after eachother. Will these two steps have the same name?

Will work on some tests, quick update for the edge case with two standardscalers right after each other. It actually does name them separately. 

Screenshot from 2020-03-31 15-00-27

koaning:
@Pverheijen when leaving an error log, or code in general on github, consider adding ticks

three ticks (` ` `) directly after eachother make a huge difference in readability 

```python
# this is a comment
print("and this is easy to read") 
```

Pverheijen:
Adapted some of the test_debug_pipeline.py tests, didn't manage to adapt test_classifier_gridsearch unfortunately. 

Added another test to see if the names of two repeated steps (e.g. StandardScalar) would cause name issues. 

Could someone review the code and the tests please?

koaning:
You've added tests, this is good, but it feels like the tests you're adding can also just be placed in the original debug pipeline file. By making another tests file you're also copying a lot of the code which is making it harder for us to maintain. 

Also, 2-3 tests is fine.

Pverheijen:
> You've added tests, this is good, but it feels like the tests you're adding can also just be placed in the original debug pipeline file. By making another tests file you're also copying a lot of the code which is making it harder for us to maintain.
> 
> Also, 2-3 tests is fine.

Good point, I did not think about that. I'll make the changes later this week. 
Thank you for the feedback so far.

Pverheijen:
I've made the changes by reducing the number of tests (keeping the ones that made most sense to me) and merging it with the original test file to improve maintainability. 

Ready for review.

koaning:
Urgh ... another connectiong issue with travis.

koaning:
I have no comments at this point, but I do wanna fix the CI before I merge this.

sklego/pipeline.py
```diff
@@ -301,3 +303,78 @@ def log_callback(self, func):
         self._log_callback = func
         if self._log_callback == "default":
             self._log_callback = default_log_callback
+
+def _name_estimators(estimators):
```
JCZuurmond, line 307: Why not use the `_name_estimators` from sklearn? Am I missing something?

koaning:
Tests are back. But I'm curious to hear the reponse on @JCZuurmond's question before I merge.

sklego/pipeline.py
Pverheijen: Good catch, not missing anything. Importing it is simply better. I've made the suggested change and made a commit for it. Thanks Cor :-)

PR #312, opened 2020-04-01
Include __all__ statement s.t. preprocessing classes are documented

massiung:
By including an `__all__` statement in the `__init__.py` we ensure that sphinx picks up the right components from the submodule. A side effect is that you can now use `from preprocessing import *`

**Downside:** Needs maintaining whenever new functionality is added to the submodule: this is already kind of the case because of the include statements in the init.

### DIFF
--- a/sklego/preprocessing/__init__.py
+++ b/sklego/preprocessing/__init__.py
@@ -1,3 +1,16 @@
+__all__ = [
+    "IntervalEncoder",
+    "RandomAdder",
+    "Patsytransformer",
+    "ColumnSelector",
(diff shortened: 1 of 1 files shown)


koaning:
I think I like this change. Just to check @MBrouns agree?

MBrouns:
Definitely! Thanks for the pr!

koaning:
Sweet. Merging is proceeding. Once the tests are green I will merge. 

Also ... @massiung thanks! Feel free to remind us about a sticker next time you're at a meeutp!

Issue #232, opened 2019-10-28
[FEATURE] TimeGapSplit add n_splits param


stephanecollot:
Currently in TimeGapSplit you can specify exactly the behaviour in production with:
- `train_duration`
- `valid_duration`
- `gap_duration`

and the class will define for you automatically the number folds.

We could add a new parameter `n_splits`, that will allow to give for example only `gap_duration` and `valid_duration`. The use case is the following: usually we know the gap duration (given by target definition) and validation duration (given by the retraining frequency), and we usually want 5 folds for hyper-parameters tuning.

In other words, we will have 4 parameters, and you need to provide 3 of them.

PR #324, opened 2020-04-20
Fix start of train split in TimeGapSplit and added n_split parameter

rpauli:
Addresses changes in #192  and #232 
I am currently working on a Time Series problem with vibration data where I needed a functionality like the one suggested in #232 so I decided to add it here.

I tried to explain the changes in functionality in the docstring:

> Each validation fold doesn't overlap. The entire 'window' moves by 1 valid_duration until there is not enough data.
> If this would lead to more splits then specified with n_splits, the 'window' moves by
>     the validation_duration times the fraction of possible splits and requested splits
>      -- n_possible_splits = (total_length-train_duration-gap_duration)//valid_duration
>      -- time_shift = valid_duratiopn n_possible_splits/n_slits
>     so the CV spans the whole dataset.
>     If train_duration is not passed but n_split,
>     the training duration is increased to
>      -- train_duration = total_length-(self.gap_duration + self.valid_duration * self.n_splits)
>      such that the shifting the entire window by one validation duration spans the whole training set

The changes are also added to the docs notebook for visualization.


sklego/model_selection.py
```diff
@@ -80,29 +95,61 @@ def split(self, X, y=None, groups=None):
...
         date_max = X_index_df["__date__"].max()
+        date_length = X_index_df["__date__"].max() - \
+            X_index_df["__date__"].min()
+
+        if ((self.train_duration is None)
+                and (self.n_splits is not None)):
+            self.train_duration = date_length - \
+                (self.gap_duration + self.valid_duration * self.n_splits)
+
+        if ((self.train_duration is not None)
```
MBrouns, line 106: This check already happens in the `__init__` right? Is it needed to do it again here/

sklego/model_selection.py
```diff
@@ -10,18 +10,23 @@
...
     is going to be used in production in batch learning.
-
     Each validation fold doesn't overlap. The entire 'window' moves by 1 valid_duration until there is not enough data.
-    The number of folds is automatically defined that way.
+    If this would lead to more splits then specified with n_splits, the 'window' moves by
+    the validation_duration times the fraction of possible splits and requested splits
+     -- n_possible_splits = (total_length-train_duration-gap_duration)//valid_duration
+     -- time_shift = valid_duratiopn n_possible_splits/n_slits
+    so the CV spans the whole dataset.
+    If train_duration is not passed but n_split,
```
MBrouns, line 26: missing "is" at the end of the line?

sklego/model_selection.py
```diff
@@ -49,14 +61,17 @@ def __init__(
         self.train_duration = train_duration
         self.valid_duration = valid_duration
         self.gap_duration = gap_duration
+        self.n_splits = n_splits
+        self.maximise_training_set = False
```
MBrouns, line 65: Any reason to not add this as an argument to `__init__`? Also, I think changing it into something like `window_mode='rolling'` and `window_mode='expanding'` is a bit more clear. @koaning?

sklego/model_selection.py
rpauli: Yes it does, train_duration can be None and thus the check only happens partly in the init, if train duration is calculated using valid_duration and n_splits, I do the check again

sklego/model_selection.py
rpauli: In the Issue someone suggested to not add another parameter to the API to not crowd it, but I like the window_mode Idea

rpauli:
> Would it make sense to rename n_splits to max_splits? In the current implementation it seems possible to end up with fewer than n_splits splits, but never more.


Fewer shouldnt be possible, I added this check for it:
 `if n_split_max < self.n_splits`

   

> Could you add some of the checks you do in the notebook as automated pytest tests? Let me know if you need help with that.
I'll come up with a suggestion later today, thanks!

To make this explicit:
The prior behavior was to take train_duration and subtract gap_duration to get the train split.
I found this behavior unintuitive and confusing. Although I think I understand why it was done this way I now add the gap between train and validation without subtracting it from train_duration.
if this doesn't work for you I can change it back

sklego/model_selection.py
koaning: If the object has 7 inputs I might agree, but here the alternative is to add another cross validator that does have the feature which would be worse. I'm only wondering if `window_mode` is the best name. Why not just `window`?

doc/timegapsplit.ipynb
```diff
@@ -2,14 +2,53 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 2,
```
koaning, line 5: Could you remove the pip install here?

koaning: Also, feel free to add documentation for your setting, and usecase, here. You don't have to do it, but if typically people won't notice the feature unless it is in a getting started guide.

sklego/model_selection.py
```diff
@@ -49,14 +61,17 @@ def __init__(
...
 
     def join_date_and_x(self, X):
```
koaning, line 67: This should have happened earlier, but maybe make this a hidden method? 

From

```
def join_date_and_x(self, X):
```

Into

```
def join_date_and_x_(self, X):
```

After all, the user should never touch it.

MBrouns: leading underscore right, not trailing?

sklego/model_selection.py
MBrouns: yep, `window` is probably better

doc/timegapsplit.ipynb
rpauli: yes ofc, just an artifact of me failing to properly use virtual environments

koaning:
@rpauli I like these changes. Two quick things though. 

1. Could you have a look at the tests? It feels like there's `assert`s missing
2. The notebook situation with the docs is not ideal, and part of me is embarassed to say this, but I fear you need to render it for the docs to render.

koaning:
There is no rush, but just to check; @rpauli are you waiting for feedback from us?

rpauli:
I added some more explicit assert statements and aded the strict tag to the tests that are supposed to fail. I noticed you handled that differently in test_timegapsplit_too_big_gap should I also change it to catch the except? 

I ran the ipynb before pushing, is this what you mean with render it?

koaning:
> I ran the ipynb before pushing, is this what you mean with render it? 

Yes 👍

tests/test_model_selection/test_timegapsplit.py
```diff
@@ -55,6 +60,55 @@ def test_timegapsplit_too_big_gap():
...
+        date_serie=df["date"],
+        train_duration=timedelta(days=5),
+        valid_duration=timedelta(days=3),
+        gap_duration=timedelta(days=1),
+        n_splits=7
+    )
+    list(cv.split(X_train, y_train))
+
+
+@pytest.mark.xfail(raises=ValueError, strict=True)
```
koaning, line 86: I have never used `xfail` before. I've just checked the documentation but I wonder; what is the main benefit of using this as opposed to using this:

```python
def test_timegapsplit_train_or_nsplit():
   with pytest.raises(ValueError):
	    cv = TimeGapSplit(
	        date_serie=df["date"],
	        train_duration=None,
	        valid_duration=timedelta(days=3),
	        gap_duration=timedelta(days=5),
	        n_splits=None
	    )
	    cv.split(X_train, y_train)
```

koaning:
I've only got one comment about the tests, but it is starting to look green to me. @MBrouns?

rpauli:
I reread some pytest docs and it seems I misunderstood how xfail is supposed to be used, changed it to use 
`with pytest.raise`

MBrouns:
LGTM! We can merge when @koaning approves as well.

Thanks a lot for the pull request @rpauli! If you reach out to me on Twitter or LinkedIn with your address I'll make sure you'll get some stickers

koaning:
It is merged, I will now also make a release with this feature in it so that you can use it right away. 

@rpauli just to check, have we met in real life at a PyData by any chance? I'm curious to hear how you discovered this package.

koaning:
Also, @rpauli if you have a twitter handle I can mention you when I announce the update.

koaning:
And it's live with a version bump: https://pypi.org/project/scikit-lego/0.4.2/#history

rpauli:
> @rpauli just to check, have we met in real life at a PyData by any chance? I'm curious to hear how you discovered this package.

Saw one of your pyData talks on gaussian processes and outlier detection and found this package with things I also implemented at work (although less structured) so I decided to contribute a bit

rpauli:
> And it's live with a version bump: https://pypi.org/project/scikit-lego/0.4.2/#history

Great to hear, first open source contribution I wasn't paid for!

PR #331, opened 2020-05-01
Meta cleaning

koaning:
Quarantine means cleanup time.


koaning:
@MBrouns I am removing appveyor. GH-actions is *much* faster for windows. Cool?

Issue #323, opened 2020-04-19
[FEATURE] make_outlier_scorer 


koaning:
From [this](https://koaning.io/posts/outliers-selection-vs-detection/) blogpost. It would be nice if we have a variant for "make_scorer" that can take metrics meant for classification and allow them to also be used to measure outlier detection algorithms.

```python
from sklearn.metrics import precision_score, recall_score
from sklearn.model_selection import GridSearchCV

df_subset = df.sort_values('class', ascending=False)[:80000]
X_subset = df_subset.drop(columns=['amount', 'class', 'time']).values
y_subset = df_subset['class']

def outlier_precision(mod, X, y):
    preds = mod.predict(X)
    return precision_score(y, np.where(preds == 1, 0, 1))

def outlier_recall(mod, X, y):
    preds = mod.predict(X)
    return recall_score(y, np.where(preds == 1, 0, 1))

forest = IsolationForest(contamination=0.1, behaviour="new", max_features=0.2)

mod = GridSearchCV(estimator=forest, 
                   cv=5,
                   n_jobs=-1,
                   scoring={"precision": outlier_precision,
                            "recall": outlier_recall}, 
                   refit="precision",
                   param_grid={'contamination': np.linspace(0.0001, 0.02, 30)})

mod.fit(X, y)
```

koaning:
This might be better as a Meta model now that I think about it. 

@MBrouns, can I go ahead or do you have concerns?

koaning:
This is my idea now. Instead of this; 

```python
from sklearn.metrics import precision_score, recall_score
from sklearn.model_selection import GridSearchCV

df_subset = df.sort_values('class', ascending=False)[:80000]
X_subset = df_subset.drop(columns=['amount', 'class', 'time']).values
y_subset = df_subset['class']

def outlier_precision(mod, X, y):
    preds = mod.predict(X)
    return precision_score(y, np.where(preds == 1, 0, 1))

def outlier_recall(mod, X, y):
    preds = mod.predict(X)
    return recall_score(y, np.where(preds == 1, 0, 1))

forest = IsolationForest(contamination=0.1, behaviour="new", max_features=0.2)

mod = GridSearchCV(estimator=forest, 
                   cv=5,
                   n_jobs=-1,
                   scoring={"precision": outlier_precision,
                            "recall": outlier_recall}, 
                   refit="precision",
                   param_grid={'contamination': np.linspace(0.0001, 0.02, 30)})

mod.fit(X, y)
[...]

PR #334, opened 2020-05-02
Meta outlier classifier

koaning:
This turns outlier detectors into classifiers.


koaning:
@MBrouns some of the classifier tests totally break here, but I think this one will resolve https://github.com/koaning/scikit-lego/issues/323.

sklego/meta/outlier_classifier.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Morphs an estimator that performs outlier detection into a classifier.
+    When an outlier is detected it will output 1 and 0 otherwise.
+    This way you can use familiar metrics again and this allows you
+    to consider outlier models as a fraud detector.
+    """
+    def __init__(self, model):
+        self.model = model
+
+    def _is_outlier_model(self):
```
MBrouns, line 19: Would this be solved in a nicer way by implementing another meta model, like the `ProbabilisticClassifier`


```class OutlierModelMeta(type):
    def __instancecheck__(self, other):
        return isinstance(self.model, OutlierMixin)

class OutlierModel(metaclass=OutlierModelMeta):
    pass
```

koaning: Valid. Made the changes.

koaning:
@MBrouns should be fixed. I added your request, removed the notebook and added one extra test.

PR #336, opened 2020-05-02
Regression Based Outlier Detection

koaning:
Implementation of a regression based outlier model. 

```
import numpy as np
import matplotlib.pylab as plt

X = np.random.normal(0, 1, (100, 1))
y = 1 + np.sum(X, axis=1).reshape(-1, 1) + np.random.normal(0, 0.2, (100, 1))
for i in [20, 25, 50, 80]:
    y[i] += 2
X = np.concatenate([X, y], axis=1)

from sklearn.linear_model import LinearRegression
mod = RegressionOutlierDetector(LinearRegression(), column=1)
mod.fit(X)
plt.scatter(X[:, 0], X[:, 1], c=mod.predict(X));
```

image


koaning:
@MBrouns this might be like-able :)

koaning:
@MBrouns I think this is a PR that has been done for a long time. Got any objections if I merge this?

MBrouns (approved):
LGTM

Issue #335, opened 2020-05-02
[BUG] OutlierRemover is in wrong submodule.


koaning:
`OutlierRemover` should not be in meta. It should be in `preprocessing`. 

![image](https://user-images.githubusercontent.com/1019791/80860371-85a62600-8c67-11ea-9273-2527db1f5110.png)

koaning:
@MBrouns should we put a deprecation warning here while we move it?

Abo7atm:
Hi, does moving the class requires more than moving `outlier_remover.py` to `preprocessing` directory? Also, can I take care of this?

MBrouns:
so it would probably be nicest to copy the class to `preprocessing`, and mark the original class in `meta` with an `@deprecated` decorator as such:

```
@deprecated(
    version="0.4.1",
    reason="Please use `sklego.preprocessing.OutlierRemover instead`",
)
```

Feel free to pick this up!

koaning:
I agree with @MBrouns but I might add the extra text; 

```
@deprecated(
    version="0.4.1",
    reason="Please use `sklego.preprocessing.OutlierRemover instead`. This object will be removed from the meta submodule in version 0.6.0.",
)
```

Just to make it explicit that we intend on removing it later.

PR #338, opened 2020-05-03
Move OutlierRemover to correct submodule (#335)

Abo7atm:
Fix the bug in #335


sklego/meta/outlier_remover.py
```diff
@@ -7,7 +7,11 @@
 
 from sklego.common import TrainOnlyTransformerMixin
 
-
+@deprecated(
+    version="0.4.1",
```
koaning, line 11: We just bumped the verison. It's 0.4.2 now.

Abo7atm:
Thanks for the tutorial on flake8! I'll take care of the import and the unit test.

sklego/meta/outlier_remover.py
```diff
@@ -7,7 +7,11 @@
 
 from sklego.common import TrainOnlyTransformerMixin
 
-
+@deprecated(
```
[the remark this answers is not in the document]
Abo7atm: Forgot to push it, sorry.

koaning:
@Abo7atm I would keep the old tests for now. Deprecation should only imply that the change will happen later. The code in `meta` should still be functional until then. 

But yes, feel free to copy the tests and add a new file to `test_preprocessing`.

tests/test_meta/test_outlier_remover.py
```diff
@@ -6,7 +6,7 @@
 from sklearn.utils import estimator_checks
 
 from sklego.common import flatten
-from sklego.meta import OutlierRemover
```
koaning, line 9: I think this test file should still point to `meta`. Otherwise we'll have two tests checking in the same code. The goal is to make sure that the original code in `meta` still works until we deprecate it.

tests/test_preprocessing/test_outlier_remover.py
```diff
@@ -0,0 +1,87 @@
...
+    outlier_remover = OutlierRemover(outlier_detector=mock_outlier_detector, refit=True)
+    outlier_remover.fit(X=np.array([[5, 5]]))
+    assert len(outlier_remover.transform_train(np.array([[0, 0]]))) == 0
+
+
+def test_do_not_refit(mocker):
+    mock_outlier_detector = mocker.Mock()
+    mock_outlier_detector.fit.return_value = None
+    mock_outlier_detector.predict.return_value = np.array([-1])
+    mocker.patch("sklego.meta.outlier_remover.clone").return_value = mock_outlier_detector
```
koaning, line 65: The reason why the tests are failing is probably due to this line. It's a long story to explain what the `mocker` does here exactly but you should be able to varify that pytest fails locally on your machine as well. If you're less experienced with pytest you may enjoy [this resource](https://calmcode.io/pytest/introduction.html). 

The import issue above should fix the tests though.

Abo7atm: Again, thanks for the neat tutorial!
It makes sense that the tests are failing due to the import. I've fixed the import and verified that the tests have passed.

koaning:
Looks good to me! Thanks for the effort, I will now click merge and you have hereby earned the sticker that I gave you during the training 😄

PR #347, opened 2020-05-13
Fix fairclassifier tests

MBrouns:
It turns out that inside pretty printing the classifier there's some logic that re-inspects the method signature to actually get the parameters for printing. This bypasses the get_param_names and get_params which were properly implemented for all other functionalities


koaning:
Curious. I tried something similar and that gave all sorts of errors. This seems to run fine. One the unit tests are green I'll approve + merge.

PR #349, opened 2020-05-15
Change give_pandas to as_frame

ktiamur:
Hi, I've heard about this project from @MBrouns and felt like solving a small issue to make a contribution. 😄 

The keyword give_pandas has been changed to as_frame when returning pandas dataframes as scikit-learn has chosen for this standard.


koaning:
Thanks for the changes! I wonder though ... 

@MBrouns do we need deprecation messages here?

MBrouns:
Deprecation message would be nice, just in case. @ktiamur could you add that? There's an example on how it works in the linear_models section on the `FairClassifier`

ktiamur:
Yes, I'll add deprecation messages

ktiamur:
I had a look at this today and found it not so straightforward to implement. The deprecated package used in linear_models can only deprecate functions or classes. In this case I don't think we want to deprecate the function but only the argument.

One idea I had was to write something like this:
```python3
def load_arrests(return_X_y=False, as_frame=False, **kwargs):
    if "give_pandas" in kwargs:
        warnings.warn("give_pandas is deprecated; use as_frame", DeprecationWarning)
        as_frame = kwargs["give_pandas"]

    ...
```

The downside of this is that it would have to be done for all functions that use this argument. However, I found a rather neat alternative on stack overflow to get this functionality: https://stackoverflow.com/questions/49802412/how-to-implement-deprecation-in-python-with-argument-alias

If I were to proceed with the approach shown in the link, where would a good place be to put the helper functions (deprecated_alias and rename_kwargs)? 

@MBrouns @koaning I'd like to hear your thoughts on this

koaning:
I certainly like the idea of having a decorator for this. Then again ... 

Part of me thinks it may just be fine to have these inline for now. I expect this submodule to be the only one for now that has this re-naming issue and before generalising it it would be nice to understand the deprecation policy that will work for our package a bit better.

ktiamur:
I added the warnings inline. I also changed DeprecationWarning to FutureWarning as, if I understand correctly, the intend is to remove give_pandas in version 0.7.0.

koaning:
@MBrouns I think we're good on merging this, correct?

PR #350, opened 2020-05-20
Small Test Speedup

koaning:
I've added the `pytest-xdist` package which allows unit tests to be run in parallel. 

A github action worker should have two cores so we'll see what the exact effect of it is. Locally it's made a *huge* difference. From  223.29s to 75.38s.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -36,4 +36,4 @@ jobs:
     - name: Test with pytest
       run: |
         pip install pytest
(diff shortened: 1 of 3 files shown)


koaning:
I've also reduced the packages to be installed. It seems to shave off 40s of the windows install step.

koaning:
I'd love to get the CI time data from github actions, that would allow me to make a good conclusion, but it seems the extra cores is barely notice-able while the reduction of packages does make a modest impact.

PR #352, opened 2020-05-23
added fetch_creditcard

koaning:
@MBrouns I've added the function in this PR. I'm not super sure what we want to do in terms of unit testing. I am using a scikit-learn internal function that should already be unit testsed and I think it is bad practice to download during every CI. If you see a nice way to test this, I'm all ears.


koaning:
Added some nice docs tho. 

image

PR #354, opened 2020-05-24
Doctests were off, turned 'em back on

koaning:
@JCZuurmond it seems that during the travis -> github actions migration we forgot to add the doctests. I just re-ran and it seems like the debug pipeline is failing. I've added them back again but I cannot see the issue? I figured you might be able to instantly spot it. If not, no worries, this is not high priority.


MBrouns:
test should be green now :)

MBrouns:
ooor.. maybe not. Interesting, seems like there's a difference between running with `pytest --doctest-modules` and running doctests from PyCharm

PR #358, opened 2020-05-26
Minor: Fix linear model docs mistake

Pverheijen:
In constraint it limits the values to be larger than 1, I believe this should be larger than 0.


Pverheijen:
It looks like I accidentally also removed two files.... let me close this and try again.

koaning:
Good catch. There are two constraints tho. Everything must be smaller than 1 and larger than 0.

PR #359, opened 2020-05-26
Minor mistake in linear models docs, probs geq than zero not 1

Pverheijen:


koaning:
Nice catch!

koaning:
Nice catch!

PR #363, opened 2020-05-28
PCA + UMAP outlier detection

koaning:
I guess it would have been nicer to add an issue for this first but I felt that outlier-detection via reconstruction would be valid to have. Implementing this for PCA and UMAP.


koaning:
This should cause another version bump. @MBrouns opposed to 0.5.0?

MBrouns:
0.5.0 sounds good!

Issue #372, opened 2020-06-23
[FEATURE] DictMapper


koaning:
![image](https://user-images.githubusercontent.com/1019791/85385682-e4d73700-b542-11ea-8604-6ddd444a048d.png)
In this example you pass the transformer a column to transform but also a dict. In this case you'd pass; 

```
{'young': 1, 'adult': 2, 'old': 3}
```

A transformer that transformers based on a dictionary. I don't know if sklearn has this feature but I couldn't find it.

kayhoogland:
https://github.com/koaning/scikit-lego/pull/375

koaning:
Typically I'd assume that @MBrouns confirms that ideas that I pitch here is a sound one before I start working on an issue and in this particular case I was hoping to use this example in a [calmcode](https://calmcode.io) series. In this case it went fine but @kayhoogland feel free to in the future claim an issue before you start working on it. Otherwise folks might do double work.

That said, I'll gladly have a look at the PR.

kayhoogland:
That makes sense! Next time I will leave a short comment before I start working on it. Since I created this tranformer before, it did not take a lot of time to code.

kayhoogland:
I think this encoder goes beyond the OrdinalEncoder as I can also decide to map a string to a string. For example:

```
{
"Netherlands": "Europe",
"Belgium": "Europe",
"Brazil": "South-America"
}
```

I imagine this can also be GridSearched if I would like to try different mappings. For example:
`{"Amsterdam": "Netherlands", ... }` versus `{"Amsterdam": "Europe", ...}`

koaning:
@kayhoogland this is why I usually wait until @MBrouns gives feedback before I implement something ;). 

From my end: I've tried using the Ordinal Encoder in the past but it is *not* an intuitive API. Note that I think of it, you could also technically use the FunctionTransformer with a lambda to implement this since nothing is really learned here. I do think this approach wouldn't gridsearch nicely because you'd have a lambda function in there.

I can imagine if you map strings to numeric values you could get some gridsearch value out of it though. I like you could enfore monotonicity across a continous variable this way and how you might add the hard "up-down-up" shape too if you'd like. 

```
{"young": 1, "adult": 2, "old": 3}
{"young": 1, "adult": 2, "old": 5}
```

Suppose you'd use these mappings in a disease suseptibility study, I can imagine it can be use-ful to gridsearch.

kayhoogland:
@koaning I learned that the hard way now 😉. 

[...]

PR #375, opened 2020-06-27
DictMapper

kayhoogland:
Implementation of the DictMapper described in https://github.com/koaning/scikit-lego/issues/372

Let me know what you think!


sklego/preprocessing/dictmapper.py
```diff
@@ -0,0 +1,48 @@
...
+    :param mapper: The dictionary containing the mapping of the values
+    :param default: The value to fall back to if the value is not in the mapper
+    """
+
+    def __init__(self, mapper, default):
+        self.mapper = mapper
+        self.default = default
+
+    def fit(self, X, y=None):
+        X = check_array(
```
koaning, line 22: Should we maybe check here if X is a single column? If multiple columns are passed I think this object should break.

kayhoogland: I agree a single column would be the typical use case. I can also imagine use cases where we would like to use it one multiple columns:

```
  value_before value_after
0          foo         bar
```

koaning:
I just confirmed with @MBrouns over slack. We'll allow it and I'll handle the PR for the review.

sklego/preprocessing/dictmapper.py
koaning: Fair enough.

sklego/preprocessing/dictmapper.py
```diff
@@ -0,0 +1,48 @@
...
+
+    :param mapper: The dictionary containing the mapping of the values
+    :param default: The value to fall back to if the value is not in the mapper
+    """
+
+    def __init__(self, mapper, default):
+        self.mapper = mapper
+        self.default = default
+
+    def fit(self, X, y=None):
```
koaning, line 21: nitpick, but could you add a docstring here? feel free to copy it from another class.

sklego/preprocessing/dictmapper.py
```diff
@@ -0,0 +1,48 @@
...
+            copy=True,
+            estimator=self,
+            force_all_finite=True,
+            dtype=None,
+            ensure_2d=True,
+        )
+        self.dim_ = X.shape[1]
+        return self
+
+    def transform(self, X):
```
koaning, line 33: nitpick, but could you add a docstring here? feel free to copy it from another class.

sklego/preprocessing/dictmapper.py
```diff
@@ -0,0 +1,48 @@
...
+        self.mapper = mapper
+        self.default = default
+
+    def fit(self, X, y=None):
+        X = check_array(
+            X,
+            copy=True,
+            estimator=self,
+            force_all_finite=True,
+            dtype=None,
```
koaning, line 27: about the dtype ... do we maybe want to exclude floats?

sklego/preprocessing/dictmapper.py
```diff
@@ -0,0 +1,75 @@
...
+
+        :type X: pandas.DataFrame or numpy.ndarray
+        :param X: The column(s) from which the mapping will be applied
+
+        :param y: Ignored.
+
+        :rtype: sklego.preprocessing.DictMapper
+        :returns: The fitted object.
+
+        :raises:
```
koaning, line 33: Do we raise a ValueError if X contains a non-numeric column here? This might have ben a copy and paste too much.

koaning: I think this is my final comment.

kayhoogland: My bad, was still working on it

sklego/preprocessing/dictmapper.py
koaning: I think this is fine for now actually. Just checked the documentation: the function might do an implicit conversion if this param is not `None`, which is iffy too.

Issue #376, opened 2020-06-29
[BUG] Cannot install on Mac OS X High Sierra


davideanastasia:
I am currently using a pipenv environment with Python 3.7.7 originally sourced via Homebrew.
When I run `pipenv install scikit-lego`, I get an error which renders the installation impossibile:

```[pipenv.exceptions.InstallError]:   File "/usr/local/lib/python3.7/site-packages/pipenv/cli/command.py", line 254, in install
[pipenv.exceptions.InstallError]:       editable_packages=state.installstate.editables,
[pipenv.exceptions.InstallError]:   File "/usr/local/lib/python3.7/site-packages/pipenv/core.py", line 1992, in do_install
[pipenv.exceptions.InstallError]:       skip_lock=skip_lock,
[pipenv.exceptions.InstallError]:   File "/usr/local/lib/python3.7/site-packages/pipenv/core.py", line 1253, in do_init
[pipenv.exceptions.InstallError]:       pypi_mirror=pypi_mirror,
[pipenv.exceptions.InstallError]:   File "/usr/local/lib/python3.7/site-packages/pipenv/core.py", line 862, in do_install_dependencies
[pipenv.exceptions.InstallError]:       _cleanup_procs(procs, False, failed_deps_queue, retry=False)
[pipenv.exceptions.InstallError]:   File "/usr/local/lib/python3.7/site-packages/pipenv/core.py", line 681, in _cleanup_procs
[pipenv.exceptions.InstallError]:       raise exceptions.InstallError(c.dep.name, extra=err_lines)
[pipenv.exceptions.InstallError]: []
[...]

PR #378, opened 2020-06-30
Schedule dependencies workflow file added

vtoliveira:
I did a minimal file just to install scikit-lego from time to time using schedule feature of github actions.


MBrouns:
Hi there, thanks for the PR! This will be a good addition.

Some questions:
- Is it worth testing on Windows?
- Is it worth also running the test suite?

I don't necessarily want those questions to block this PR though, so if there's no definitive answer on those I'm +1 on merging this

vtoliveira:
Actually, it is totally up to you guys. I was testing first on mac-os because of issue #376 as mentioned by @koaning. I didn't add windows because in your test-windows.yml file you tested only with python 3.6, so I don't know if there is a problem. 

But as he suggested, I made PR so you could @MBrouns make suggestions.

koaning:
My main painpoint is that when scikit-learn releases a new version we need to hear about it from an issue. It'd be nicer if we got an email of a failed master cron job such that we can check our dependencies. I wonder about extra information that it might be able to give us ... but just the cronjob would be nice to have already.

koaning:
Actually come to think of it, let's add windows just in case.

.github/workflows/schedule-dependencies.yml
```diff
@@ -0,0 +1,34 @@
...
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
```
koaning, line 19: I think it is fine if we only check for 3.7 for now.

.github/workflows/schedule-dependencies.yml
```diff
@@ -0,0 +1,34 @@
...
+        os: [macos-10.15, ubuntu-latest]
```
koaning, line 20: Would be nice to add windows. Windows latest would be fine.

.github/workflows/schedule-dependencies.yml
```diff
@@ -0,0 +1,34 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install scikit-lego
+        pip freeze
```
koaning, line 34: We should run pytest here as well.

koaning: Dependencies might install just fine even if their APIs changed. To detect if something broke, we'd still need to run the tests.

vtoliveira:
Sorry for the delayed answer. I've committed the changes as requested. Also, I ran locally and it went everything ok. 

@koaning @MBrouns 
https://github.com/vtoliveira/scikit-lego/actions/runs/160992788

koaning:
looks good to me, thanks!

Issue #288, opened 2020-02-09
[DOCS] datasets are currently undocumented


koaning:
It would be nice to have a notebook that shows a plot and a few rows of all the datasets that we have.

koaning:
The API lists the datasets, which is good, but it would improve the documentation if there was a small overview available as well. Maybe just showing the first few rows and maybe a plot. Nothing too fancy but enough to get somebody new up to scratch very quickly.

nvanommeren:
Hello! I just submitted a merge request and still have a question:

- There is a notebooks folder inside the docs folder, but there are also many notebooks outside of this folder. What is the preferred location for the datasets notebook? 

Let me know if you have any feedback.

koaning:
This issue is now fixed.

PR #385, opened 2020-07-12
add notebook explaining the datasets

nvanommeren:
Should solve issue: [DOCS] datasets are currently undocumented #288

### DIFF
--- a/doc/datasets.ipynb
+++ b/doc/datasets.ipynb
@@ -1,0 +1,857 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import matplotlib.pylab as plt\n",
+    "import numpy as np\n",
+    "\n",
+    "from collections import Counter"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Datasets\n",
+    "\n",
+    "Scikit-lego includes several datasets which can be used for testing purposes. Each dataset has different options for returning the data:\n",
+    "\n",
+    "- When setting ``as_frame`` to True, the data, including the target, is returned as a dataframe.\n",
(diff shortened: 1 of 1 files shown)


koaning:
I think this looks good as a first start. I'm certainly willing to add this but there are some datasets missing. 

You can check [here](https://scikit-lego.readthedocs.io/en/latest/api/datasets.html) I think this PR is missing; 

- make_simpleseries
- fetch_creditcard
- load_penguins 

I'm certainly willing to merge this work for now. But if you prefer I can also wait until these datasets have also been added.

koaning:
You've placed this notebook in the correct folder by the way. All notebooks related to documentation should go there.

nvanommeren:
How can I forget the penguins ... ;) 

Thanks, added the last three datasets. 

One small remark: the behaviour of `fetch_creditcard` is slightly different compared to the other datasets. When I run it with `as_frame` = True it returns a dictionary including the frame instead of the frame directly.

koaning:
This is true! Well spotted. 

I'm not 100% with how I'd like to handle this forward. Technically all of the `load_*` functions use an API that is inconsistent with the scikit-learn API. The `fetch_*` functions will use a tool inside of scikit-learn which means that it is tricky to not use the official scikit-learn standard. For now I think it is fine if we just clearly document this but in the long run a proper decision has to be made.

I think we're going to move this portion of the library to be 100% consistent with scikit-learn at some point in the future. Don't know when yet though.

koaning:
That said this document looks really great, so I'll go ahead and merge. Thanks! 

If you've got a twitter account I'd love to give you a shoutout and announce this by the way.

nvanommeren:
Yes agree, it would be nicest to be consistent with the scikit-learn library. 

Thanks! I don't have twitter, but added you on LinkdIn.

koaning:
Just gave a small shoutout. Thanks again!

koaning:
Also ... we have stickers! Feel free to ping myself or Matthijs at a meetup in the future.

PR #387, opened 2020-07-16
GroupedTransformer

### LINKED ISSUE #383: [FEATURE] Grouped Transformations
I think it'd be nice to have the following feature, or something similar: The group mean featurizer transforms a numeric feature based on a categorical feature, by deleting the mean of the feature in each category. That way, if we know that the numeric and the categorical feature are very correlated, we can obtain some kind of decorrelated feature.

I think it makes sense to be used both in supervised learning and anomaly detection. I've used it in several supervised problems and heard about it on Kaggle.

I'll give an example. The numeric variable might be the size of an apartment and the category might be the city where the apartment is. These features might be useful to predict apartment prices, for instance. 

The bigger an apartment, the more expensive it should be. However, if we have data from London and Cardiff, London has more expensive and small apartments. The data might actually tell us that smaller apartments are very expensive. If we pass the apartment size and the city through the group mean featurizer, we can mitigate this issue, as the apartment size will be in relation to the mean of the city. 

An apartment of 100 squared meters will be huge in London and regular in Cardiff. For this reason, the feature will give a very high value for a London apartment and a regular value for a Cardiff apartment.

image

I think it is suited as a transformer, since the idea is to obtain the mean per group at training time in order to avoid leakage.

We might have other use-cases, such:
- Number of downloads of a movie with respect to its genre.
- Amount spent by a customer with respect to some marketing segmentation.

I am aware this is kind of a freaky trick and not many people may want to use it, so let me know if you think it makes sense to include it or not.

pim-hoeven:
Started working on the GroupedTransformer, which fixes #383 .

Included tests, will now start working on the implementation.

### DIFF
--- a/sklego/meta/__init__.py
+++ b/sklego/meta/__init__.py
@@ -1,18 +1,20 @@
 __all__ = [
-    'ConfusionBalancer',
-    'DecayEstimator',
-    'EstimatorTransformer',
-    'GroupedEstimator',
-    'OutlierRemover',
-    'SubjectiveClassifier',
-    'Thresholder',
-    'OutlierClassifier'
+    "ConfusionBalancer",
+    "DecayEstimator",
+    "EstimatorTransformer",
+    "GroupedEstimator",
+    "GroupedTransformer",
+    "OutlierRemover",
+    "SubjectiveClassifier",
+    "Thresholder",
+    "OutlierClassifier",
 ]
 
 from .confusion_balancer import ConfusionBalancer
 from .decay_estimator import DecayEstimator
 from .estimator_transformer import EstimatorTransformer
 from .grouped_estimator import GroupedEstimator
+from .grouped_transformer import GroupedTransformer
 from .outlier_remover import OutlierRemover
 from .subjective_classifier import SubjectiveClassifier
 from .thresholder import Thresholder
--- a/sklego/meta/_grouped_utils.py
+++ b/sklego/meta/_grouped_utils.py
@@ -9,10 +9,10 @@ from sklearn.utils.validation import _ensure_no_complex_data
 from sklego.common import as_list
 
 
-def split_groups_and_values(
+def _split_groups_and_values(
     X, groups, name="", **kwargs
 ) -> Tuple[pd.DataFrame, np.ndarray]:
-    general_checks(X, name=name)
+    _data_format_checks(X, name=name)
 
     try:
         if isinstance(X, pd.DataFrame):
@@ -24,13 +24,13 @@ def split_groups_and_values(
     except (KeyError, IndexError):
         raise ValueError(f"Could not drop groups {groups} from columns of X")
 
-    X_group = check_grouping_columns(X_group, **kwargs)
-    X_value = check_value_columns(X_value, **kwargs)
+    X_group = _check_grouping_columns(X_group, **kwargs)
+    X_value = check_array(X_value, **kwargs)
 
     return X_group, X_value
 
 
-def general_checks(X, name):
+def _data_format_checks(X, name):
     _ensure_no_complex_data(X)
 
     if issparse(X):  # sklearn.validation._ensure_sparse_format to complicated
@@ -42,12 +42,7 @@ def general_checks(X, name):
         )
 
 
-def check_value_columns(X_value, **kwargs) -> np.ndarray:
-    """Do basic checks on the value columns"""
-    return check_array(X_value, **kwargs)
-
-
-def check_grouping_columns(X_group, **kwargs) -> pd.DataFrame:
+def _check_grouping_columns(X_group, **kwargs) -> pd.DataFrame:
     """Do basic checks on grouping columns"""
     # Do regular checks on numeric columns
     X_group_num = X_group.select_dtypes(include="number")
--- a/sklego/meta/_grouped_utils.py
+++ b/sklego/meta/_grouped_utils.py
@@ -23,10 +23,6 @@ def split_groups_and_values(
             X_value = np.delete(X, as_list(groups), axis=1)
     except (KeyError, IndexError):
         raise ValueError(f"Could not drop groups {groups} from columns of X")
-    except Exception:
-        # Check if we can leverage check_array for standard exceptions, otherwise raise
-        check_array(X, **kwargs)
-        raise
 
     X_group = check_grouping_columns(X_group, **kwargs)
     X_value = check_value_columns(X_value, **kwargs)
--- a/sklego/meta/_grouped_utils.py
+++ b/sklego/meta/_grouped_utils.py
@@ -1,0 +1,66 @@
+from typing import Tuple
+
+import numpy as np
+import pandas as pd
+from scipy.sparse import issparse
+from sklearn.utils import check_array
+from sklearn.utils.validation import _ensure_no_complex_data
+
+from sklego.common import as_list
+
+
+def split_groups_and_values(
+    X, groups, name="", **kwargs
+) -> Tuple[pd.DataFrame, np.ndarray]:
+    general_checks(X, name=name)
+
+    try:
+        if isinstance(X, pd.DataFrame):
+            X_group = X.loc[:, as_list(groups)]
+            X_value = X.drop(columns=groups).values
+        else:
+            X_group = pd.DataFrame(X[:, as_list(groups)])
+            X_value = np.delete(X, as_list(groups), axis=1)
+    except (KeyError, IndexError):
+        raise ValueError(f"Could not drop groups {groups} from columns of X")
+    except Exception:
+        # Check if we can leverage check_array for standard exceptions, otherwise raise
+        check_array(X, **kwargs)
+        raise
+
+    X_group = check_grouping_columns(X_group, **kwargs)
+    X_value = check_value_columns(X_value, **kwargs)
+
+    return X_group, X_value
+
+
+def general_checks(X, name):
+    _ensure_no_complex_data(X)
+
+    if issparse(X):  # sklearn.validation._ensure_sparse_format to complicated
+        raise ValueError(f"The estimator {name} does not work on sparse matrices")
+
+    if X.ndim == 1 or X.shape[1] < 2:
+        raise ValueError(
+            f"0 feature(s) (shape={X.shape}) while a minimum of 2 is required."
+        )
+
+
+def check_value_columns(X_value, **kwargs) -> np.ndarray:
+    """Do basic checks on the value columns"""
+    return check_array(X_value, **kwargs)
+
+
+def check_grouping_columns(X_group, **kwargs) -> pd.DataFrame:
+    """Do basic checks on grouping columns"""
+    # Do regular checks on numeric columns
+    X_group_num = X_group.select_dtypes(include="number")
+    if X_group_num.shape[1]:
+        check_array(X_group_num, **kwargs)
+
+    # Only check missingness in object columns
+    if X_group.select_dtypes(exclude="number").isnull().any(axis=None):
+        raise ValueError("X has NaN values")
+
+    # The grouping part we always want as a DataFrame with range index
+    return X_group.reset_index(drop=True)
--- a/sklego/meta/grouped_transformer.py
+++ b/sklego/meta/grouped_transformer.py
@@ -13,7 +13,8 @@ class GroupedTransformer(BaseEstimator, TransformerMixin):
     and transforms remaining columns using the transformers corresponding to the groups.
 
     :param transformer: the transformer to be applied per group
-    :param groups: the column(s) of the matrix/dataframe to select as a grouping parameter set
+    :param groups: the column(s) of the matrix/dataframe to select as a grouping parameter set. If None,
+                   the transformer will be applied to the entire input without grouping
     :param use_global_model: Whether or not to fall back to a general transformation in case a group
                              is not found during `.transform()`
     """
@@ -72,14 +73,17 @@ class GroupedTransformer(BaseEstimator, TransformerMixin):
         """
         self.__check_transformer()
 
-        X_group, X_value = split_groups_and_values(X, self.groups, **self._check_kwargs)
-
         self.fallback_ = None
+
+        if self.groups is None:
+            self.transformers_ = clone(self.transformer).fit(X, y)
+            return self
+
+        X_group, X_value = split_groups_and_values(X, self.groups, **self._check_kwargs)
+        self.transformers_ = self.__fit_grouped_transformer(X_group, X_value)
 
         if self.use_global_model:
             self.fallback_ = clone(self.transformer).fit(X_value)
-
-        self.transformers_ = self.__fit_grouped_transformer(X_group, X_value)
 
         return self
 
@@ -127,6 +131,9 @@ class GroupedTransformer(BaseEstimator, TransformerMixin):
         :param X: Array-like with columns corresponding to the ones in .fit()
         """
         check_is_fitted(self, ["fallback_", "transformers_"])
+
+        if self.groups is None:
+            return self.transformers_.transform(X)
 
         X_group, X_value = split_groups_and_values(X, self.groups, **self._check_kwargs)
 
--- a/sklego/meta/grouped_transformer.py
+++ b/sklego/meta/grouped_transformer.py
@@ -1,0 +1,23 @@
+from sklearn.base import BaseEstimator, TransformerMixin
+
+
+class GroupedTransformer(BaseEstimator, TransformerMixin):
+    """
+    Construct a transformer per data group. Splits data by groups from single or multiple columns
+    and transforms remaining columns using the transformers corresponding to the groups.
+
+    :param transformer: the transformer to be applied per group
+    :param groups: the column(s) of the matrix/dataframe to select as a grouping parameter set
+    :param use_global_model: Whether or not to fall back to a general transformation in case a group
+                             is not found during `.transform()`
+    """
+
+    def __init__(self, transformer, groups=0, use_global_model=True):
+        self.transformer = transformer
+        self.groups = groups
+
+    def fit(self, X, y=None):
+        return self
+
+    def transform(self, X, y=None):
+        return X
--- a/sklego/meta/grouped_transformer.py
+++ b/sklego/meta/grouped_transformer.py
@@ -43,7 +43,10 @@ class GroupedTransformer(BaseEstimator, TransformerMixin):
             # This one is needed because of line #918 of sklearn/utils/estimator_checks
             raise TypeError("argument must be a string, date or number")
 
-        if y:
+        if y is not None:
+            if isinstance(y, pd.Series):
+                y.index = X_group.index
+
             grouped_transformers = {
                 # Fit a clone of the transformer to each group
                 group: self.__fit_single_group(group, X_value[indices, :], y[indices])
@@ -80,7 +83,7 @@ class GroupedTransformer(BaseEstimator, TransformerMixin):
             return self
 
         X_group, X_value = split_groups_and_values(X, self.groups, **self._check_kwargs)
-        self.transformers_ = self.__fit_grouped_transformer(X_group, X_value)
+        self.transformers_ = self.__fit_grouped_transformer(X_group, X_value, y)
 
         if self.use_global_model:
             self.fallback_ = clone(self.transformer).fit(X_value)
--- a/sklego/meta/grouped_transformer.py
+++ b/sklego/meta/grouped_transformer.py
@@ -4,7 +4,7 @@ import pandas as pd
 from sklearn.base import BaseEstimator, TransformerMixin, clone
 from sklearn.utils.validation import check_is_fitted
 
-from ._grouped_utils import split_groups_and_values
+from ._grouped_utils import _split_groups_and_values
 
 
 class GroupedTransformer(BaseEstimator, TransformerMixin):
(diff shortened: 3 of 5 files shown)


sklego/meta/__init__.py
```diff
@@ -1,18 +1,20 @@
...
-    'OutlierClassifier'
+    "ConfusionBalancer",
+    "DecayEstimator",
+    "EstimatorTransformer",
+    "GroupedEstimator",
+    "GroupedTransformer",
+    "OutlierRemover",
+    "SubjectiveClassifier",
+    "Thresholder",
+    "OutlierClassifier",
```
pim-hoeven, line 10: Black did this, hope it does not lead to merge conflicts somewhere else

tests/test_meta/test_grouped_transformer.py
```diff
@@ -0,0 +1,203 @@
...
+        "check_fit2d_1feature",
+        "check_fit2d_predict1d",
+    ]
+
+
+@pytest.mark.parametrize(
+    "test_fn", flatten([transformer_checks, nonmeta_checks, general_checks])
+)
+def test_estimator_checks(test_fn, tests_to_skip):
+    if test_fn.__name__ not in tests_to_skip:
```
pim-hoeven, line 28: What do you think of this approach for skipping tests from the list? I think it looks cleaner than copying the other list and filtering the ones you don't want, do you agree? Otherwise I'll change it.

One downside is that we get some 'false positives', the tests that are skipped are counted as passed test.

sklego/meta/grouped_transformer.py
```diff
@@ -0,0 +1,153 @@
...
+    Construct a transformer per data group. Splits data by groups from single or multiple columns
+    and transforms remaining columns using the transformers corresponding to the groups.
+
+    :param transformer: the transformer to be applied per group
+    :param groups: the column(s) of the matrix/dataframe to select as a grouping parameter set
+    :param use_global_model: Whether or not to fall back to a general transformation in case a group
+                             is not found during `.transform()`
+    """
+
+    def __init__(self, transformer, groups=0, use_global_model=True):
```
koaning, line 21: Groups is something that the user *must* supply. It'd be nicer if it weren't initialized with 0 with this in mind.

tests/test_meta/test_grouped_transformer.py
koaning: I actually *really* like this, it feels much cleaner to me. @MBrouns?

sklego/meta/grouped_transformer.py
```diff
@@ -0,0 +1,153 @@
...
+        else:
+            X_group = check_array(X)[:, as_list(self.groups)]
+            if np.isnan(X_group).sum(axis=None) > 0:
+                missings = True
+
+        if missings:
+            raise ValueError("X has NaN values")
+
+        # The grouping part we always want as a DataFrame with range index
+        return pd.DataFrame(X_group).reset_index(drop=True)
```
koaning, line 54: Are these methods the exact same as in the `GroupedEstimator`? Since the methods are static, they might as well be functions. This might be nicer with regards to both unit testing as well as less code.

tests/test_meta/test_grouped_transformer.py
```diff
@@ -0,0 +1,203 @@
...
+
+    # Make sure all groups are present
+    groups = np.repeat([0, 1], repeats=(X.shape[0] + 1) // 2)[:X.shape[0], np.newaxis]
+    X_with_groups = np.concatenate([groups, X], axis=1)
+
+    # Some weird interval to make sure we test the right values
+    scaling_range = (13, 42)
+
+    trf = MinMaxScaler(scaling_range)
+    transformer = GroupedTransformer(clone(trf), groups=0)
```
koaning, line 45: I wonder ... why clone here?

tests/test_meta/test_grouped_transformer.py
MBrouns: It does feel good, but i'd rather have it inside the `parametrize` than i n the test function itself. something like:

@pytest.mark.parametrize(
    "test_fn", tests(
         include=[transformer_checks, nonmeta_checks, general_checks], 
         exclude=["check_fit2d_1feature", "check_fit2d_predict1d"]
    )
)
```

File: sklego/meta/grouped_transformer.py (hunk 3 above)
Comment (MBrouns) [reply sklego/meta/grouped_transformer.py:21]: I think `sklearn` expects that an estimator can be initialised without any arguments specified by the user, although I'm not sure how strictly that is enforced. would a `groups=None` make sense to have a no grouping at all option as a default?

Comment (koaning) [reply sklego/meta/grouped_transformer.py:21]: Ah yeah you're right. `None` would be better. That said, `meta` estimators *must* receive an estimator in order for it to work.

File: sklego/meta/grouped_transformer.py [hunk 6]
```diff
@@ -0,0 +1,153 @@
...
+            pd.concat(
+                [
+                    self.__transform_single_group(group, X_value.loc[indices, :])
+                    for group, indices in group_indices.items()
+                ],
+                axis=0,
+            )
+            .sort_index()
+            .values
+        )
```
koaning, line 140: Is the output here always a pandas dataframe? Even when the input is an nd-array?

tests/test_meta/test_grouped_transformer.py
```diff
@@ -0,0 +1,132 @@
...
+    transformed = transformer.fit(X_with_groups, y).transform(X_with_groups)
+
+    assert transformed.shape == X.shape
```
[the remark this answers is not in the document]
koaning: I think the shape is already tested in the standard sklearn checks. I don't mind it here, but feel free to remove.

tests/test_meta/test_grouped_transformer.py
```diff
@@ -0,0 +1,203 @@
...
+@pytest.mark.parametrize(
+    "test_fn", flatten([transformer_checks, nonmeta_checks, general_checks])
+)
+def test_estimator_checks(test_fn, tests_to_skip):
+    if test_fn.__name__ not in tests_to_skip:
+        trf = GroupedTransformer(StandardScaler())
+        test_fn(GroupedTransformer.__name__, trf)
+
+
+def test_values(random_xy_dataset_clf):
```
koaning, line 33: nitpick: maybe a more descriptive name for the test.

tests/test_meta/test_grouped_transformer.py
```diff
@@ -0,0 +1,203 @@
...
+
+    assert transformed.shape == X.shape
+
+    df_with_groups = pd.concat([pd.Series(groups.flatten(), name="G"), pd.DataFrame(transformed)], axis=1)
+
+    assert np.allclose(df_with_groups.groupby("G").min(), scaling_range[0])
+    assert np.allclose(df_with_groups.groupby("G").max(), scaling_range[1])
+
+
+def test_group_correlation_standardscaler(random_xy_dataset_clf):
```
koaning, line 56: I'm a bit unsure what this test is doing. In part because I wonder what you're covering here by using correlation but also because the name of the test suggests a standardscaler while there is none in the code.

tests/test_meta/test_grouped_transformer.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def test_missing_groups_transform_noglobal(random_xy_dataset_clf):
+    X, y = random_xy_dataset_clf
+    X, y = check_X_y(X, y)
+
+    # Make sure all groups are present
+    groups = np.repeat([0, 1], repeats=(X.shape[0] + 1) // 2)[:X.shape[0], np.newaxis]
+    X_with_groups = np.concatenate([groups, X], axis=1)
+
```
koaning, line 189: You're currently using the `random_xy_dataset_clf` which is fine, but I don't think you really need it for the tests here. I also think that you can cover the same ground by introducing a fixture in this file that generates `X, y, groups, X_with_groups`. That way we won't need this code block everywhere:

```
    X, y = random_xy_dataset_clf
    X, y = check_X_y(X, y)

    # Make sure all groups are present
    groups = np.repeat([0, 1], repeats=(X.shape[0] + 1) // 2)[:X.shape[0], np.newaxis]
    X_with_groups = np.concatenate([groups, X], axis=1)
```


Would result in less code while still testing the same material. 

It should also massively speed up the tests.

pim-hoeven:
@koaning ; thanks! @MBrouns thought me well


Regarding the edge case, `StandardScaler` handles this well by using `std=1` in case of size 1:
```
>>> StandardScaler().fit(np.array([[1, 2, 3]])).transform(np.array([[1, 2, 3]]))
... array([[0., 0., 0.]])
````
So in that specific case we're fine. I could add a general exception catcher to the `.fit` that catches everything and raises a special exception in the case that the group has size 1?

File: sklego/meta/grouped_transformer.py (hunk 4 above)
Comment (pim-hoeven) [reply sklego/meta/grouped_transformer.py:54]: Unfortunately not. I think that the setup in `GroupedTransformer` is way cleaner and easier than the one in `GroupedEstimator` (from the start splitting the data in a grouper-part and a values-part vs keeping everything together). The check on the values part is just a regular `check_array`. 

The biggest difference is that shrinkage can add an extra column to the grouper-part. 

I would say opening a new issue to make both in the 'new style' would be better than doing it in this PR, do you agree?

File: tests/test_meta/test_grouped_transformer.py (hunk 5 above)
Comment (pim-hoeven) [reply tests/test_meta/test_grouped_transformer.py:45]: I think the first time I did this my brain was turned of. The other times I copied the line. Will get rid of it :)

File: sklego/meta/grouped_transformer.py (hunk 6 above)
Comment (pim-hoeven) [reply sklego/meta/grouped_transformer.py:140]: Yes, in this specific line because it is the result of `pd.concat`. In general because I use the indices to keep track of the order of rows over groups (to make sure that input and output are in the same order, not sorted by group).

File: tests/test_meta/test_grouped_transformer.py (hunk 9 above)
Comment (pim-hoeven) [reply tests/test_meta/test_grouped_transformer.py:56]: Forgot to change the name, should be `test_group_correlation_minmaxscaler`. Basically, if the transformer as we want, we expect that the correlation between the original value and the transformed value within each group is 1. This ensures that the order within groups of the result is the same as the input order

Comment (pim-hoeven) [reply tests/test_meta/test_grouped_transformer.py:56]: (Renamed the test)

File: tests/test_meta/test_grouped_transformer.py (hunk 10 above)
Comment (pim-hoeven) [reply tests/test_meta/test_grouped_transformer.py:189]: Good point, I agree. Will change it.

**Comment by koaning** [comment]
MHmm. How about this: if the transformer fails for a certain group, we try to catch the error and tell the user what group there was a failure? I can also imagine that we might have an edge case with a group that has no variance ... I'd be nice if we can do something to help a user debug in such a scenario.

File: sklego/meta/grouped_transformer.py (hunk 4 above)
Comment (koaning) [reply sklego/meta/grouped_transformer.py:54]: Let's keep it as is for now then. That grouped estimator needs a big refactor at some point anyway, I'm fine if that happens in a different PR.

**Comment by pim-hoeven** [comment]
@koaning I think indeed it could be a useful feature in general. Added it to the PR!

File: sklego/meta/grouped_transformer.py (hunk 6 above)
Comment (koaning) [reply sklego/meta/grouped_transformer.py:146]: That we use a pandas dataframe as an intermediate result is fine, but it may be safer to respect the input of the user. If a user passes a numpy array into this transfromer they might expect a numpy array to also come out of it. I'm also not 100% sure if all the scikit-learn tools are 100% pandas compatible.

Comment (pim-hoeven) [reply sklego/meta/grouped_transformer.py:146]: Sorry - misread the question. The intermediate result is always a pandas dataframe, the end result is always a numpy array (because of the `.values`). If you want, I can easily adapt it such that the output is a dataframe if the input is. I didn't do it now because most (all?) sklego transformers also return a numpy array.

Comment (koaning) [reply sklego/meta/grouped_transformer.py:146]: If the output remains numpy I think that is fine for now. It's what all the scikit-learn models do anyway. Cool. I'll give it one big final review now.

Comment (koaning) [reply sklego/meta/grouped_transformer.py:146]: Ah yeah. I only now see the `.values`.

File: sklego/meta/grouped_transformer.py [hunk 11]
```diff
@@ -0,0 +1,159 @@
...
+        """
+        Fit the transformers to the groups in X
+
+        :param X: Array-like with at least two columns, of which at least one corresponds to groups defined in init,
+                  and the remaining columns represent the values to transform.
+        :param y: (Optional) target variable
+        """
+        self.__check_transformer()
+
+        X_value = self.__check_value_columns(X)
```
koaning, line 99: Yep. Found a bug here.

```python
from sklego.datasets import load_penguins
from sklego.meta import GroupedEstimator
from sklearn.preprocessing import StandardScaler

X = load_penguins()['data']
meta = GroupedTransformer(StandardScaler(), groups=[0])
meta.fit_transform(X[:, :3])
```

If you run this you'll get an error. Here's the context, this is what `X[:, :3]` looks like;

```python
array([['Torgersen', 39.1, 18.7],
       ['Torgersen', 39.5, 17.4],
       ['Torgersen', 40.3, 18.0],
       ...,
       ['Dream', 49.6, 18.2],
       ['Dream', 50.8, 19.0],
       ['Dream', 50.2, 18.7]], dtype=object)
```

You'll notice that column 0 (the one I want to group over) is still being checked as if it isn't a variable that I am grouping over. In this case, grouping over a string should be fine but I'm getting a 

```
ValueError: could not convert string to float: 'Torgersen'
```

pim-hoeven: This one was terrible. Didn't know about arrays with mixed types, but I'll steer clear of them as much as possible. 
I don't think my solution is very nice, so if you have ideas for improvements let me know!

koaning: I think there's still a similar bug. 

```python
from sklego.datasets import load_penguins
from sklego.meta import GroupedTransformer, EstimatorTransformer
from sklearn.preprocessing import StandardScaler

tfm = EstimatorTransformer(StandardScaler(), predict_func='predict')
meta = GroupedTransformer(StandardScaler(), groups=[0, -1])

df = load_penguins(as_frame=True).dropna()
X = df.drop(columns='species').values
```

This is what the dataset looks like; 

```python
array([['Torgersen', 39.1, 18.7, 181.0, 3750.0, 'male'],
       ['Torgersen', 39.5, 17.4, 186.0, 3800.0, 'female'],
       ['Torgersen', 40.3, 18.0, 195.0, 3250.0, 'female'],
       ...,
       ['Dream', 49.6, 18.2, 193.0, 3775.0, 'male'],
       ['Dream', 50.8, 19.0, 210.0, 4100.0, 'male'],
       ['Dream', 50.2, 18.7, 198.0, 3775.0, 'female']], dtype=object)
```

Alas, we still get an error. 

```python
meta.fit_transform(X)
# ValueError: could not convert string to float: 'male'
```

Might be good to add this as a unit-test. It seems to work fine if `groups=[0, 5]` but it fails when `groups=[0, -1]`.

koaning: I've found another interesting edge case. 

```
from sklego.datasets import load_penguins
from sklego.meta import GroupedTransformer, EstimatorTransformer
from sklearn.preprocessing import StandardScaler

tfm = EstimatorTransformer(StandardScaler(), predict_func='predict')
meta = GroupedTransformer(StandardScaler(), groups=[0, 5])

df = load_penguins(as_frame=True).dropna()
X = df.drop(columns='species').values
meta.fit_transform(X[:, 3])
```

It gives a reshape error instead of a "the dataframe you pass in has three columns, we can't select column 5".

pim-hoeven: The first one seems to have been resolved by the last commit(`Grouping on string columns in numpy arrays`, because I don't have this error. I will implement a test though for this specific case.

pim-hoeven: The other edge case is a typo (I think?), you have `X[:, 3]`, so selecting only the third column. If I do `X[:, :3]` it does raise an IndexError

pim-hoeven:
Just added some more tests because I saw in the coverage report that I didn't test the dataframe

tests/test_meta/test_grouped_transformer.py
```diff
@@ -0,0 +1,326 @@
...
+        meta.fit_transform(X)
+
+
+def test_array_with_multiple_string_cols(penguins):
+    X = penguins
+
+    meta = GroupedTransformer(StandardScaler(), groups=[0, -1])
+
+    # This should work fine
+    meta.fit_transform(X)
```
koaning, line 289: Let's add an `assert` here. I've noticed that we still haven't checked for the output shape. Would be good to do here.

tests/test_meta/test_grouped_transformer.py
```diff
@@ -0,0 +1,326 @@
...
+@pytest.fixture(scope="module")
+def penguins(penguins_df):
+    return penguins_df.values
+
+
+def test_df(penguins_df):
+    meta = GroupedTransformer(StandardScaler(), groups=["island", "sex"])
+
+    # This should work fine
+    meta.fit_transform(penguins_df)
```
koaning, line 269: Let's add an assert here. I've noticed that we still haven't checked for the output shape. Would be good to do here.

tests/test_meta/test_grouped_transformer.py
```diff
@@ -0,0 +1,326 @@
...
+    transformer = GroupedTransformer(trf, groups=0, use_global_model=False)
+    transformer.fit_transform(X)
+
+
+@pytest.fixture(scope="module")
+def penguins_df():
+    df = load_penguins(as_frame=True).dropna()
+    X = df.drop(columns='species')
+
+    return X
```
koaning, line 257: Nitpick, let's keep all the fixtures close to eachother at the top of this file.

tests/test_meta/test_grouped_transformer.py
```diff
@@ -0,0 +1,326 @@
...
+        def fit(self, X, y=None):
+            if len(X) <= 1:
+                raise ValueError("Need more than 1 value to fit")
+
+            return self
+
+        def transform(X):
+            return X
+
+    return MultipleObsFitter()
```
koaning, line 220: Let's move this fixture on top just like the rest.

sklego/meta/_grouped_utils.py
```diff
@@ -0,0 +1,62 @@
...
+import numpy as np
+import pandas as pd
+from scipy.sparse import issparse
+from sklearn.utils import check_array
+from sklearn.utils.validation import _ensure_no_complex_data
+
+from sklego.common import as_list
+
+
+def split_groups_and_values(
```
koaning, line 12: Is this a function that the user is supposed to touch? If not, might be better to implement them as `_split_groups_and_values`. 

Out of curiosity, do we use these functions in multiple places or just in the groupedtransformer? If it's the latter, is there a particular reason why you felt it better to have them in a separate file?

sklego/meta/_grouped_utils.py
```diff
@@ -0,0 +1,62 @@
...
+
+    if X.ndim == 1 or X.shape[1] < 2:
+        raise ValueError(
+            f"0 feature(s) (shape={X.shape}) while a minimum of 2 is required."
+        )
+
+
+def check_value_columns(X_value, **kwargs) -> np.ndarray:
+    """Do basic checks on the value columns"""
+    return check_array(X_value, **kwargs)
```
koaning, line 47: Doesn't this function basically re-implement `check_array`?

sklego/meta/_grouped_utils.py
```diff
@@ -0,0 +1,62 @@
...
+    except (KeyError, IndexError):
+        raise ValueError(f"Could not drop groups {groups} from columns of X")
+
+    X_group = check_grouping_columns(X_group, **kwargs)
+    X_value = check_value_columns(X_value, **kwargs)
+
+    return X_group, X_value
+
+
+def general_checks(X, name):
```
koaning, line 33: `_ensure_non_sparse` might be a better name.

sklego/meta/_grouped_utils.py
pim-hoeven: Alright, will add the underscores.

I put them here because I think in a different PR it would be nice to use these for the `GroupedEstimator` as well (#390), do you prefer to keep them in this file for now?

sklego/meta/_grouped_utils.py
pim-hoeven: Yes right now it does, I put it this way for consistency in:
```
X_group = check_grouping_columns(X_group, **kwargs)
X_value = check_value_columns(X_value, **kwargs)
```
Would you prefer to have
```
X_group = check_grouping_columns(X_group, **kwargs)
X_value = check_array(X_value, **kwargs)
```
?

sklego/meta/_grouped_utils.py
pim-hoeven: Renamed to `data_format_checks`, because it also checks non-complexity and shape (grouping should have at least two columns)

sklego/meta/_grouped_utils.py
koaning: This is fine. Given the future plan I can certainly see the merits.

sklego/meta/_grouped_utils.py
koaning: The latter feels better yeah. Mainly because having another function implies that we're treating it differently.

pim-hoeven: Alright, changed it!

koaning:
I've decided to merge this. @pim-hoeven well done! 

This feature is big enough to warrant a new release, however I would like to check if you intend on picking up https://github.com/koaning/scikit-lego/issues/390 in the short term? We also don't have any documentation on this tool yet and it might be good to add before doing another release. 

Let me know. I can also work on the documentation if you prefer this to be released quickly.

pim-hoeven:
I will start working on #390 this week, but on the same pace as I worked on this one, so will probably take a week or two. If you have anything specific in mind for the documentation then go ahead, otherwise I'll start working on it once 390 is fixed.

koaning:
I'll probably make a quick example this week. Will keep you in the loop.

PR #388, opened 2020-07-17
Make CVXPY optional by adding NotInstalledPackage class

### LINKED ISSUE #380: [FEATURE] A light version that does not depend on cvxpy (and others?)
Please explain clearly what you'd like to see added. 

- [x] convince us of the use-case, we're open to many suggestions but we prefer to solve problems with pipelines that are at least somewhat general 
**When using the package in a Docker container without C installed, installation (can?) fail on CVXPY**

- [x] ~~add a screenshot if applicable (ML stuff is hard to explain with words, pictures say 1000 words)~~
- [x] ~~make sure that the feature you want is not already supported by sklearn~~

Happy to work on this if you agree

pim-hoeven:
Fixes #380


pim-hoeven:
@koaning  @MBrouns 
For now I moved `cvxpy` to an extra, but not really sure about it. If this is the way you want to move forward I'll change the documentation and things such as workflows. Will try to come up with a different solution, but let me know if you guys think of something!

pim-hoeven:
Any thoughts @koaning @MBrouns?

koaning:
Hi @pim-hoeven. Sorry for the long delay. I was waiting until @MBrouns got back from holiday and then I got distracted. I'm still interested in this feature but there's a few things about your PR that I'd like to adress.

The main thing is that it may be better to notify the user to this command; 

```
python -m pip install scikit-lego[cvxpy]
```

As opposed to sending them to `python -m pip install cvxpy`. The main reason is that as a package, we have slightly more control and we can send users to a documentation page that explains why this is needed. It'd also be nice if we could introduce a command such that a user can do; 

```
python -m pip install scikit-lego[all] 
```

@pim-hoeven I kind of want to issue fixed this week because this bug is now breaking installs on my local mac. If you don't have the time for it this week let me know and I'll pick this up and make a new release.

koaning:
This is also a VeryStrangeError[tm] in CI. @pim-hoeven did you see that error before?

pim-hoeven:
I have some time to work on this, think only some small things are left so should be fine! 

Your remark about pip install is about the error message in the NotInstalledClass, right? I'll fix that in a bit. 

Regarding the error: there used to be an error because cvxpy wasn't installed and so all related tests filled. I fixed that in the latest commit. The more recent package is indeed very strange, will take a look at it later today.

pim-hoeven:
🎉  found the issue: https://github.com/pytest-dev/pytest-xdist/issues/580. 

There are two possible solutions:
1. Fix pytest-xdist on 1.34.0
2. Upgrade pytest (which is currently fixed at 5.4.1)
Which one do you prefer? (I currently opted for number 1, didn't check number 2 yet)

pim-hoeven:
Fixed your remarks, looking forward to the rest of the review

koaning:
This is nice, thanks!, but I'd like to add an extra test (lesson learned from another repo). Something similar to [this](https://github.com/RasaHQ/whatlies/blob/master/.github/workflows/dependencies.yml). 

It's a separate workflow where the idea is that we try to import all sorts of things, even if the cvxpy dependency is missing. 

No errors should be raised when importing the `PatsyTransformer` for example even if cvxpy is not there. This is a test that is currently missing but crucial to future development.

pim-hoeven:
Something similar became something completely stolen, hope you don't mind. It's a very nice workflow! I never worked with those  before, so very glad to learn about them here, thanks!

pim-hoeven:
Oops - forget to update the docs build. 

Is it necessary to have it anyway? Since we already have a separate workflow for it anyway?

koaning:
I might also add a note to the installation documentation that we now support dependency installs. That way we can also have the error message send people to our documentation where we can explain this in more detail.

pim-hoeven:
Just the installation documentation or readme as well?

koaning:
Feel free to do both, but mainly the docs. That way your error message can properly link to it.

pim-hoeven:
Alright, I added it to the docs only because then people don't need to bother with the extra's unless they run into an issue

koaning:
Sweet. I'm going down to cook now. Will give it another, hopefully final, review after dinnen 👍 thanks for the prompt reply!

koaning:
Strange. It seems like the render docs build phase is gone. Did you remove it? It'd be nice to have as a seperate workflow. Sometimes a small change can break the docs but since the docsbuild can be intensive it's not something I'd want to run for every python-version/platform.

pim-hoeven:
That is indeed strange, it's still on master, not in the diff and locally I have it as well. It is on my branch as well: https://github.com/pim-hoeven/scikit-lego/tree/feature/make-cvxpy-optional/.github/workflows.

No clue what went wrong here. Maybe a limit?

pim-hoeven:
This specific check hasn't been run for any of the previous pushes either for some reason. I'll take a look tomorrow!

pim-hoeven:
Not close to my pc, but might be that renderdocs only runs if master gets updated and not if prs are updated?

koaning:
Here's the deal, the changes look good to me. So I'll just merge to master and start fixing any issues from there. But these changes look great!

koaning (approved):

PR #395, opened 2020-07-28
Improving GroupedPredictor

### LINKED ISSUE #390: [IMPROVEMENT] Apply lessons of GroupedTransformer to GroupedEstimator
I feel this isn't really a bug nor a feature, so labelled it improvement.

While working on #387 I noticed that grouping can be implemented cleaner and easier if we separate grouping and value data from the start. There is also quite some overlap which I think should be taken away.

In the process we can deprecate `GroupedEstimator` in favour of `GroupedPredictor`

pim-hoeven:
Closes #390 

- [x] Copy `GroupedEstimator` to `GroupedPredictor`
- [x] Unittests for `GroupedPredictor` instead of `GroupedEstimator`
- [x] Deprecate `GroupedEstimator` in favor of `GroupedPredictor`
- [x] Apply lessons of `GroupedTransformer` to `GroupedPredictor`


sklego/meta/grouped_estimator.py
```diff
@@ -9,43 +10,14 @@
...
-    res = np.zeros(len(group_sizes))
-    res[np.argmin(np.array(group_sizes) >= min_n_obs) - 1] = 1
-    return res
-
-
+@deprecated(
+    version="0.5.2",
+    reason="Please use `sklego.meta.GroupedPredictor` instead. "
+    "This object will be removed from the meta submodule in version 0.7.0.",
+)
```
pim-hoeven, line 20: @koaning Do you agree with the versions?

koaning: Looks good to me.

pim-hoeven:
@koaning Tests pass, so I removed the WIP. Will take another look somewhere this week because I think I can make it a bit cleaner still. Do let me know if you have remarks already :)

pim-hoeven:
@koaning @MBrouns 
I made some small improvements and think it's ready for review!

sklego/meta/grouped_estimator.py
```diff
@@ -107,7 +79,7 @@ def __set_shrinkage_function(self):
             self.shrinkage_function_ = self.shrinkage
```
koaning, line 79: Since we're no longer testing `GroupedEstimator`, might it be preferable if this implementation just points to `GroupedPredictor`? If there's any code improvements to be made, they'll still go to one place that way.

koaning:
There's actually something else that's missing. 

@pim-hoeven you've made some considerable contributions here so I would hereby like to suggest that you also contribute to the `this.py` file. It's a poem with the lessons learned while making this package. If you've got a nice lesson that you've learned (and you're able to get it to rhyme) I'd love to see a small contribution there as well.

sklego/meta/grouped_estimator.py
pim-hoeven: That was my first thought, but I did it like this to prevent things from breaking at the users' end, because in the new version the argument `value_columns` does not exist anymore (instead all non-grouping columns are used, which I believe is more in line with the sklearn API). 

We could also rely on version management on the side of the user of course.

Then I would do:
```
from deprecated import deprecated
from .grouped_predictor import GroupedPredictor

@deprecated(
    version="0.5.2",
    reason="Please use `sklego.meta.GroupedPredictor` instead. "
    "This object will be removed from the meta submodule in version 0.7.0.",
)
def GroupedEstimator(*args, **kwargs):
    return GroupedPredictor(*args, **kwargs)

```

Would you prefer this?

koaning: Now that I think of it, I prefer the previous method. `*args`/`**kwargs` don't play nice with the scikit-learn internals.

Issue #399, opened 2020-07-29
[BUG] GroupedTransformer fails with new numpy version


koaning:
@pim-hoeven it seems like pandas/numpy just did a new release and it seems to break a test. The test involves using `-1` to indicate a column that you'd like to group over. 

Here's what the code currently does.

![image](https://user-images.githubusercontent.com/1019791/88776415-34360600-d186-11ea-9b54-115cde39cb29.png)

Here's what it should do. 

![image](https://user-images.githubusercontent.com/1019791/88776475-4a43c680-d186-11ea-84e3-033315a08c1e.png)

I'm wondering what the best course of action is to deal with this and figured sending a ping here. Something tells me that dropping support for `-1` would be something I could live with for now. Another alternative is that we use pandas to fix this.

PR #400, opened 2020-07-29
Fixed tests for Grouped Transformer

koaning:
@pim-hoeven this fix might be good enough?

### DIFF
--- a/.github/workflows/schedule-dependencies.yml
+++ b/.github/workflows/schedule-dependencies.yml
@@ -35,5 +35,4 @@ jobs:
     - name: Test with pytest
       run: |
         pip install -e ".[test]"
-        pip install pytest
         pytest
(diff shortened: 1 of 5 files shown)


koaning:
It's for this issue: https://github.com/koaning/scikit-lego/issues/399.

pim-hoeven:
I'll rebase my PR as soon as you merge this.

koaning:
Mhm. There also seems to be something up with the logging.

koaning:
I didn't come up with this one ... this seems to be a `pytest` bug. They just released version 6.0.0 and it seems to have a broken `caplog` fixture. Nothing gets recorded. Will move to an old version of pytest in our build tools I suppose. 

Can confirm locally that the issue does not exist for `pytest==5.4.1`.

Issue #389, opened 2020-07-23
[FEATURE] SVD Imputer


koaning:
This [tweet](https://twitter.com/WomenInStat/status/1285614472065953793) makes a good point. SVD could be an amazing method to impute missing values.

pim-hoeven:
I can start on this today?

koaning:
I'll be closing this for now. There's a package that already has this feature: https://github.com/iskandr/fancyimpute.

PR #404, opened 2020-07-30
Feature/svd imputer

pim-hoeven:
Adding feature #389 

Imputation process
1. Fill missing elements (e.g. with column mean)
2. Compute SVD to get rank-K approximation
    $$X \approx D_{11} U_1 V_1^T + D_{22} U_2 V_2^T + \ldots +  D_{KK} U_K V_K^T$$
3. Replace filled elements with rank-K approximation
4. Repeat 2 and 3 until convergence


pim-hoeven:
So, I started working on this but a number of things I would like to discuss. I opened a PR such that you can play around a bit.

To use it:
```
X = get_X()

svdi = SVDImputer(2)

svdi.fit(X)
```


**What to fit/transform**
Basically, I understand the concept if we only transform the train set: we calculate SVD, take product of first k vectors as approximation and repeat until convergence.
How to apply this to a new dataset though? My thoughts:
- Do the fitting 'freshly' on the new dataset, i.e. only in `.transform()`. That would fail if we only have one row, or a column with only missings
- Do something with the filled values in fit (per column), e.g. take the mean? But what does this mean mathematically?


**Why does it always converge after at most 2 steps?**
This was quite surprising to me, I tried a number of combinations for X (different shapes, distributions, percentage missing etc), but in all cases it does one or two iterations and then stops.


**How to tests**
Is there a nice way to know what to expect if we call `fit_transform()` on X? Obviously no more missing values, but what about the values themselves? They should probably be 'close' to the values that we already have?

I'll dive into this a bit deeper as well, but let me know if you have any thoughts :)

tests/test_impute.py
```diff
@@ -0,0 +1,21 @@
...
+from sklego.impute import SVDImputer
+
+
+@pytest.fixture
+def X_test():
+    np.random.seed(42)
+    return np.random.normal(size=(100, 5))
+
+
+def test_get_kth_approximation_shape(X_test):
```
koaning, line 13: We've got a continous dataset fixture that comes in all shapes and sizes take could be used here.

koaning:
## How to fit/transform

Haha. I actually just posted after reading the tweet thinking "this sounds useful, I should think about it someday". Then you picked up the issue. So I'll need to think about this now too 😄 . 

## Convergence

I've tried reading the paper that they refer to but it is *dense*. They do some stochastic tricks to allow for sparse data and it does indeed converge quickly. I'm not very knowledge-able on this, but I'm willing to assume scikit-learn as a given. 

## How to Test 

Testing machine learning models is really tricky but one thing you can always do is test towards an "obvious case". Say, there's a 2 column dataframe and there is a 1-1 mapping between the two columns. SVD should be able to capture this, so there's a test case. 

I think what you're doing with testing some of the internal object functionality is also a nice step. Especially if you'd like to confirm that a hyperparameter behaves as expected, that's also a common approach.

koaning:
For the new dataset, you'd take the train dataset and append it to test. This gives a new "big" dataset, from which you can learn the most. You apply the transformation to the big dataset and then you only return the rows belonging to the test set. 

This means that "training" effectively means "just store the train set for use-later".

pim-hoeven:
Just a small update on something that I've been kinda stuck on.

I tried to implement the test that you suggested, but it does not work as expected. Even with a very low number of missing values, which I find very strange. I'll take a look again this Thursday, but maybe you spot something immediately (or @MBrouns ?)

With this code and all combinations of variables that I tried, I keep getting outliers (see image below). It stops every time after two iterations, in other words: the second run yields the same result as the first run.

Not sure if it's expected behavior (I think not), or I'm doing something wrong (I think yes)
```
import matplotlib.pyplot as plt
import numpy as np

np.random.seed(42)

from sklego.impute import SVDImputer

# Change these
n, k, factor, p = (10_000, 2, 2, 0.0001)
k_rank = k-1


X1 = np.random.normal(0, 1, size=(n, k))
X2 = (factor * X1).sum(axis=1)

mask_indices = np.random.choice([True, False], size=X2.shape, p=(p, 1 - p))
X2[mask_indices] = np.nan

print(f"Number of masked values: {sum(mask_indices)}")

X = np.concatenate([X1, X2[:, np.newaxis]], axis=1)

svdi = SVDImputer(k_rank)

X_transformed = svdi.fit_transform(X)

plt.scatter(X_transformed[:, :k].sum(axis=1), X_transformed[:, k])
```

image

koaning:
You're not getting outliers ... you're getting something that is near the mean value. 

I've changed the code slightly to get a better impression of what is happening. 

```python
import numpy as np
import matplotlib.pyplot as plt
from sklego.impute import SVDImputer

np.random.seed(42)

# Change these
n, k, factor, p = (10_000, 2, 2, 0.01)
k_rank = k-1


X1 = np.random.normal(0, 1, size=(n, k))
X2 = (factor * X1).sum(axis=1)

mask_indices = np.random.choice([True, False], size=X2.shape, p=(p, 1 - p))
X2[mask_indices] = np.nan

print(f"Number of masked values: {sum(mask_indices)}")

X = np.concatenate([X1, X2[:, np.newaxis]], axis=1)

svdi = SVDImputer(k_rank)

X_transformed = svdi.fit_transform(X)

plt.scatter(X_transformed[:, :k].sum(axis=1), X_transformed[:, k], c=mask_indices)
```

The yellow dots are the ones that are masked and fixed. 

image

koaning:
I did just get something really similar to work by using `PCA` instead. 

```python
x1 = np.random.normal(0, 1, size=(n, 1))
x2 = x1 + np.random.normal(0, 0.2, size=(n, 1))

X = np.concatenate([x1, x2], axis=1)
mask_indices = np.random.choice([True, False], size=X.shape, p=(p, 1 - p))
X[mask_indices] = np.nan

from sklearn.decomposition import PCA

means = np.nanmean(X, axis=0)
missing_idx = np.where(np.isnan(X))
X[missing_idx] = np.take(means, missing_idx[1])

x_train = X[(mask_indices == False).sum(axis=1) == 2, :]
pca = PCA(1).fit(x_train)
X_imp = pca.inverse_transform(pca.transform(X))

X_plt = X_imp[mask_indices.sum(axis=1) > 0, :]
plt.scatter(x_train[:, 0], x_train[:, 1])
plt.scatter(X_plt[:, 0], X_plt[:, 1])
```

image

pim-hoeven:
Thanks for your remarks! Outliers wasn't the best word to describe the thing I was seeing, but I meant the yellow dots indeed. I think the biggest difference between your PCA and SVD approach is in the way outliers are replaced: in the SVD, only the missing values are replaced. Your PCA approach replaces all values in a row that have at least one missing value. If I take the SVD approach and apply it to your PCA code, the result is similar to the SVD result (note that there are two lines in this case because both columns have missings):

```
import numpy as np
import matplotlib.pyplot as plt
from sklearn.decomposition import PCA

n = 10_000
p = 0.01

x1 = np.random.normal(0, 1, size=(n, 1))
x2 = x1 + np.random.normal(0, 0.2, size=(n, 1))

X = np.concatenate([x1, x2], axis=1)
mask_indices = np.random.choice([True, False], size=X.shape, p=(p, 1 - p))
X[mask_indices] = np.nan

means = np.nanmean(X, axis=0)
missing_idx = np.where(np.isnan(X))
X[missing_idx] = np.take(means, missing_idx[1])

x_train = X[(mask_indices == False).sum(axis=1) == 2, :]
pca = PCA(1).fit(x_train)
X_imp = pca.inverse_transform(pca.transform(X))

X[missing_idx] = X_imp[missing_idx]
plt.scatter(X[:, 0], X[:, 1], c=mask_indices.any(axis=1))
```

image

Using SVD to replace only missing values:
image

Using SVD to replace all values in rows with at least one missing:
image


I did another code check (using `TruncatedSVD` from `sklearn` instead of 'my' SVD calculation), but that didn't change much.

koaning:
Relevant observation. Are we still chasing this work? One reason for mentioning; it seems supported elsewhere too -> https://github.com/iskandr/fancyimpute.

pim-hoeven:
I would be willing to finish it, if you want me to. If I remember correctly I wasn't sure whether the results (as plotted above) were good enough to proceed. Other than that I think it was nearly finished. 

Not sure whether I have much time in the coming weeks though, but should be able to squeeze some sklego work in

pim-hoeven:
Just did a quick review and I don't immediately see things that I would like to change. 

I'll remove wip for now, you can decide whether to review or close.

koaning:
I'll close it for now but I won't remove the branch. I'm trying to make the maintenance of this package a bit more lightweight. There were a lot of lingering PRs piling up so I'm in cleanup mode now. Since the feature is also hosted elsewhere I don't feel the 'must have' need so much anymore.

pim-hoeven:
I completely understand!

Issue #407, opened 2020-08-13
[FEATURE] Log shape delta 


pim-hoeven:
Please explain clearly what you'd like to see added. 

I think this is not possible with the `extra` of `log_step_custom` in #397 , but I think it would be a nice feature. 

My current dataframe has shape `(5595919, 9)`, if it loses 100 rows it would be `(5595819, 9)`. I would argue that this is quite easy to miss if we have multiple logs per second (even with colors). Having the delta would result in (-100, 0) for this function, which would stand out if most other logs look like `(0, ...)`. For example, let `read_data` report the actual shape, all intermediate steps report the delta, and the final step report the delta and the final shape, you would get something like:
```
[08/10/20 07:28:32] DEBUG    [read_some_data    ] shape=(5595919, 9),        decorators.py:21
                             time=0:00:30.424626                                             

[08/10/20 07:28:32] DEBUG    [parse_dates       ] shape_delta=(-100, 0),     decorators.py:21
                             time=0:00:00.575011                                             

[08/10/20 07:28:32] DEBUG    [add_features      ] shape_delta=(0, +5),       decorators.py:21
                             time=0:00:01.8340287                                 
            
[08/10/20 07:28:32] DEBUG    [result            ] shape=(5595819, 14),       decorators.py:21
                             time=0:00:30.424626                
```

Another way to overcome this problem is to have a `shape_assertion` decorator on all intermediate steps.

- [x] convince us of the use-case, we're open to many suggestions but we prefer to solve problems with pipelines that are at least somewhat general 
- [x] add a screenshot if applicable (ML stuff is hard to explain with words, pictures say 1000 words)
- [x] make sure that the feature you want is not already supported by sklearn

koaning:
You can do this with the custom logger in that we discuss [here](https://github.com/koaning/scikit-lego/pull/397#issuecomment-670935881). You need to make a stateful function though.

koaning:
Now that I think about it ... I'm actually in favor of perhaps adding `shape_delta` to our standard loggers.

pim-hoeven:
Will add this while working on #397

koaning:
Merged.

Issue #371, opened 2020-06-23
[FEATURE] pandas log level print


koaning:
This sort of thing can happen currently with out log_level; 

![image](https://user-images.githubusercontent.com/1019791/85380387-2bc22e00-b53d-11ea-86f7-5ebb9208e2c4.png)

If I re-run the matplotlib cell it does go away, but this behavior is a bit annoying. 

Would we have anything against a `print_step` addition to the library?

MBrouns:
this is if you set your log level to debug? I'm not necessarily against print step but maybe this is a documentation issue instead.

```
logging.getLogger('matplotlib').setLevel(logging.INFO)
```

 also solves it

koaning:
This is true, but you'd need to do this for every module that you potentially import. But I agree that two valid PRs can now happen; 

- open to adding `print_step` 
- open to an update to the docs that also explains this phenomenon

koaning:
This got addressed.

Issue #99, opened 2019-04-03
[FEATURE] add log_dtypes to pandas_utils


koaning:
I might also want to log the `dtypes` between each step in a pandas pipeline. It's probably best to add a separate logger for both the `dtype` as well as the shape. 

Should the column names be seperate too? Maybe worth a discussion. 

It would also be nice if these features were documented on the documentation page together with some other `pandas_utils` functions.

koaning:
After thinking about it, we probably want to have these three:

- log_shape
- log_names
- log_dtypes 

Note too sure about `log_names` and `log_dtypes`.

david26694:
Is this still open? I'd be happy to take it

koaning:
It is still open, but let's first make sure we're clear on what will get implemented. Do you want to implement `log_dtypes` or also `log_names`? 

Relevant: there's a related issue too https://github.com/koaning/scikit-lego/issues/371.

david26694:
I was thinking about both. I don't fully understand what is happening in #371, why is matplotlib generating warnings and what would the print step do?

koaning:
The print step would do a very similar thing to what `log_step` would do but it would use `print` instead of using the python logging framework. The reason why matplotlib is generating warnings is because we've configured a logger to work in the notebook. This allows you to receive logs from `log_step` (this is awesome for production) but you might also get the logs from other libraries (which is not awesome in a notebook).

koaning:
This got adressed.

PR #411, opened 2020-09-17
Feature/improve logging

pim-hoeven:
Continuing on PR #397


pim-hoeven:
@koaning do you agree with 
```
"n_obs=!" 
"n_obs=+1"
"n_obs=-1"
```
for the shape_delta? Or would you prefer
- regular ints, (0, 1, -1)
- just something else for stayed the same
- omitting the n_obs if it didn't change

koaning:
I think my preference would be what you described your issue [here](https://github.com/koaning/scikit-lego/issues/407). 

Something like

```
delta = (0, 0)
delta = (-10, +2)
```

seems simple to me.

pim-hoeven:
I'm a bit stuck on the partial function, that's the only test that's still failing. Do you know what's going wrong? 

Next steps will be 

- [ ] implementing the extra logger function
- [ ] updating the docstrings

pim-hoeven:
@koaning ready for review :)

sklego/pandas_utils.py
```diff
@@ -11,45 +11,154 @@
 from scipy.ndimage.interpolation import shift
 
 
-def log_step(func=None, *, level=logging.INFO):
+def get_shape_delta(old_shape, new_shape):
```
koaning, line 14: Since this function isn't public, better translate it to `_get_shape_delta`.

sklego/pandas_utils.py
```diff
@@ -11,45 +11,154 @@
...
+    diffs = [
+        ("+" if new > old else "") + str(new - old)
+        for new, old in zip(new_shape, old_shape)
+    ]
+
+    return f"delta=({', '.join(diffs)})"
+
+
+def log_step(
```
koaning, line 23: Since you're addin the `log_*` updates. Might it also make sense to add the `print_*` variants? This might involve a small refactor but I'd love to also have something that is easy to hack around with in a jupyter notebook.

koaning: It relates to https://github.com/koaning/scikit-lego/issues/371

koaning (changes requested):
@pim-hoeven thanks! 

There's a small discussion to be had about `print_` here. It might make sense to add it here because it would require a refactor anyway. 

There's also no update to the documentation yet. This feature *definately* deserves it.

koaning:
Also for bookkeeping purposes. Here's a list of related issues. 

- https://github.com/koaning/scikit-lego/issues/99
- https://github.com/koaning/scikit-lego/issues/371
- https://github.com/koaning/scikit-lego/issues/407

sklego/pandas_utils.py
pim-hoeven: Yeah sure! Wouldn't adding a parameter to the call make more sense? That would make implementation easy and clean, and makes it less cumbersome (IMHO) to change if you move from notebooks to production.

koaning: That's a fair consideraction that I hadn't thought of. One concern that I have is that the `log_step` function, as is, already has **a lot** of arguments and in my mind you'd need to change the decorator anyway for production because the logging configuration probably requires customisation.

Just in case I'll ping @MBrouns on this one too. I haven't done a lot of "pandas in prod" lately so I might be missing a vantage point.

pim-hoeven: I've already added a logging parameter as well, which makes it easier to add your own handlers and formats (I'm using something similar at klm to have these logs in rich format)

Regarding the arguments: I agree. My current flow is to define a standard decorator per project (otherwise the logs will be very messy), and then it's probably less of a problem. I.e. you would define 
`project_log_step = log_step(...)` somewhere and import that log step from your own utils (or whatever) instead of the general one from sklego. But yeah let's see what Matthijs has to say :)

pim-hoeven: That's quite neat!

For my use cases, it would need a small adaption still, because I generally only change the log_level within my code, i.e. I always use the same log_step and the only parameter that I change is the log_level. 

Maybe we could just allow log_level to be `print` or some negative value (since the loglevels are all int)?

sklego/pandas_utils.py
pim-hoeven: Done

sklego/pandas_utils.py
pim-hoeven: Now that I think of it though, it doesn't matter match whether it doesn't matter much whether I do `print_fn=logger.info` than `logger=logging.INFO`.

koaning: I have a slight preference to assuming "print" as opposed to something-logging as a default. Mainly because my assumption is that most of the users of this tool will be folks working from a notebook. Notebooks appreciate prints more than loggers.

pim-hoeven: Perfect, so let's go with matthijs' idea: `print_fn=print`. Will implement it somewhere this weekend

pim-hoeven: Done!

sklego/pandas_utils.py
```diff
@@ -11,45 +9,139 @@
...
+def log_step_extra(
+    *log_functions, print_fn=print, **log_func_kwargs,
+):
+    """
+    Decorates a function that transforms a pandas dataframe to add automated logging statements
+
+    :param *log_functions: callable(s), functions that take the output of the decorated function and turn it into a log.
+                                        Note that the output of each log_function is casted to string using `str()`
+    :param print_fn: callable, print function (e.g. print or logger.info), defaults to print
+    :param **log_func_kwargs: keyword arguments to be passed to log_functions
```
koaning, line 105: Might be good to explicitly menion here that this is useful for setting log-levels, etc.

koaning (changes requested):
Thanks @pim-hoeven I like the new `delta` stuff! 

However, I think I may have found a bug while playing with `rich`. 

```python
from sklego.datasets import load_heroes
df = load_heroes(as_frame=True)

import time
from sklego.pandas_utils import log_step, log_step_extra
from rich.console import Console

console = Console()

def n_types(dataf):
    return f"n_types: {dataf['attack_type'].nunique()}"

@log_step_extra(n_types, print_fn=console.log)
@log_step(shape_delta=True, print_fn=console.log)
def identity(dataf):
    return dataf.copy()

@log_step_extra(n_types, print_fn=console.log, log_locals=True)
@log_step(shape_delta=True, print_fn=console.log)
def sample(dataf):
    time.sleep(1.0)
    return dataf.sample(n=10)

df.pipe(identity).pipe(sample).pipe(sample).pipe(sample)
```

Part of the output is pretty highting, exactly what I expected. But then ... 

image

Here's the traceback. 

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/Development/scikit-lego/sklego/pandas_utils.py in wrapper(*args, **kwargs)
    128                 extra_logs = " ".join(
--> 129                     [str(log_f(result, **log_func_kwargs)) for log_f in log_functions]
    130                 )

~/Development/scikit-lego/sklego/pandas_utils.py in <listcomp>(.0)
    128                 extra_logs = " ".join(
--> 129                     [str(log_f(result, **log_func_kwargs)) for log_f in log_functions]
    130                 )

TypeError: n_types() got an unexpected keyword argument 'log_locals'

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-10-115ef85ca8f1> in <module>
     19     return dataf.sample(n=10)
     20 
---> 21 df.pipe(identity).pipe(sample).pipe(sample).pipe(sample)

~/Development/scikit-lego/venv/lib/python3.7/site-packages/pandas/core/generic.py in pipe(self, func, *args, **kwargs)
   5043         ...  )  # doctest: +SKIP
   5044     """
-> 5045         return com.pipe(self, func, *args, **kwargs)
   5046 
   5047     _shared_docs["aggregate"] = dedent(

~/Development/scikit-lego/venv/lib/python3.7/site-packages/pandas/core/common.py in pipe(obj, func, *args, **kwargs)
    469         return func(*args, **kwargs)
    470     else:
--> 471         return func(obj, *args, **kwargs)
    472 
    473 

~/Development/scikit-lego/sklego/pandas_utils.py in wrapper(*args, **kwargs)
    131             except TypeError:
    132                 raise ValueError(
--> 133                     f"All log functions should be callable, got {[type(log_f) for log_f in log_functions]}"
    134                 )
    135 

ValueError: All log functions should be callable, got [<class 'function'>]
```

I think the `log_locals=True` argument is not being passed to `console.log` somehow. Might be good to add something like this as a unit test.

pim-hoeven:
This is actually intended behavior (at least on my part). The kwargs are passed to the extra log function, not to the the 'printing' function. I think the former we definitely want to have as an option, but I can add the second as well?

koaning:
Ahhhh right. That's a requirement. Guess I need to; 

1. Not review something immedaitely after working 8 hours. 
2. In practice you might need to "curry" the `console.log` then.

tests/test_pandas_utils/test_pandas_utils.py
```diff
@@ -73,8 +72,260 @@ def do_nothing(df, *args, **kwargs):
...
+            return df
+
+        test_df.pipe(do_nothing)
+
+        assert "log_function" in str(e)
+
+
+def test_log_extra_not_callable_func(test_df):
+    with pytest.raises(ValueError) as e:
+        @log_step_extra(1)
```
koaning, line 324: It might be nice to either add a docstring or a comment explaining what we're testing with these unit tests. Our memory is still fresh now but a year from now somebody might like to read "you must pass a callable to `log_step_extra`" in this test.

koaning (changes requested):
LGTM! 

My only comment is that it might be nice to add a few comments/docstrings to the tests. Just to make it friendlier to anybody reviewing this code a year from now. Other than that; thanks!

tests/test_pandas_utils/test_pandas_utils.py
pim-hoeven: Done!

koaning:
Cool. I might push this new version live tonight on pypi. Probably I'll market it a few weeks from now once my honeymoon is done. 

@pim-hoeven @david26694 thanks for your efforts!

pim-hoeven:
Congratulations!! And enjoy your honeymoon!!

PR #435, opened 2021-02-05
Modernize Dependencies

koaning:
I noticed [some strange dependency issues](https://github.com/koaning/scikit-lego/runs/1835218488?check_suite_focus=true) when I realized that we haven't updated our dependencies in setup.py in a *long while*. 

This PR will "modernize" our dependencies. Hopefully, that also removes some bugs in the future.


koaning:
Mhm. This is rather strange. 

@MBrouns just to check, any idea what's going wrong here? Locally it all installs fine. I can't figure out where the hard-pinned numpy dependency comes from.

MBrouns:
Upstream issues seems to be fixed with cvxpy 1.1.9 https://github.com/cvxgrp/cvxpy/issues/1236

PR #437, opened 2021-02-09
A few edits for grammar and clarity

RoyalTS:
Thought I might as well make myself useful while benefitting from your material :)


MBrouns (approved):

koaning:
Out of curiosity, what parts specifically do you find useful @RoyalTS ?

RoyalTS:
I'm pretty new to fairness. I had always thought that people who are just throwing "tricky variables" out of the model to deal with the issue are deluding myself so even just reading about the other approaches you're dealing with was interesting to me. Also: I enjoy your style, both in written form and in your videos, and feel like I've learned a ton of stuff from you. So: Ta!

PR #439, opened 2021-02-18
Added the regressors

Garve:
Hi!

I added the regressors. it's my first pull request, so I don't know if I did everything correctly.

Best regards
Robert


koaning:
Could you add tests for the regressors? There's a few examples in the `tests` section of the package that should make it clear how we typically test components. The most important part is to run the standard tests from scikit-learn on your tools.

Garve:
Heya!

So far I used some of my own tests and added your check_shape_remains_same_regressor test. Is this ok?

Best Robert

koaning:
One main thing to add is [this](https://github.com/koaning/scikit-lego/blob/master/tests/test_estimators/test_gmm_naive_bayes.py#L9-L27).  This is how we run against scikit-learns own testing framework which guarantees compatibility. It's not a hard requirement that we pass *all* the checks, but we like to understand which ones do not pass before we hit merge.

Garve:
All checks passed!

koaning:
Nice! In that case, we've entered the phase where @MBrouns and myself will do an in-depth review. 

In the meantime, have you thought about documentation?

Garve:
Perfect!

I was looking around, but couldn't find any public 'real life' dataset where LADRegression is useful. Scikit-learn's documentation also only offers artificial datasets. Maybe we can go for this one too?

The same goes for the ImbalancedRegression. At work, I stumbled upon the problem that "underfitting is worse than overfitting". Can't share the dataset, though :D The only way I see at the moment is playing with artificial datasets here, too.

koaning:
A very short benchmark then? Simulated data is also fine, as long as there's a plausible practical motivation around.

Garve:
For the LADRegression, how about something like this:

# Least Absolute Deviation Regression

Imagine that you have a dataset with some outliers.
```python
import numpy as np
import matplotlib.pyplot as plt

np.random.seed(0)
X = np.linspace(0, 1, 20)
y = 3*X + 1 + 0.5*np.random.randn(20)
X = X.reshape(-1, 1)

y[10] = 8
y[15] = 15

plt.figure(figsize=(16, 4))
plt.scatter(X, y)
```

grafik



A simple linear regression will not do a good job since it is distracted by the outliers. That is because it optimizes the mean squared error which penalizes a few large errors more than many tiny errors. For example, if y-model(X) = 4 for some single observation, the MSE here is 16. If there are two observations with y_1 - model(X_1) = 2 and y_2 - model(X_2) = 2, the MSE is 8 in total, which is less than for one larger error. Note that the sum of the errors is the same in both cases.

Hence, linear regression does the following:

grafik

By changing the loss function to the mean absolute deviation |y - model(X)|, we can let the model put the same focus on each error. This yields the least absolute deviation (LAD) regression that tries to agree with the majority of the points.

grafik

koaning:
That works swell for an explainer. The only thing I'm wondering is if it makes sense to mention similar tools in scikit-learn. It occurs to me as natural that a person reads this and thinks "didn't scikit-learn also have this?".

Garve:
Added a small passage to the description above.

tests/test_estimators/test_lad_regression.py
```diff
@@ -0,0 +1,106 @@
...
+@pytest.mark.parametrize("test_fn", [check_shape_remains_same_regressor])
+def test_lad(test_fn):
+    regr = LADRegression()
+    test_fn(LADRegression.__name__, regr)
+
+
+@pytest.mark.parametrize(
+    "test_fn",
+    select_tests(
+        flatten([general_checks, nonmeta_checks]),
```
koaning, line 95: There are also `regressor_checks`. These should be added here.

tests/test_estimators/test_imbalanced_linear_regression.py
```diff
@@ -0,0 +1,129 @@
...
+@pytest.mark.parametrize("test_fn", [check_shape_remains_same_regressor])
+def test_imbalanced(test_fn):
+    regr = ImbalancedLinearRegression()
+    test_fn(ImbalancedLinearRegression.__name__, regr)
+
+
+@pytest.mark.parametrize(
+    "test_fn",
+    select_tests(
+        flatten([general_checks, nonmeta_checks]),
```
koaning, line 118: There are also `regressor_checks`. These should be added here.

tests/test_estimators/test_imbalanced_linear_regression.py
```diff
@@ -0,0 +1,129 @@
...
+    imb.fit(X, y)
+
+    assert imb.intercept_ == 0.0
+
+
+def test_check_estimator():
+    """Conduct all scikit-learn estimator tests."""
+    imb = ImbalancedLinearRegression()
+
+    check_estimator(imb)
```
koaning, line 106: In our library we use the `select_tests` functionality to select specific checks from the `check_estimator`. What you're doing here is good, but I think we're doing some tests twice now. I personally have a preference to use the following pattern everywhere;

```python
@pytest.mark.parametrize(
    "test_fn",
    select_tests(
        flatten([general_checks, nonmeta_checks]),
   )
)
```

This allows us to select a subset of the relevant tests, which is especially useful in our constrained estimators. @MBrouns do you have any strong opinions on this?

koaning (changes requested):
Nice! 

There's one discussion that should be had on the `check_estimators` but other than that this looks like a really clean implementation. 

@Garve could you add a section to the docs? You should be able to add a section to the `linear_models` notebook.

tests/test_estimators/test_imbalanced_linear_regression.py
MBrouns: I would prefer to be consistent with the rest of the library. I also prefer our way slightly as it makes it more obvious which exact check fails

tests/test_estimators/test_imbalanced_linear_regression.py
MBrouns: I would also prefer to parametrize over the different instances of the `ImbalancedLinearRegression`. E.g.:

```
@pytest.mark.parametrize(
    "regr", [
         (ImbalancedLinearRegression.__name__, ImbalancedLinearRegression()), 
         (ImbalancedLinearRegression.__name__ + "_positive", ImbalancedLinearRegression(positive=True)),
         ... 
     ]
)
@pytest.mark.parametrize(
    "test_fn",
    select_tests(
        flatten([general_checks, nonmeta_checks]),
 )
)
def test_estimator_checks(regr, test_fn):
    test_fn(*regr)
```

tests/test_estimators/test_lad_regression.py
MBrouns: Same here as above

Garve: Heya! Done.

I also added a section.

tests/test_estimators/test_imbalanced_linear_regression.py
Garve: Done!

tests/test_estimators/test_imbalanced_linear_regression.py
Garve: Done!

koaning:
LGTM! 

Looks like I'll be preparing a release this week. Thanks for the work!

koaning (approved):

koaning:
@Garve it's been added. One thing: do you have a twitter account? We usually post a shoutout when we announce new features.

Garve:
Heya!

Happy to hear that :) 

And I don't have a Twitter account, but thank you anyway!

PR #442, opened 2021-03-02
added ZeroInflatedRegressor

Garve:
How about this?


tests/test_meta/test_zero_inflated_regressor.py
```diff
@@ -0,0 +1,24 @@
...
+from sklego.meta import ZeroInflatedRegressor
+from sklego.testing import check_shape_remains_same_regressor
+from tests.conftest import general_checks, select_tests, regressor_checks
+
+
+@pytest.mark.parametrize("test_fn", [check_shape_remains_same_regressor])
+def test_lad(test_fn):
+    regr = ZeroInflatedRegressor()
+    test_fn(ZeroInflatedRegressor.__name__, regr)
+
```
koaning, line 15: I might add a test with an "obvious example" to demonstrate that it indeed is able to separate the zeros from a normal regression pattern.

sklego/meta/zero_inflated_regressor.py
```diff
@@ -0,0 +1,109 @@
...
+        - The regressor's task is to output a (usually positive) prediction whenever the classifier indicates that the there should be a non-zero prediction.
+
+    The regressor is only trained on examples where the target is non-zero, which makes it easier for it to focus.
+
+    At prediction time, the classifier is first asked if the output should be zero. If yes, output zero.
+    Otherwise, ask the regressor for its prediction and output it.
+
+    Parameters
+    ----------
+    classifier : Any, scikit-learn classifier, default=`DummyClassifier(strategy="constant", constant=1)`
```
koaning, line 26: Do we really add a default here?

sklego/meta/zero_inflated_regressor.py
```diff
@@ -0,0 +1,109 @@
...
+        A classifier that answers the question "Should the output be zero?".
+
+    regressor : Any, scikit-learn regressor, default=`LinearRegression()`
```
koaning, line 29: In general, I might prefer `Ridge` over `LinearRegression` but I'm wondering if it's better to not use a default parameter here. It feels like the responsibility of the end-users to come up with something meaningful here that fits their task. Any opportunity to nudge a user to actually think about their problem should be cherished.  😄

sklego/meta/zero_inflated_regressor.py
```diff
@@ -0,0 +1,109 @@
...
+        ZeroInflatedRegressor
+            Fitted regressor.
+        """
+        X, y = check_X_y(X, y)
+        self._check_n_features(X, reset=True)
+
+        self.classifier_ = (
+            clone(self.classifier)
+            if self.classifier is not None
+            else DummyClassifier(strategy="constant", constant=1)
```
koaning, line 74: Wouldn't it be better to force the user to give us a proper classifier here?

koaning (changes requested):
The code looks pretty good, but it'd be good to discuss the default parameters. It might make sense to force the user to supply something meaningful instead of making assumptions.

sklego/meta/zero_inflated_regressor.py
Garve: I'm fine with it!

sklego/meta/zero_inflated_regressor.py
Garve: Let's remove it then together with the classifier.

sklego/meta/zero_inflated_regressor.py
Garve: Classifier begone!

Garve:
Ah I know why I added default estimators. If I do this:

```python
check_estimator(
    ZeroInflatedRegressor(
        classifier=RandomForestClassifier(random_state=0),
        regressor=RandomForestRegressor(random_state=0)
    )
)
```

then I get 

```bash
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-71-f1b5b02cf256> in <module>
----> 1 check_estimator(ZeroInflatedRegressor(classifier=RandomForestClassifier(random_state=0), regressor=RandomForestRegressor(random_state=0)))

~\AppData\Roaming\Python\Python37\site-packages\sklearn\utils\estimator_checks.py in check_estimator(Estimator, generate_only)
    545     for estimator, check in checks_generator():
    546         try:
--> 547             check(estimator)
    548         except SkipTest as exception:
    549             # SkipTest is thrown when pandas can't be imported, or by checks

~\AppData\Roaming\Python\Python37\site-packages\sklearn\utils\estimator_checks.py in check_parameters_default_constructible(name, Estimator)
   2639 
   2640     with ignore_warnings(category=FutureWarning):
-> 2641         estimator = _construct_instance(Estimator)
   2642         # test cloning
   2643         clone(estimator)

~\AppData\Roaming\Python\Python37\site-packages\sklearn\utils\estimator_checks.py in _construct_instance(Estimator)
    371             raise SkipTest(msg)
    372     else:
--> 373         estimator = Estimator()
    374     return estimator
    375 

TypeError: __init__() missing 2 required positional arguments: 'classifier' and 'regressor'
```

Do you know how I can fix that without defining default estimators? Or is this test just not suited for meta estimators?

tests/test_meta/test_zero_inflated_regressor.py
Garve: I used the one from the feature requests, is this ok?

sklego/meta/zero_inflated_regressor.py
```diff
@@ -0,0 +1,108 @@
...
+            The target values.
+
+        Returns
+        -------
+        ZeroInflatedRegressor
+            Fitted regressor.
+        """
+        X, y = check_X_y(X, y)
+        self._check_n_features(X, reset=True)
+
```
koaning, line 70: Should we perhaps check if the `regressor` is actually a regressor? Same for the `classifier? We should be able to throw a proper `ValueError` if the appropriate mixin is missing.

koaning: It's a relatively minor thing but that's the only thing I might add to this PR. Maybe add a unit test that checks that we're indeed throwing the right errors.

sklego/meta/zero_inflated_regressor.py
```diff
@@ -0,0 +1,108 @@
...
+        self.classifier_ = clone(self.classifier)
+        self.classifier_.fit(X, y != 0)
+
+        non_zero_indices = np.where(self.classifier_.predict(X) == 1)[0]
+
+        if non_zero_indices.size > 0:
+            self.regressor_ = clone(self.regressor)
+            self.regressor_.fit(X[non_zero_indices], y[non_zero_indices])
+        else:
+            self.regressor_ = None
```
MBrouns, line 80: I think this should either throw a warning or an exception.
I'm leaning towards an exception since I don't think having an unfitted regressor would ever be the behaviour you wanted out of this model, but I'm happy to be convinced that it could be useful and a warning suffices

MBrouns (changes requested):

sklego/meta/zero_inflated_regressor.py
MBrouns: I think checking whether it's a regressor or classifier is indeed a nice addition. There's `is_classifier` and `is_regressor` in `sklearn.base`

Garve: Thanks for pointing this out!

sklego/meta/zero_inflated_regressor.py
Garve: How about this:

```python
        if non_zero_indices.size > 0:
            self.regressor_ = clone(self.regressor)
            self.regressor_.fit(X[non_zero_indices], y[non_zero_indices])
        else:
            self.regressor_ = DummyRegressor(strategy="constant", constant=0)
            self.regressor_.fit(X, y)
```

including some warning that the regressor got overridden because there were no training samples for it? The regressor won't get called anyway for sensible classifiers since the model will output zero all the time.

Garve: I changed it in the PR.

sklego/meta/zero_inflated_regressor.py
```diff
@@ -0,0 +1,122 @@
...
+        self.classifier_ = clone(self.classifier)
+        self.classifier_.fit(X, y != 0)
+
+        non_zero_indices = np.where(self.classifier_.predict(X) == 1)[0]
+
+        if non_zero_indices.size > 0:
+            self.regressor_ = clone(self.regressor)
+            self.regressor_.fit(X[non_zero_indices], y[non_zero_indices])
+        else:
+            warnings.warn("The training labels are all zero. Hence, the passed regressor cannot be trained and is replaced by a model that constantly outputs zero.")
```
MBrouns, line 92:
```suggestion
            warnings.warn("The predicted training labels are all zero. Hence, the passed regressor cannot be trained and is replaced by a model that constantly outputs zero.")
```

sklego/meta/zero_inflated_regressor.py
koaning: I would prefer an exception too. Warnings are too often ignored, especially when run from the command line and we really want to make the user aware if we never predict anything but a zero.

sklego/meta/zero_inflated_regressor.py
```diff
@@ -0,0 +1,122 @@
...
+            If `classifier` is not a classifier or `regressor` is not a regressor.
+        """
+        X, y = check_X_y(X, y)
+        self._check_n_features(X, reset=True)
+        if not is_classifier(self.classifier):
+            raise ValueError(f"`classifier` has to be a classifier. Received instance of {type(self.classifier)} instead.")
+        if not is_regressor(self.regressor):
+            raise ValueError(f"`regressor` has to be a regressor. Received instance of {type(self.regressor)} instead.")
+
+        self.classifier_ = clone(self.classifier)
```
koaning, line 83: I wonder. What if the user wants to pass a `classifier` that has already been fitted before? Do we really want to refit in these scenarios?

Garve: True, we could check if the estimators were fitted before. If yes, use them, otherwise proceed as before. But then we could also think about some `refit` keyword. Maybe someone wants to pass in a fitted estimator but wants it to be refit.

sklego/meta/zero_inflated_regressor.py
Garve: Done.

sklego/meta/zero_inflated_regressor.py
Garve: I added it without the new parameter, maybe it's better not to introduce another parameter.

koaning: I'm fine with a `refit` parameter. It's also something we have in [other meta models](https://scikit-lego.netlify.app/api/meta.html#sklego.meta.Thresholder).

Garve: I just struggle a bit with how to implement it. Either as a single parameter for both models, which costs some flexibility, or as two separate parameters such as `refit_classifier` and `refit_regressor`, but this introduces so much parameter overhead. What looks better to you?

koaning: Ah yeah I see what you mean there. 

I'm fine with forcing both models to be trained for now. We can always add the feature if users really start asking for it. It's not a dealbreaker for me.

Garve: ...to be trained during ZIR fit time, right? :D

MBrouns:
I think this is good to go, right? I didn't render the docs to check whether everything is okay there as well but I'll do that tomorrow. if @koaning agrees I'll merge it afterwards

MBrouns (approved):

koaning (approved):

koaning:
This feature is now live! https://scikit-lego.netlify.app/meta.html#Zero-Inflated-Regressor

PR #449, opened 2021-03-21
implements predict_proba for outlier_classifier

fritshermans:
Hi guys, I added a `predict_proba` method to the `OutlierClassifier` class such that sklearn's StackingClassifier also works with 'predict_proba' and not only with hard predictions. Often, whether a datapoint is an outlier or not is not so black or white. I used a sigmoid function to be able to cover for all kinds of outlier models, independent of the range of their `decision_function` output. It can be used like this:

```
estimators = [
    ('anomaly', OutlierClassifier(IsolationForest())),
    ('classifier', RandomForestClassifier())
    ]

stacker = StackingClassifier(estimators, stack_method='predict_proba', cv=4, passthrough=True)
```


koaning:
1. I like this idea, but I want to give it a proper review which is something I'll probably only get to later this week.
2. For future commits, it's usually a better idea to first raise an issue. In this case, I can assure you that I like this idea, but sometimes we need to double-check if a direction makes sense. To prevent wasting time on a PR, it's preferable to start an issue before working on an implementation.

sklego/meta/outlier_classifier.py
```diff
@@ -31,6 +32,11 @@ def fit(self, X, y=None):
             raise ValueError("Passed model does not detect outliers!")
         self.estimator_ = self.model.fit(X, y)
         self.classes_ = np.array([0, 1])
+
+        # fit sigmoid function for `predict_proba`
+        decision_function_scores = self.estimator_.decision_function(X)
```
koaning, line 37: I just checked, *technically* there [is an outlier detection mechanism in scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.LocalOutlierFactor.html) that doesn't support `decision_function`. That it's a bit of an oddball since it only supports `fit_predict`. Theoretically we might want to check/confirm that the decision function exists, but practically it's not something we should expect.

tests/test_meta/test_outlier_classifier.py
```diff
@@ -38,6 +38,8 @@ def test_obvious_usecase_quantile(dataset):
     clf_quantile.fit(X, y)
     assert clf_quantile.predict([[10, 10]]) == np.array([1])
     assert clf_quantile.predict([[0, 0]]) == np.array([0])
+    np.testing.assert_array_almost_equal(clf_quantile.predict_proba([[10, 10]]), np.array([[0, 1]]), decimal=4)
+    np.testing.assert_array_almost_equal(clf_quantile.predict_proba([[0, 0]]), np.array([[1, 0]]), decimal=4)
     assert isinstance(clf_quantile.score(X, y), float)
```
koaning, line 43: As far as tests go I'd argue this is fine. I only wonder if it's perhaps good to also add a parameterized for the outlier detection method. We're only testing `GMMOutlierDetector` here. I wouldn't mind adding [isolation forests](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.IsolationForest.html#sklearn.ensemble.IsolationForest) and [one class svm](https://scikit-learn.org/stable/modules/generated/sklearn.svm.OneClassSVM.html#sklearn.svm.OneClassSVM).

koaning (commented):
I'm in favor of this, nice PR @fritshermans! There are a few minor details to discuss first but all in all: nice work 👍!

sklego/meta/outlier_classifier.py
fritshermans: if `novelty=True` it also works with the LocalOutlierFactor. I can add an assertion for that and raise an error if this setting is not chosen.

tests/test_meta/test_outlier_classifier.py
fritshermans: sure, i'll do that!

sklego/meta/outlier_classifier.py
koaning: Checking for a setting on a specific outlier detection model might be too "specific". Now that I think about it. I suppose wrapping it with a `try`/`except` such that a meaningful error message is given is sufficient.

fritshermans: we could also check if a `decision_function` is available, then it's not specific for the LocalOutlierFactor but also not so wide as using a try/except. I'll commit it, let me know what you think about it.

sklego/meta/outlier_classifier.py
```diff
@@ -29,8 +30,16 @@ def fit(self, X, y=None):
         X, y = check_X_y(X, y, estimator=self)
         if not self._is_outlier_model():
             raise ValueError("Passed model does not detect outliers!")
+        if not hasattr(self.model, 'decision_function'):
+            raise ValueError('Passed model does not have a `decision_function` method. This is required for '
```
koaning, line 34: This is a nitpick, but it would be better to also list the model in question. Maybe instead raise something like: 

```python
raise ValueError(f'Passed model {self.model} does not have a `decision_function` '
                 f'method. This is required for `predict_proba` estimation.')
```

tests/test_meta/test_outlier_classifier.py
```diff
@@ -38,13 +41,52 @@ def test_obvious_usecase_quantile(dataset):
...
+    X = dataset
+    y = (dataset.max(axis=1) > 3).astype(np.int)
+    outlier_model = OneClassSVM(nu=0.05)
+    clf_model = OutlierClassifier(outlier_model)
+    clf_model.fit(X, y)
+    assert clf_model.predict([[10, 10]]) == np.array([1])
+    assert clf_model.predict([[0, 0]]) == np.array([0])
+    np.testing.assert_array_almost_equal(clf_model.predict_proba([[10, 10]]), np.array([[0, 1]]), decimal=3)
+    np.testing.assert_array_almost_equal(clf_model.predict_proba([[0, 0]]), np.array([[1, 0]]), decimal=3)
+    assert isinstance(clf_model.score(X, y), float)
```
koaning, line 74: Could these two really not be combined via a `pytest.mark.parametrize` into a single test?

koaning (changes requested):
Good work! We're nearly green, but there's two nitpicks to discuss.

tests/test_meta/test_outlier_classifier.py
fritshermans: I'm not familiar with `pytest.mark.parametrize` so I will look into it and implement it as you suggest

koaning: Oh! It's amazing! 

You might appreciate this small guide: https://calmcode.io/pytest/clean-params.html

fritshermans: I changed it such that pytest parametrization is used. Should I move `check_predict_proba` to `estimator_checks.py` and import it via `conftest.py`?

koaning:
I fear that I should have been more clear. 

### Before 

This was the pytest code during my initial review. 

```python
def test_obvious_usecase_isolationforest(dataset):
    X = dataset
    y = (dataset.max(axis=1) > 3).astype(np.int)
    outlier_model = IsolationForest(contamination=y.sum()/len(y))
    clf_model = OutlierClassifier(outlier_model)
    clf_model.fit(X, y)
    assert clf_model.predict([[10, 10]]) == np.array([1])
    assert clf_model.predict([[0, 0]]) == np.array([0])
    np.testing.assert_array_almost_equal(clf_model.predict_proba([[0, 0]]), np.array([[1, 0]]), decimal=4)
    outlier_proba = clf_model.predict_proba([[10, 10]])[0]
    assert outlier_proba[0]<0.2
    assert outlier_proba[1]>0.8
    assert isinstance(clf_model.score(X, y), float)


def test_obvious_usecase_oneclass_svm(dataset):
    X = dataset
    y = (dataset.max(axis=1) > 3).astype(np.int)
    outlier_model = OneClassSVM(nu=0.05)
    clf_model = OutlierClassifier(outlier_model)
    clf_model.fit(X, y)
    assert clf_model.predict([[10, 10]]) == np.array([1])
    assert clf_model.predict([[0, 0]]) == np.array([0])
    np.testing.assert_array_almost_equal(clf_model.predict_proba([[10, 10]]), np.array([[0, 1]]), decimal=3)
    np.testing.assert_array_almost_equal(clf_model.predict_proba([[0, 0]]), np.array([[1, 0]]), decimal=3)
    assert isinstance(clf_model.score(X, y), float)
```

My concern was the repetition in both of the functions. The main difference between these two tests is the `outlier_model` so that is something that could be parametrized. 

## My Goal 

When reviewing I suggested to use `pytest.mark.parametrize` with the intent of achieving something like this: 

```python
@pytest.mark.parametrize('mod_quantile', [GMMOutlierDetector(), OneClassSVM(nu=0.05),])
def test_obvious_usecase(dataset, mod_quantile):
    clf_quantile = OutlierClassifier(mod_quantile)
    X = dataset
    y = (dataset.max(axis=1) > 3).astype(np.int)
    clf_quantile.fit(X, y)
    assert clf_model.predict([[10, 10]]) == np.array([1])
    assert clf_model.predict([[0, 0]]) == np.array([0])
    np.testing.assert_array_almost_equal(clf_model.predict_proba([[0, 0]]), np.array([[1, 0]]), decimal=4)
    outlier_proba = clf_model.predict_proba([[10, 10]])[0]
    assert outlier_proba[0]<0.2
    assert outlier_proba[1]>0.8
    assert isinstance(clf_model.score(X, y), float)
```

In this case, we're passing `mod_quantile` into the test. Pytest will turn this decorated function into two separate tests. That means that both instances will appear in any testing logs. This pattern allows us to apply one specific check over many models. 

@fritshermans you mentioned you were unfamiliar with `parametrize` and in hindsight, I should have picked up on that to be more explicit. Apologies! If I had recognized that it would've saved us time.

tests/test_meta/test_outlier_classifier.py
fritshermans: thanks for your remarks below, this is indeed the way I should have used the `parametrize` function. I have adjusted in in the latest commit. Thanks for your help! 🙂

koaning:
Perfect! This looks great! 

@fritshermans thanks for the PR! This is a good addition. Feel free to drop a line if you notice something else 👍

koaning (approved):
😄

fritshermans:
@koaning , shall I create a pull request with documentation about the `OutlierClassifier`?

PR #451, opened 2021-03-24
Update pandas_utils.py

daimonie:
One of my pipeline steps (that I wanted to log) actually performs a merge operation, which led to some extensive debug output because it basically printed the head() of the "right" dataframe. It would be helpful to be able to disable that behavior.

### DIFF
--- a/sklego/pandas_utils.py
+++ b/sklego/pandas_utils.py
@@ -27,7 +27,7 @@ def log_step(
     names=False,
     dtypes=False,
     print_fn=print,
-    print_args=True
+    display_args=True
 ):
     """
     Decorates a function that transforms a pandas dataframe to add automated logging statements
@@ -62,7 +62,7 @@ def log_step(
             names=names,
             dtypes=dtypes,
             print_fn=print_fn,
-            print_args=print_args
+            display_args=display_args
         )
 
     names = False if dtypes else names
@@ -85,7 +85,7 @@ def log_step(
 
         combined = " ".join([s for s in optional_strings if s])
 
-        if print_args:
+        if display_args:
 
             func_args = inspect.signature(func).bind(*args, **kwargs).arguments
             func_args_str = "".join(
--- a/sklego/pandas_utils.py
+++ b/sklego/pandas_utils.py
@@ -27,6 +27,7 @@ def log_step(
     names=False,
     dtypes=False,
     print_fn=print,
+    print_args=True
 ):
     """
     Decorates a function that transforms a pandas dataframe to add automated logging statements
@@ -38,6 +39,7 @@ def log_step(
     :param names: bool, log the names of the columns of the result, defaults to False
     :param dtypes: bool, log the dtypes of the results, defaults to False
     :param print_fn: callable, print function (e.g. print or logger.info), defaults to print
+    :param print_args: bool, whether or not to print the arguments given to the function.
     :returns: the result of the function
 
     :Example:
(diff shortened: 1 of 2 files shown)


koaning:
I see the merit in this change but my first response is to consider referring to the parameter as `show_args` as opposed to `print_args`. @MBrouns just to check, got any comments on this?

MBrouns:
I see how that can be painful, and I think allowing users to select whether to print the args or not makes sense. I do have some comments which I'll put inline in the PR

MBrouns:
Oh right, `show_args` or `display_args` makes more sense than `print_args` as the latter implies that they're args that get passed to the `print` function

sklego/pandas_utils.py
```diff
@@ -86,8 +89,10 @@ def wrapper(*args, **kwargs):
         func_args_str = "".join(
             ", {} = {!r}".format(*item) for item in list(func_args.items())[1:]
         )
-        print_fn(f"[{func.__name__}(df{func_args_str})] " + combined,)
-
+        if print_args:
+            print_fn(f"[{func.__name__}(df{func_args_str})] " + combined,)
+        else:
+            print_fn(f"[{func.__name__}()")
```
MBrouns, line 95: The formatting here is a bit off. 
```suggestion
            print_fn(f"[{func.__name__}]" + combined,)
```

sklego/pandas_utils.py
```diff
@@ -86,8 +89,10 @@ def wrapper(*args, **kwargs):
         func_args_str = "".join(
```
MBrouns, line 89: `inspect.signature` only needs to happen if we print the args, so it's probably better to move it into the `if` statement below

MBrouns (changes requested):

daimonie:
Hey guys, 

That was quick :) I made the suggested changes. I agree,   `display_args`  is the better name

MBrouns:
👍 

One (I think) last thing, could you add a test case to the `tests/pandas_utils.py`? There are already some examples on how you can test what output gets printed

tests/test_pandas_utils/test_pandas_utils.py
```diff
@@ -81,6 +81,26 @@ def do_nothing(df, *args, **kwargs):
     assert print_statements[1].startswith("[do_nothing(df, kwargs = {'a': '1'})]")
     assert print_statements[2].startswith("[do_something(df)]")
 
+def test_log_step_display_args(capsys, test_df):
+    """Base test of log_step without any arguments to the logger"""
```
MBrouns, line 85:
```suggestion
    """Test that we can disable printing function arguments in the log_step""
```

MBrouns (approved):

tests/test_pandas_utils/test_pandas_utils.py
```diff
@@ -81,6 +81,26 @@ def do_nothing(df, *args, **kwargs):
     assert print_statements[1].startswith("[do_nothing(df, kwargs = {'a': '1'})]")
     assert print_statements[2].startswith("[do_something(df)]")
 
+def test_log_step_display_args(capsys, test_df):
+    """Test that we can disable printing function arguments in the log_step""
```
MBrouns, line 85:
```suggestion
    """Test that we can disable printing function arguments in the log_step"""
```

MBrouns:
Thanks for the pr @daimonie!

daimonie:
Haha, thanks for the library @MBrouns  @koaning

PR #455, opened 2021-04-05
adds `OutlierClassifier` explanation to docs

fritshermans:
as discussed in #449


koaning:
Thanks, @fritshermans! Looks great! 

My only comment is, would you agree that it's fair to mention that this technique is a reasonable thing to try in fraud use cases? I'm wondering if there's something about possible applications that we might also be able to mention here.

koaning (approved):

koaning:
LGTM. Let me deploy a new version of the docs right away.

PR #465, opened 2021-04-29
QuantileRegression

Garve:
Alright, second try.


tests/test_estimators/test_lad_regression.py
```diff
@@ -59,7 +59,7 @@ def test_coefs_and_intercept__no_noise_regularization(coefs, intercept):
     """Test model with regularization. The size of the coef vector should shrink the larger alpha gets."""
     X, y = _create_dataset(coefs, intercept)
 
-    lads = [LADRegression(alpha=alpha, l1_ratio=0.5).fit(X, y) for alpha in range(3)]
+    lads = [LADRegression(alpha=alpha, l1_ratio=0.).fit(X, y) for alpha in range(3)]
```
koaning, line 62: Is there a reason why the LADRegression test needed to be changed?

koaning (changes requested):
Only did a quick review now, left a comment. Hope to dive in more this weekend.

tests/test_estimators/test_lad_regression.py
Garve: I just noticed by accident. Dunno why I put a 0.5 there in the first place.

I just wanted to impose an L2 regularization, and that's what I check I also check afterward. in the next lines, I check if the sum of squares of coefficients goes down the larger alpha gets, and not a mix of the L1 and L2 norms of the coefficient.

The test still worked for the LADRegression, but it was just a happy little accident. For the QuantileRegression, the same check didn't pass, and that's when I noticed this mistake.

tests/test_estimators/test_quantile_regression.py
```diff
@@ -0,0 +1,111 @@
...
+
+@pytest.mark.parametrize("coefs, intercept", test_batch)
+@pytest.mark.parametrize("quantile", np.linspace(0, 1, 7))
+def test_quantile(coefs, intercept, quantile):
+    """Tests with noise on an easy problem. A good score should be possible."""
+    X, y = _create_dataset(coefs, intercept, noise=1.0)
+    quant = QuantileRegression(quantile=quantile)
+    quant.fit(X, y)
+
+    np.testing.assert_almost_equal((quant.predict(X)>=y).mean(), quantile, decimal=2)
```
koaning, line 54: I didn't have the time to download and go fully into the numerics myself, but from my end ... the fact that this test passes gives me enough confidence that it's green to merge. 

Before I do that. @MBrouns just to check if you had any comments?

koaning (commented):
Green on my end, just want to give @MBrouns the opportunity to pitch in.

MBrouns:
I'm not sure whether I love the decorators for the regularization terms. What's wrong with doing

```
def grad_quantile_loss(params):
      return (
          -(
              sample_weight
              * np.where(X @ params < y, self.quantile, 1 - self.quantile)
              * np.sign(y - X @ params)
          )
          @ X
          / X.shape[0]
         + self._grad_loss_regularize(params)
      )
```

Garve:
Hi!

This should be fine, too. But then we should change it for all regressors, perhaps in a different PR.

koaning:
@Garve from my end it's preferable to do that in this PR. Once this PR is in, it's time for another release.

koaning:
Oh! And there is a minor thing actually! Could you double check that the component list in the `README.md` is up to date?

Garve:
Decorators gone, readme is up to date.

koaning:
LGTM

koaning (approved):

PR #481, opened 2021-08-30
allow passing sample_weight to underlying classifier when using the Thresholder meta model

MBrouns:


tests/test_meta/test_thresholder.py
```diff
@@ -85,3 +85,25 @@ def test_refit_auto():
...
+
+@pytest.mark.parametrize('refit', [True, False])
+def test_passes_sample_weight(refit):
+    old_fit = LogisticRegression.fit
+
+    def mock_fit(self, X, y, sample_weight=None):
+        assert sample_weight is not None
+        old_fit(self, X, y)
+
+    old_fit = LogisticRegression.fit
```
koaning, line 98: duplicate from line 92.

tests/test_meta/test_thresholder.py
```diff
@@ -85,3 +85,25 @@ def test_refit_auto():
...
+    lr = LogisticRegression()
+
+    mod = Thresholder(lr, threshold=0.5, refit=refit)
+    np.random.seed(42)
+    X = np.random.normal(0, 1, (100, 3))
+    y = np.random.normal(0, 1, (100,)) < 0
+    weight = np.random.random(100)
+
+    mod.fit(X, y, sample_weight=weight)
+    LogisticRegression.fit = old_fit
```
koaning, line 109: Wouldn't it be safer/neater to do something like; 

```
import copy

TempLogisticRegression = copy.deepcopy(LogisticRegression)
```

That way, you don't need to reset the `fit`.

MBrouns: This is a bit hacky for sure. I just pushed a new version where I subclassed the LR inline instead which is probably nicer

koaning (approved):
LGTM

PR #486, opened 2021-09-09
fix tests sklearn one

koaning:
@MBrouns I think these are the tests that are the culprits for scikit-learn v1.

We can merge this and specify sklearn>=1.0.0 once it's officially released. Sound good?


MBrouns:
There's no need to enforce `sklearn>=1` for end-users though right?

koaning:
That's true. No real need to upgrade to another version of scikit-learn as long as we keep cron-testing for a modern version. 

I do feel like it makes sense to keep the stack modern though. We shouldn't immediately upgrade our stuff, but it'll be a burden for us to maintain many old versions of pandas/scikit-learn.

MBrouns:
Yea that's fair, but sklearn 0.24 is still quite recent

koaning:
True. 

Can I merge this though?

MBrouns:
Ah yes, of course!

MBrouns (approved):

PR #487, opened 2021-09-29
Update test-linux.yml

koaning:
Shouldn't we start testing for 3.9?


MBrouns:
Yep, that makes a lot of sense!

Last time I checked there weren't any numpy or pandas wheels for 3.10 on mac yet making the installation quite slow. I'm not sure how that is now, but if it's not too bad I'd suggest adding 3.10 to the list as well

PR #493, opened 2021-12-09
Capture error message in logfile

Hugovdberg:
When a function that is logged with `sklego.pandas_utils.log_step` raises an error there is no entry in the log. For interactive purposes this might not be a big issue but when running the script automatically it is kind of annoying that the step that failed and the associated error is actually missing from the log. This PR wraps the call to `func` in a `try-except-finally` block to always write the step to the log with a message that it failed (and without trying to parse the result). It also adds a flag to the `log_step` signature whether to capture the actual error message in the log or just the notice that it failed.

### DIFF
--- a/sklego/pandas_utils.py
+++ b/sklego/pandas_utils.py
@@ -1,12 +1,12 @@
+import datetime as dt
 import inspect
+from functools import partial, wraps
 
 import numpy as np
 import pandas as pd
-import datetime as dt
-
-from functools import wraps, partial
-from sklego.common import as_list
 from scipy.ndimage.interpolation import shift
+
+from sklego.common import as_list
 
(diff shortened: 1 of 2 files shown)


MBrouns:
I really like this addition! it looks good to me, what do you think @koaning ?

koaning:
This is a totally sensible addition. In fact, I'll gladly make a new release just to include this feature 👍

koaning (approved):

koaning:
New release is out there. 

@Hugovdberg got a twitter? I usually like to give shout-outs to PRs like this.

Hugovdberg:
@koaning I'm Hugovandenberg on Twitter

Issue #433, opened 2021-01-23
[BUG]  AttributeError: 'GroupedPredictor' object has no attribute 'predict_proba'


cmougan:
I am not sure if it's me that I am not using correctly the API.

When doing a classification problem and using a GroupedPredictor to estimate probabilities I can't get the probabilities

```python
from sklego.meta import GroupedPredictor
import numpy as np
from sklearn.linear_model import LogisticRegression
from sklego.datasets import load_chicken

df = load_chicken(as_frame=True)

# Create a binary target
X = df.drop(columns='weight')
y = np.where(df.weight>df.weight.mean(),1,0)

mod = GroupedPredictor(LogisticRegression(), groups=["diet"])

mod.fit(X,y)

mod.predict_proba(X)
```

`
Traceback (most recent call last):
  File "bug.py", line 25, in <module>
    mod.predict_proba(X)
AttributeError: 'GroupedPredictor' object has no attribute 'predict_proba'
`

koaning:
It might be the case that since our `GroupedPredictor` needs to support both regression and classification that we didn't implement `predict_proba`.

koaning:
Just check the API docs [here](https://scikit-lego.netlify.app/api/meta.html#sklego.meta.GroupedPredictor). Yeah, predict_proba isn't implemented. I'm open to a PR for this feature as long as we make sure we add meaningful error messages when users try to use `predict_proba` on a regression model. 

Also, not every classification model allows for `predict_proba` either.

MBrouns:
`scikit-learn` uses a `@if_delegate_has_method` decorator in these circumstances. (https://scikit-learn.org/stable/modules/generated/sklearn.utils.metaestimators.if_delegate_has_method.html). It raises an `AttributeError` when the method doesn't exist on the base estimator, which seems fine to me

cmougan:
> It might be the case that since our `GroupedPredictor` needs to support both regression and classification that we didn't implement `predict_proba`. Can you confirm that `predict` does work?

I do confirm that in this exact code, if you change mod.predict_proba, by mod. predict you get an output

skylarbpayne:
This can be closed, right?

PR #494, opened 2021-12-11
Grouped pred_proba

david26694:
Hi @koaning and @MBrouns, this implements https://github.com/koaning/scikit-lego/issues/433.

Two things I'm not quite happy about:
* I've basically copied all the predict-related functions in GroupedPredictor and created their predict_proba analog. Can you think of a way of not duplicating that much code? I haven't found a simple one.
* I'm not writing any tests specific to the implementation of predict_proba, I've just copied `test_chickweight_can_do_fallback` with its predict_proba analog. Do you have any tests in mind that you would like this code to pass? 

Thanks!


david26694:
I'm thinking  about

> I've basically copied all the predict-related functions in GroupedPredictor and created their predict_proba analog. Can you think of a way of not duplicating that much code? I haven't found a simple one.

Perhaps we could add a `proba=True/False` parameter to the `__predict_groups` function, and all the other __predict functions, in order to avoid code duplication. What do you think?

MBrouns:
I think I'd do something along these lines

```
 def __predict_shrinkage_groups(self, X_group, X_value, method='predict_proba'):
        """Make predictions for all shrinkage groups"""
        # DataFrame with predictions for each hierarchy level, per row. Missing groups errors are thrown here.
        hierarchical_predictions = pd.concat(
            [
                pd.Series(self.__predict_groups(X_group, X_value, level_columns, method='predict_proba'))
                for level_columns in self.group_colnames_hierarchical_
            ],
            axis=1,
        )

        # This is a Series with values the tuples of hierarchical grouping
        prediction_groups = pd.Series(
            [tuple(_) for _ in X_group.itertuples(index=False)]
        )

        # This is a Series of arrays
        shrinkage_factors = prediction_groups.map(self.shrinkage_factors_)

        # Convert the Series of arrays it to a DataFrame
        shrinkage_factors = pd.DataFrame.from_dict(shrinkage_factors.to_dict()).T
        return (hierarchical_predictions * shrinkage_factors).sum(axis=1)

  def __predict_groups(
      self, X_group: pd.DataFrame, X_value: np.array, group_cols=None, method
  ):
        """Predict for all groups"""
        # Reset indices such that they are the same in X_group (reset in __check_grouping_columns),
        # this way we can track the order of the result
        X_value = pd.DataFrame(X_value).reset_index(drop=True)

        if group_cols is None:
            group_cols = X_group.columns.tolist()

        # Make the groups based on the groups dataframe, use the indices on the values array
        group_indices = X_group.groupby(group_cols).indices

        return (
            pd.concat(
                [
                    self.__predict_single_group(group, X_value.loc[indices, :], method=method)
                    for group, indices in group_indices.items()
                ],
                axis=0,
            )
            .sort_index()
            .values.squeeze()

def __predict_single_group(self, group, X, method):
        """Predict a single group by getting its estimator from the fitted dict"""
        # Keep track of the original index such that we can sort in __predict_groups
        index = X.index
        try:
            group_predictor = self.estimators_[group]
        except KeyError:
            if self.fallback_:
                group_predictor = self.fallback_
            else:
                raise ValueError(
                    f"Found new group {group} during predict with use_global_model = False"
                )

        return pd.DataFrame(getattr(group_predictor, method)(X)).set_index(index)
```

Essentially, I changed the `__predict_single_group` method to get the method we want to use for prediction as an argument, and use `getattr()` to get that method out of the group predictor. The other methods can then pass either 'predict' or 'predict_proba' to get the desired result.

I didn't test this so there might be something I missed, but something along these lines should work

david26694:
Thanks @MBrouns, I've refactored according to your proposal, it looks way better IMO. What do you think about testing?

koaning:
@MBrouns any further comments on your end? If not, I think I'm OK to merge this change in.

MBrouns:
I like the changes! Looks good to me

koaning (approved):

koaning:
Thanks @david26694! I'll make a new release once https://github.com/koaning/scikit-lego/pull/495 is merged in as well.

Issue #457, opened 2021-04-09
[BUG] GroupedPredictor doesn't allow for nans in input


pavelkrizek:
The `GroupedPredictor` raises an error (`ValueError: Input contains NaN, infinity or a value too large for dtype('float64').`) even if the `estimator` can handle missing values itself.

```python
from sklego.meta import GroupedPredictor
import numpy as np
from sklearn.pipeline import make_pipeline
from sklearn.linear_model import LinearRegression
from sklearn.impute import SimpleImputer
from sklego.datasets import load_chicken

df = load_chicken(as_frame=True)

X, y = df.drop(columns='weight'),  df['weight']
# create missing value
X.loc[0, 'chick'] = np.nan
model =  make_pipeline(SimpleImputer(), LinearRegression())
group_model = GroupedPredictor(model, groups=["diet"])
group_model.fit(X,y)
```

This could be solved i.e. by exposing parameter `force_all_finite` from sklearn function `check_array` into the `GroupedPredictor`

koaning:
Ah! Good find. 

I'm wondering if it instead makes sense to make a flag that ignores the input alltogether. I can imagine that `accept_sparse` might also be "one of those settings" and to prevent an overload of settings, might it work to just allow users to turn off all checking on the grouper end? 

@MBrouns any thoughts from your end?

pavelkrizek:
Thanks for the quick response. I agree I also think that the "ignore all input checks" flag would be a better solution. Probably not losing any safety net because any sklearn `estimator` will check the inputs anyway.

PR #495, opened 2021-12-18
Make user able to disable input checking on X in GroupedPredictor

lahdjirayhan:
For context, this PR attempts to solve #457.

I make GroupedPredictor accept an extra parameter `check_X`. I also have added new test for the new behavior caused by this parameter.


koaning:
Thanks for the PR! 

There's a bigger PR for that part of the codebase [here](https://github.com/koaning/scikit-lego/pull/494) that I'd like to merge in first. Once that work is in I'll continue with the review here.

tests/test_meta/test_grouped_predictor.py
```diff
@@ -538,3 +540,27 @@ def test_bad_shrinkage_value_error():
...
+@pytest.fixture
+def missing_value_setup():
+    df = load_chicken(as_frame=True)
+
+    X, y = df.drop(columns='weight'),  df['weight']
+    # create missing value
+    X.loc[0, 'chick'] = np.nan
+    model =  make_pipeline(SimpleImputer(), LinearRegression())
+
+    return X, y, model
```
koaning, line 554: Bit of a nitpick, but since this fixture is only used once it might be preferable to just add this code into the `test_missing_check` test.

sklego/meta/_grouped_utils.py
```diff
@@ -61,8 +61,11 @@ def _split_groups_and_values(
     except (KeyError, IndexError):
         raise ValueError(f"Could not drop groups {groups} from columns of X")
 
+    check_X = kwargs.pop('check_X', True)
```
koaning, line 64: I'm curious, why does `check_X` need to be removed from `kwargs`?

sklego/meta/grouped_predictor.py
```diff
@@ -47,13 +50,18 @@ def __init__(
...
         use_global_model=True,
+        check_X=True,
         **shrinkage_kwargs,
     ):
         self.estimator = estimator
         self.groups = groups
         self.shrinkage = shrinkage
         self.use_global_model = use_global_model
         self.shrinkage_kwargs = shrinkage_kwargs
+        self.check_X = check_X
```
MBrouns, line 61: I see a slight issue with how this is implemented. When the estimator is cloned, the `set_params` won't set the `check_X` value of the `_check_kwargs` dict. I'd suggest not going through the `_check_kwargs` dictionary and instead add `self.check_X` explicitly to the calls to `_split_groups_and_values`

MBrouns (changes requested):

sklego/meta/_grouped_utils.py
lahdjirayhan: Last time I checked, the methods under that line that take kwargs (`_check_grouping_columns` and/or `check_array`) complain about unknown parameter `check_X`. Thus my the `.pop`. However, I agree to follow the suggestion here https://github.com/koaning/scikit-lego/pull/495#discussion_r772919039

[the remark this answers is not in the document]
lahdjirayhan: I was thinking that check- and split-related kwargs need to be collected into and then taken from _check_kwargs. Which is maybe a mistake for my part to assume/think.

sklego/meta/grouped_predictor.py
lahdjirayhan: I agree with your suggestion and will push new commits soon. I apologize for not understanding and taking into account the mechanism of `set_params` here.

MBrouns: No worries! It's probably the most confusing bit about writing sklearn extensions

david26694:
Hi! just curious, are you planning on merging this?

MBrouns:
Oh I'm sorry, this totally slipped my mind. I'm gonna look at it right away

MBrouns:
I think this is good to merge. any comments @koaning?

MBrouns (approved):

koaning:
LGTM!

Making a new release now.

david26694:
thank you both :)

PR #503, opened 2022-03-30
PR fixes #501 (https://github.com/koaning/scikit-lego/issues/501)

MarkusDegen:
Try first contribution to another project. Start with a spelling, then preparing #501.
New to python, new to scikit, new to lego

PR #504, opened 2022-04-02
Update documentation dependencies

koaning:
It's about time. The versions we use are from 2019 and are causing build issues in https://github.com/koaning/scikit-lego/pull/503.


koaning:
I've also decided to hard-pin all the documentation version numbers. It should help keep things stable going forward.

Issue #490, opened 2021-10-05
[FEATURE] Allow Thresholder's fit() to take arbitrary input type


jkafrouni:
### Current behavior
Currently, the `fit()` method of `Thresholder` calls `check_X_y` with `dtype=FLOAT_DTYPES`.
This prevents from wrapping for example an entire pipeline, which input is not necessarily a float. Currently, trying to do that will raise an exception. A minimum example to reproduce is the following:

```python
# get data
from sklego.datasets import load_hearts
from sklearn.model_selection import train_test_split

data = load_hearts(return_X_y=True, as_frame=True)
X, y = data.drop(columns='target'), data['target']
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2)

# build a pipeline
from sklearn.pipeline import make_union, make_pipeline
from sklearn.preprocessing import OneHotEncoder, StandardScaler
from sklearn.impute import SimpleImputer
from sklearn.linear_model import LogisticRegression
from sklearn.model_selection import GridSearchCV

from sklego.preprocessing import PandasTypeSelector

cat_features_preprocessing = make_pipeline(PandasTypeSelector(exclude='number'),
                                           SimpleImputer(strategy='constant', fill_value='unknown'), 
                                           OneHotEncoder(categories=[['normal', 'sth', 'fixed']], handle_unknown='ignore'))

num_features_preprocessing = make_pipeline(PandasTypeSelector(include='number'),
                                           SimpleImputer(strategy='median'),
                                           StandardScaler())

preprocessor = make_union(cat_features_preprocessing, num_features_preprocessing)
model = LogisticRegression()
pipe = make_pipeline(preprocessor, model)

pipe.fit(X_train, y_train)

# apply thresholder
from sklego.meta import Thresholder

final_model = Thresholder(pipe, 0.4, refit=False)
final_model.fit(X_train, y_train)
```

Will raise:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-12ce4001dcf8> in <module>
      2 
      3 final_model = Thresholder(pipe, 0.4, refit=False)
----> 4 final_model.fit(X_train, y_train)

/opt/conda/envs/Python-3.8-main/lib/python3.8/site-packages/sklego/meta/thresholder.py in fit(self, X, y)
     50         :return: Returns an instance of self.
     51         """
---> 52         X, y = check_X_y(X, y, estimator=self, dtype=FLOAT_DTYPES)
[...]

Issue #497, opened 2021-12-21
[FEATURE] Allow for Thresholder to accept NaN values if the estimator allows for NaN values


vvandermeij:
It would be nice if the Thresholder does not do a 'check_X_y' when the estimator that it is used on allows for NaN values. A small example is shown below:

```
from sklego.meta import Thresholder
from sklearn.ensemble import HistGradientBoostingClassifier
import numpy as np

X = np.array([[np.nan, 4, 7, 3, 5], [5, 7, 2, 5, 7]])
y = np.array([1, 6, 8, 3, 2])

model = Thresholder(HistGradientBoostingClassifier(), threshold=0.7)
model.fit(X, y)
```

This piece of code gives a ValueError:

---------------------------------------------------------------------------
```
ValueError                                Traceback (most recent call last)
~\AppData\Local\Temp/ipykernel_9212/2229818117.py in <module>
      6 
      7 model = Thresholder(HistGradientBoostingClassifier(), threshold=0.7)
----> 8 model.fit(X, y)

~\Anaconda3\envs\incasso\lib\site-packages\sklego\meta\thresholder.py in fit(self, X, y)
     50         :return: Returns an instance of self.
     51         """
---> 52         X, y = check_X_y(X, y, estimator=self, dtype=FLOAT_DTYPES)
     53         self.estimator_ = clone(self.model)
     54         if not isinstance(self.estimator_, ProbabilisticClassifier):

~\Anaconda3\envs\incasso\lib\site-packages\sklearn\utils\validation.py in check_X_y(X, y, accept_sparse, accept_large_sparse, dtype, order, copy, force_all_finite, ensure_2d, allow_nd, multi_output, ensure_min_samples, ensure_min_features, y_numeric, estimator)
    954         raise ValueError("y cannot be None")
    955 
--> 956     X = check_array(
    957         X,
    958         accept_sparse=accept_sparse,

~\Anaconda3\envs\incasso\lib\site-packages\sklearn\utils\validation.py in check_array(array, accept_sparse, accept_large_sparse, dtype, order, copy, force_all_finite, ensure_2d, allow_nd, ensure_min_samples, ensure_min_features, estimator)
    790 
    791         if force_all_finite:
--> 792             _assert_all_finite(array, allow_nan=force_all_finite == "allow-nan")
    793 
    794     if ensure_min_samples > 0:

[...]

PR #507, opened 2022-04-14
No longer check arrays in Thresholder

koaning:
This PR fixes https://github.com/koaning/scikit-lego/issues/490 and https://github.com/koaning/scikit-lego/issues/497.


koaning:
I'm actually wondering if this approach is what we want. It feels inconsistent with the choices made [here](https://github.com/koaning/scikit-lego/issues/457). @MBrouns do we want to remove the `check_X_y` or do we want to make it configurable?

koaning:
It became a pain to make it configurable. Mainly because we'd need to ask ourselves if we want to check `check_X` or `check_X_y`.

PR #512, opened 2022-06-03
Fix Broken Crontests

koaning:
Reaction to https://github.com/koaning/scikit-lego/actions/runs/2432756672

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,8 +23,8 @@ docs_packages = [
 test_packages = all_packages + [  # we need extras packages for their tests
     "flake8>=3.6.0",
     "nbval>=0.9.1",
-    "pytest==5.4.1",
-    "pytest-xdist==1.34.0",
+    "pytest>=6.2.5",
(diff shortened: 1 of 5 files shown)


koaning:
There's still some tests breaking. In particular: 

```python
def test_log_step_logger(caplog, test_df):
    """Base test of log_step with a logger supplied instead of default print"""
    caplog.clear()

    @log_step(print_fn=logging.info)
    def do_something(df):
        return df.drop(0)

    @log_step(print_fn=logging.info)
    def do_nothing(df, *args, **kwargs):
        return df

    (test_df.pipe(do_nothing).pipe(do_nothing, a="1").pipe(do_something))
    
    assert caplog.messages[0].startswith("[do_nothing(df)]")
    assert caplog.messages[1].startswith("[do_nothing(df, kwargs = {'a': '1'})]")
    assert caplog.messages[2].startswith("[do_something(df)]")
```

It seems the `logging.info` is not registering.

Issue #511, opened 2022-06-02
[BUG] 


zbenmo:
Had to monkey patch as follows:

```
fu = FeatureUnion([
    ("file_extension",  FunctionTransformer(lambda X: X.loc[:, ["file_extension"]])),
    ("1", make_pipeline(
        FunctionTransformer(lambda X: X.loc[:, features]))),
    # ("2", make_pipeline(
    #     FunctionTransformer(lambda X: X.loc[:, ['file_extension']]),
    #     OrdinalEncoder(handle_unknown=-1)
    # )),
])
**GroupedPredictor.classes_ = [0, 1]
GroupedPredictor.decision_function = lambda s, X: GroupedPredictor.predict_proba(s, X)[:, 1]**
cls = GroupedPredictor(RandomForestClassifier(), groups=0)
pipe = make_pipeline(
    fu,
    cls
)

```
To achieve:

```
scores = cross_val_score(pipe, df_features, df_features['label'], cv=5, scoring="roc_auc", error_score='raise') # should be AUC
scores

```

koaning:
@zbenmo in the future, please take the effort of adding ticks (```) such that code blocks render.

That said.  I think we've indeed not implemented a `decision_function` there. I wouldn't mind supporting it though. I'm open to a PR.

PR #513, opened 2022-06-04
added support for decision_function, that was needed for example in a cross validation with auc_roc scoring

zbenmo:
Re: Issue #511


koaning:
Could you also add a unit test that confirms https://github.com/koaning/scikit-lego/issues/511 won't re-occur?

tests/test_meta/test_grouped_predictor.py
```diff
@@ -557,3 +557,11 @@ def test_missing_check():
     with pytest.raises(ValueError) as e:
         GroupedPredictor(model, groups = ['diet']).fit(X, y)
         assert "contains NaN" in str(e)
+
+
+def test_has_decision_function():
+    df = load_chicken(as_frame=True)
+
+    X, y = df.drop(columns='weight'),  df['weight']
+    # This should NOT raise errors
```
koaning, line 566: Could you add a comment that links to the original issue?

koaning:
The implementation looks good to me but @MBrouns figured I'd loop you in as well.

koaning:
Some of the unit tests here seem to be breaking that are unrelated to this PR. I'm working on a fix here: https://github.com/koaning/scikit-lego/pull/512.

koaning:
Just fixed the main branch.

MBrouns:
I'm okay with the implementation for now, but we should probably remove the duplication between `decision-function`, `predict` and `predict_proba` at some point. Doesn't have to be in this PR though

koaning:
Works for me 👍 

Merged it and I'm preparing a new patch release now.

zbenmo:
Thanks. Was a fun experience to make a small contribution to scikit-lego.

PR #521, opened 2022-08-13
Propagate check_X in predict methods of GroupedPredictor

### LINKED ISSUE #520: [BUG] check_X in GroupedPredictor is incomplete (only used for fit, not for predict/predict_proba)
`GroupedPredictor` offers a `check_X` param which can be used to skip checks if the underlying model you are using can handle missing and/or non-float data. However, this is only used during `fit` and not during `predict` or `predict_proba`.

This means that I can fit a model -- but if I try to predict with data that is missing/not float, it will fail.
This is useful for me, and I imagine others, because we may be using an underlying model which can handle categorical variables encoded as string or handles missing values directly (e.g. as in many GBDT modeling frameworks).

Happy to fix this issue myself -- as without this we can't use this wonderful library 😅 and it seems incredibly straightforward

skylarbpayne:
# Context

Closes #520 . An earlier change exposed a parameter `check_X` in `GroupedPredictor` to optionally disable a conversion to a float array (which is required for compatibility with _most_ sklearn predictors). However, there are some predictors which handle missing values and even string variables directly (e.g. CatBoostClassifier for example). `check_X` allows usage of data which cannot be converted to a float array.

As described in #520, there is a small bug where `check_X` is not used when `predict` functions are called. As such, you can currently _fit_ a model with non-float array data, but you cannot predict on non-float array data.

# Change

Propagate `self.check_X` in `GroupedPredictor` to predict-like calls. This includes:

- `predict`
- `predict_proba`
- `decision_function`

# Testing
The extension to the test `tests/test_meta/test_grouped_predictor.py::test_missing_check` initially fails with:

```
FAILED tests/test_meta/test_grouped_predictor.py::test_missing_check - ValueError: Input contains NaN.
```

After implementation, all tests pass.

### DIFF
--- a/sklego/meta/grouped_predictor.py
+++ b/sklego/meta/grouped_predictor.py
@@ -339,7 +339,7 @@ class GroupedPredictor(BaseEstimator):
         check_is_fitted(self, ["estimators_", "groups_", "fallback_"])
 
         X_group, X_value = _split_groups_and_values(
-            X, self.groups, min_value_cols=0, **self._check_kwargs
+            X, self.groups, min_value_cols=0, check_X=self.check_X, **self._check_kwargs
         )
 
         X_group = self.__add_shrinkage_column(X_group)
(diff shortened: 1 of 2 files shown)


koaning:
LGTM. @MBrouns any objections?

MBrouns (approved):

Issue #524, opened 2022-09-02
[FEATURE] Add parameter to allow NaNs in IdentityTransformer


GLevV:
Make parameter `force_all_finite` in `check_array` function available to end user.

koaning:
I'm open to a PR for this, but I'm wondering if we need to check for anything here at all for this component. The `check_array` function checks for many things and I think it might be better to add a single flag for either all the checks or none.

GLevV:
So something like `check_input` bool parameter? And set `True` as default for several releases for backwards compatibility?

PR #527, opened 2022-09-06
Add check_input flag to identitytransformer

GLevV:
will solve #524

### DIFF
--- a/sklego/preprocessing/identitytransformer.py
+++ b/sklego/preprocessing/identitytransformer.py
@@ -9,16 +9,27 @@ class IdentityTransformer(BaseEstimator, TransformerMixin):
     The identity transformer returns what it is fed. Does not apply anything useful.
     The reason for having it is because you can build more expressive pipelines.
     """
+    def __init__(
+        self,
+        check_input: bool = True
+    ):
+        self.check_input = check_input
 
     def fit(self, X, y=None):
         """'Fits' the estimator."""
-        X = check_array(X, copy=True, estimator=self)
+        if self.check_input:
+            X = check_array(X, copy=True, estimator=self)
+        else:
+            X = X.copy()
         self.shape_ = X.shape
         return self
 
     def transform(self, X):
         """'Applies' the estimator."""
-        X = check_array(X, copy=True, estimator=self, )
+        if self.check_input:
+            X = check_array(X, copy=True, estimator=self)
+        else:
+            X = X.copy()
         check_is_fitted(self, 'shape_')
         if X.shape[1] != self.shape_[1]:
             raise ValueError(f"Wrong shape is passed to transform. Trained on {self.shape_[1]} cols got {X.shape[1]}")
--- a/sklego/preprocessing/identitytransformer.py
+++ b/sklego/preprocessing/identitytransformer.py
@@ -25,6 +25,8 @@ class IdentityTransformer(BaseEstimator, TransformerMixin):
         """'Fits' the estimator."""
         if self.check_X:
             X = check_array(X, copy=True, estimator=self)
+        else:
+            X = X.copy()
         self.shape_ = X.shape
         return self
 
@@ -32,6 +34,8 @@ class IdentityTransformer(BaseEstimator, TransformerMixin):
         """'Applies' the estimator."""
         if self.check_X:
             X = check_array(X, copy=True, estimator=self)
+        else:
+            X = X.copy()
         check_is_fitted(self, 'shape_')
         if X.shape[1] != self.shape_[1]:
             raise ValueError(f"Wrong shape is passed to transform. Trained on {self.shape_[1]} cols got {X.shape[1]}")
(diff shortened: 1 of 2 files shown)


koaning:
Thanks for the PR! 

I checked another method and found out that we called it `check_X`. For consistency, could you use [the same name](https://github.com/koaning/scikit-lego/blob/main/sklego/meta/grouped_predictor.py#L36)? 

Also, could you add a test?

sklego/preprocessing/identitytransformer.py
```diff
@@ -9,16 +9,27 @@ class IdentityTransformer(BaseEstimator, TransformerMixin):
     The identity transformer returns what it is fed. Does not apply anything useful.
     The reason for having it is because you can build more expressive pipelines.
     """
+    def __init__(
+        self,
+        check_X: bool = True
```
koaning, line 14: When I think about what the identitytransformer is supposed to do, I think `False` would be a better default here.

sklego/preprocessing/identitytransformer.py
```diff
@@ -9,16 +9,27 @@ class IdentityTransformer(BaseEstimator, TransformerMixin):
...
+    ):
+        self.check_X = check_X
 
     def fit(self, X, y=None):
         """'Fits' the estimator."""
-        X = check_array(X, copy=True, estimator=self)
+        if self.check_X:
+            X = check_array(X, copy=True, estimator=self)
+        else:
+            X = X.copy()
```
koaning, line 23: Why does `X` need to be copied here?

sklego/preprocessing/identitytransformer.py
```diff
@@ -9,16 +9,27 @@ class IdentityTransformer(BaseEstimator, TransformerMixin):
...
         self.shape_ = X.shape
         return self
 
     def transform(self, X):
         """'Applies' the estimator."""
-        X = check_array(X, copy=True, estimator=self, )
+        if self.check_X:
+            X = check_array(X, copy=True, estimator=self)
+        else:
+            X = X.copy()
```
koaning, line 32: Why does `X` need to be copied here?

koaning (changes requested):
Thanks for writing the test! I added some points for review.

sklego/preprocessing/identitytransformer.py
GLevV: I thought setting it to True here would be better for backward compatibility, since older versions are checking input array

sklego/preprocessing/identitytransformer.py
GLevV: `check_array` function also copies input data. The idea was even if array is not checked, data is copied for consistency. Should I remove it?

sklego/preprocessing/identitytransformer.py
koaning: I see the reasoning, but pragmatically the `IdentityTransformer` is mainly used for no-op use-cases. So it feels fine to omit the `copy`. Unless @MBrouns you see a good reason to keep it?

tests/test_preprocessing/test_identitytransformer.py
```diff
@@ -24,5 +24,14 @@ def test_estimator_checks(test_fn):
 
 def test_same_values(random_xy_dataset_regr):
     X, y = random_xy_dataset_regr
-    X_new = IdentityTransformer().fit_transform(X)
+    X_new = IdentityTransformer(check_X=True).fit_transform(X)
     assert np.isclose(X, X_new).all()
+
+
+def test_nan_inf(random_xy_dataset_regr):
+    X, y = random_xy_dataset_regr
```
koaning, line 32: It's a minor detail, and not a reason to block the PR ... 

... but in future PRs: it can be a good idea to have a docstring that links back to the GitHub issue. That way, if a test fails, there's a paper trail to the original issue that it's trying to fix.

koaning:
@GLevV I think I stand corrected, without the `X.copy()` the scikit-learn tests seem to break. My bad!

Feel free to put the copy commands back 😅

GLevV:
> The breaking tests suprised me but then I noticed that the standard tests should be:
> 
> ```python
> @pytest.mark.parametrize(
>     "test_fn",
>     select_tests(
>         flatten([general_checks, nonmeta_checks, transformer_checks]),
>         exclude=[
>             "check_sample_weights_invariance",
>             "check_sample_weights_list",
>             "check_sample_weights_pandas_series"
>         ]
>     )
> )
> def test_estimator_checks(test_fn):
>     test_fn(IdentityTransformer.__name__, IdentityTransformer(check_X=True))
> ```
> 
> Note the `check_X=True` in that final line. When you add that the tests pass. It also fixes it such that you don't need to run `X.copy`.

Sorry, my bad

koaning (approved):
LGTM

koaning:
Thanks for the PR!

PR #529, opened 2022-09-12
Adding optimizer option to linear models

### LINKED ISSUE #499: custom optimizer for QuantileRegression  : SLSQP is slow and not performing for one hot data
is it possible to use custom optimizer for QuantileRegression optimizer but not current the SLSQP algorithm
for example to use available in scipy ? 
https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize

since  SLSQP is slow and not performing for one hot data 

https://github.com/koaning/scikit-lego/blob/main/sklego/linear_model.py#L759

GLevV:
Fixes #499


MBrouns (changes requested):

tests/test_estimators/test_imbalanced_linear_regression.py
```diff
@@ -27,31 +27,31 @@ def _create_dataset(coefs, intercept, noise=0.0):
 def test_coefs_and_intercept__no_noise(coefs, intercept):
     """Regression problems without noise."""
     X, y = _create_dataset(coefs, intercept)
-    imb = ImbalancedLinearRegression()
-    imb.fit(X, y)
-
-    assert imb.score(X, y) > 0.99
+    for method in ("SLSQP", "TNC", "L-BFGS-B"):
```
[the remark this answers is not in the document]
GLevV: You mean here
https://github.com/GLevV/scikit-lego/blob/a059bfa755f3ef1263165765ae8d86f24cab86f7/tests/test_estimators/test_imbalanced_linear_regression.py#L106
for example add something like
`(ImbalancedLinearRegression.__name__, ImbalancedLinearRegression(method='TNC')),`
and etc?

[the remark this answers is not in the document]
GLevV: TNC and L-BFGS-B are failing on `check_dtype_object` when y has `object` dtype

GLevV:
So right now the fix is to set `y_numeric=True`, which will enforce y to be numeric. But still, it's a lot of changes for something that seemed as a small addition

sklego/linear_model.py
```diff
@@ -606,9 +623,9 @@ def fit(self, X, y, sample_weight=None):
             loss,
             x0=np.zeros(self.n_features_in_ + d),
             bounds=bounds,
-            method="SLSQP",
+            method=self.method,
             jac=grad_loss,
-            tol=1e-20,
+            tol=1e-6,
```
MBrouns, line 628: Is it necessary that the default here changed?

MBrouns:
One final question on the default `tol` value but otherwise it looks good!

sklego/linear_model.py
GLevV: I guess it's a leftover, for SLSQP default tol (ftol) is 1e-6, for tnc it's 0 and for lbfgsb it's 2.220446049250313e-09, so I change it to maximum among the three. Should I revert it to 1e-20?

MBrouns: Yea that'd be good as it's technically an unnecessary breaking change now.

MBrouns (approved):

MBrouns:
Thanks for the PR!

Issue #530, opened 2022-09-12
[FEATURE] Pass additional parameters to fit underlying estimator in `EstimatorTransformer`


CarloLepelaars:
In `EstimatorTransformer` the underlying estimator is being fitted without the ability to pass along additional arguments to `self.estimator_.fit`. 

This limits use cases for `EstimatorTransformer`. For example, if the underlying estimator is an `XGBClassifier` we would like to be able to pass `eval_set` to monitor validation performance and enable early stopping. This is currently not possible. Adding `*args, **kwargs` should fix this issue. 

https://github.com/koaning/scikit-lego/blob/b4d087f0131ff164e6feebf238356ba6512b3635/sklego/meta/estimator_transformer.py#L31

CarloLepelaars:
@koaning 
> In the future, please don't make a PR until the direction of the solution has been discussed in the issues.
> 
> I'll pick up the conversation there.

Ok, no problem! Will keep that in mind.

koaning:
It's a bit of an sklearn antipattern to pass lots of settings via `.fit()`. Is XGBoost part of sklearn or a 3rd party lib?

koaning:
It's a bit of an sklearn antipattern to pass lots of settings via `.fit()`. Is XGBoost part of sklearn or a 3rd party lib?

CarloLepelaars:
XGBoost is a 3rd party library maintained by [dmlc](https://github.com/dmlc). 

I agree hyperparameters shouldn't be passed via `.fit()`. Unfortunately, some parameters like `eval_set` often can be passed only with `.fit()`. 

Other example use cases are [CatBoost parameters](https://catboost.ai/en/docs/concepts/python-reference_catboost_fit) and [Lightgbm parameters](https://lightgbm.readthedocs.io/en/latest/pythonapi/lightgbm.LGBMRegressor.html#lightgbm.LGBMRegressor.fit) that can only be passed through `.fit()`.

koaning:
I'm a bit uneasy to extend this library beyond scikit-learn because the dependencies quickly start to stack up. @MBrouns what's your opinion on this?

MBrouns:
sklearn does describe the use of `kwargs` in fit methods: https://scikit-learn.org/stable/developers/develop.html#fitting. I'm not sure I like varargs, but I don't see a lot of problems with varkwargs. I would like to see a test added in the PR though before accepting it

koaning:
TIL. 

[...]

PR #532, opened 2022-09-13
Pass kwargs to `fit` method of `EstimatorTransformer`.

CarloLepelaars:
Option to pass along `**kwargs` to underlying estimator in `EstimatorTransformer`. This allows for passing parameters like `sample_weight` in scikit-learn objects and parameters like `eval_set` for XGBoost, Catboost and LightGBM that can only be passed in `.fit()`.

Resolves Issue #530 

scikit-learn describes usage of kwargs for passing optional data-dependent parameters here:
https://scikit-learn.org/stable/developers/develop.html#fitting

### DIFF
--- a/tests/test_meta/test_estimatortransformer.py
+++ b/tests/test_meta/test_estimatortransformer.py
@@ -75,6 +75,7 @@ def test_shape(random_xy_dataset_regr):
 def test_shape_multitarget(random_xy_dataset_multitarget):
     X, y = random_xy_dataset_multitarget
     m = X.shape[0]
+    print('hello')
     n = y.shape[1]
     pipeline = Pipeline(
         [
@@ -98,9 +99,14 @@ def test_kwargs(random_xy_dataset_clf):
             )
         ]
     )
-    # If the additional parameter sample_weight is passed transform_no_kwargs and transform_kwargs should not be equal.
     transform_no_kwargs = pipeline.fit(X, y).transform(X)
-    transform_kwargs = pipeline.fit(X, y, model__sample_weight=0.1).transform(X)
+
+    # First sample is weighted twice as much
+    sample_weights = np.ones(shape=len(y))
+    sample_weights[0] = 2
+
+    # If sample_weights are passed transform_no_kwargs and transform_kwargs should not be equal.
+    transform_kwargs = pipeline.fit(X, y, model__sample_weight=sample_weights).transform(X)
     assert not np.array_equal(transform_no_kwargs, transform_kwargs)
 
     # Fitting with sample_weight=1. should be the same as fitting with sample_weight not specified.
--- a/tests/test_meta/test_estimatortransformer.py
+++ b/tests/test_meta/test_estimatortransformer.py
@@ -1,5 +1,6 @@
 import pytest
 import numpy as np
+from unittest.mock import Mock
 from sklearn import clone
 from sklearn.dummy import DummyClassifier
 from sklearn.multioutput import MultiOutputRegressor
@@ -91,16 +92,12 @@ def test_kwargs(random_xy_dataset_clf):
     """ Test if kwargs are properly passed to an underlying estimator. """
     X, y = random_xy_dataset_clf
     pipeline = EstimatorTransformer(LinearRegression())
-    transform_no_kwargs = pipeline.fit(X, y).transform(X)
 
     # First sample is weighted twice as much
     sample_weights = np.ones(shape=len(y))
     sample_weights[0] = 2
 
-    # If sample_weights are passed transform_no_kwargs and transform_kwargs should not be equal.
-    transform_kwargs = pipeline.fit(X, y, model__sample_weight=sample_weights).transform(X)
-    assert not np.array_equal(transform_no_kwargs, transform_kwargs)
-
-    # Fitting with sample_weight=1. should be the same as fitting with sample_weight not specified.
-    transform_kwargs2 = pipeline.fit(X, y, model__sample_weight=1.).transform(X)
-    assert np.array_equal(transform_no_kwargs, transform_kwargs2)
+    # Check if fit method is explicitly called with sample weights
+    pipeline.fit = Mock()
+    pipeline.fit(X, y, model__sample_weight=sample_weights)
+    pipeline.fit.assert_called_with(X, y, model__sample_weight=sample_weights)
--- a/tests/test_meta/test_estimatortransformer.py
+++ b/tests/test_meta/test_estimatortransformer.py
@@ -75,7 +75,6 @@ def test_shape(random_xy_dataset_regr):
 def test_shape_multitarget(random_xy_dataset_multitarget):
     X, y = random_xy_dataset_multitarget
     m = X.shape[0]
-    print('hello')
     n = y.shape[1]
     pipeline = Pipeline(
         [
--- a/tests/test_meta/test_estimatortransformer.py
+++ b/tests/test_meta/test_estimatortransformer.py
@@ -102,3 +102,7 @@ def test_kwargs(random_xy_dataset_clf):
     transform_no_kwargs = pipeline.fit(X, y).transform(X)
     transform_kwargs = pipeline.fit(X, y, model__sample_weight=0.1).transform(X)
     assert not np.array_equal(transform_no_kwargs, transform_kwargs)
+
+    # Fitting with sample_weight=1 should be the same as fitting with sample_weight not specified.
+    transform_kwargs2 = pipeline.fit(X, y, model__sample_weight=1.).transform(X)
+    assert np.array_equal(transform_no_kwargs, transform_kwargs2)
--- a/tests/test_meta/test_estimatortransformer.py
+++ b/tests/test_meta/test_estimatortransformer.py
@@ -98,11 +98,11 @@ def test_kwargs(random_xy_dataset_clf):
             )
         ]
     )
-    # If the additional parameter sample_weight is passed transform_no_kwargs and transform_kwargs cannot be equal.
+    # If the additional parameter sample_weight is passed transform_no_kwargs and transform_kwargs should not be equal.
     transform_no_kwargs = pipeline.fit(X, y).transform(X)
     transform_kwargs = pipeline.fit(X, y, model__sample_weight=0.1).transform(X)
     assert not np.array_equal(transform_no_kwargs, transform_kwargs)
 
-    # Fitting with sample_weight=1 should be the same as fitting with sample_weight not specified.
+    # Fitting with sample_weight=1. should be the same as fitting with sample_weight not specified.
     transform_kwargs2 = pipeline.fit(X, y, model__sample_weight=1.).transform(X)
     assert np.array_equal(transform_no_kwargs, transform_kwargs2)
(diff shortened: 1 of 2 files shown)


koaning:
This is advice unrelated to the PR, but in general, it's advisable to work on a branch that isn't main. Especially when you're potentially working on multiple features.

CleanShot 2022-09-14 at 15 46 16

tests/test_meta/test_estimatortransformer.py
```diff
@@ -85,3 +85,24 @@ def test_shape_multitarget(random_xy_dataset_multitarget):
...
+            (
+                "model",
+                EstimatorTransformer(Ridge(random_state=0))
+            )
+        ]
+    )
+    # If the additional parameter sample_weight is passed transform_no_kwargs and transform_kwargs should not be equal.
+    transform_no_kwargs = pipeline.fit(X, y).transform(X)
+    transform_kwargs = pipeline.fit(X, y, model__sample_weight=0.1).transform(X)
+    assert not np.array_equal(transform_no_kwargs, transform_kwargs)
```
koaning, line 104: I'm a bit surprised that the predictions are different when we apply a uniform weight to all the examples. Aren't we just dividing the loss function by a constant here?

CarloLepelaars: > This is advice unrelated to the PR, but in general, it's advisable to work on a branch that isn't main. Especially when you're potentially working on multiple features.
> 
> CleanShot 2022-09-14 at 15 46 16

👍 That makes sense!

Thanks for all this feedback, really appreciate it!

CarloLepelaars: > This is advice unrelated to the PR, but in general, it's advisable to work on a branch that isn't main. Especially when you're potentially working on multiple features.
> 
> CleanShot 2022-09-14 at 15 46 16

That makes sense, will create a new branch next time when working on a features. Thanks for all this feedback, really appreciate it!

CarloLepelaars:
> This is advice unrelated to the PR, but in general, it's advisable to work on a branch that isn't main. Especially when you're potentially working on multiple features.
> 
> CleanShot 2022-09-14 at 15 46 16

That makes sense, will create a new branch next time when working on a features. Thanks for all this feedback, really appreciate it!

tests/test_meta/test_estimatortransformer.py
koaning: I suppose so, it's more that I'm worried about what the test is testing when it's assuming an unintuitive result to be true.

CarloLepelaars: [The data is rescaled](https://github.com/scikit-learn/scikit-learn/blob/36958fb240fbe435673a9e3c52e769f01f36bec0/sklearn/linear_model/_base.py#L326) and it seems to me here is where a constant sample_weight < 1 can change model training, namely learning less from each sample. I adjusted the test to use `LinearRegression` so its easier to analyse this. Rescaling data is being called [here](https://github.com/scikit-learn/scikit-learn/blob/36958fb240fbe435673a9e3c52e769f01f36bec0/sklearn/linear_model/_base.py#L702).

I also adjusted the test to use an array of sample weights (i.e. all samples weighted equally except the 1st sample has a weight of 2).

tests/test_meta/test_estimatortransformer.py
```diff
@@ -75,6 +75,7 @@ def test_shape(random_xy_dataset_regr):
 def test_shape_multitarget(random_xy_dataset_multitarget):
     X, y = random_xy_dataset_multitarget
     m = X.shape[0]
+    print('hello')
```
MBrouns, line 78:
```suggestion

```

tests/test_meta/test_estimatortransformer.py
```diff
@@ -85,3 +86,29 @@ def test_shape_multitarget(random_xy_dataset_multitarget):
...
+    """ Test if kwargs are properly passed to an underlying estimator. """
+    X, y = random_xy_dataset_clf
+    pipeline = Pipeline(
+        [
+            (
+                "model",
+                EstimatorTransformer(LinearRegression())
+            )
+        ]
+    )
```
MBrouns, line 101:
```suggestion
    pipeline = EstimatorTransformer(LinearRegression())
```

tests/test_meta/test_estimatortransformer.py
MBrouns: Rather than checking whether the results are the same or different, I'd rather test more explicitly whether the right parameters are called. 

I think the best thing to do here would be to make the estimator passed to the EstimatorTransformers a `unittest.Mock` object with a `fit` method. Then we can use `assert_called_with` to verify the sample_weights are indeed passed. That way we make sure we test explicitly the parameter passing, rather than any other reason the results might be different (randomness for example)

MBrouns (changes requested):

tests/test_meta/test_estimatortransformer.py
```diff
@@ -85,3 +86,18 @@ def test_shape_multitarget(random_xy_dataset_multitarget):
...
+    pipeline = EstimatorTransformer(LinearRegression())
+
+    # First sample is weighted twice as much
+    sample_weights = np.ones(shape=len(y))
+    sample_weights[0] = 2
+
+    # Check if fit method is explicitly called with sample weights
+    pipeline.fit = Mock()
+    pipeline.fit(X, y, model__sample_weight=sample_weights)
+    pipeline.fit.assert_called_with(X, y, model__sample_weight=sample_weights)
```
CarloLepelaars, line 103: @MBrouns Is this the correct way to use a `Mock` method for checking if the `sample_weight` argument is explcitily passed?

MBrouns: almost, what you're checking now is whether the fit method of the `EstimatorTransformer` is called correctly (which it is because you do exactly that on line 102). We want to instead check if the `estimator` in the `Estimatortransformer` is called correctly


```
estimator = Mock()
pipeline = EstimatorTransformer(estimator)

...

pipeline.fit(X, y, model__sample_weight=sample_weights)
estimator.fit.assert_called_with(X, y, model__sample_weights=sample_weights)
```

CarloLepelaars: That makes sense, although struggling with implementation. `self.estimator_` is only introduced within the fit method as a clone of `self.estimator`. The adding of `Mock()` should therefore be created just after the clone, but before `EstimatorTransformer.estimator_.fit` in `EstimatorTransformer.fit`. Any idea how I can wrap `EstimatorTransformer.fit` so I can introducte `Mock()` just before the underlying estimator is being fitted or is there another way?

MBrouns: Ah right, I forgot about the cloning! Interesting case. That essentially means we need to change how `clone` works, which I think we can do with `unittest.mock.patch`. 

```
from unittest.mock import patch


@patch('sklearn.clone')
def test_kwargs(patched_clone, random_xy_dataset_clf):
    """ Test if kwargs are properly passed to an underlying estimator. """
    X, y = random_xy_dataset_clf
    estimator = Mock()
    patched_clone.return_value = estimator
    
    pipeline = EstimatorTransformer(estimator)
    sample_weights = np.ones(shape=len(y))
    pipeline.fit(X, y, model__sample_weight=sample_weights)

    estimator.assert_called_with(X, y, model__sample_weight=sample_weights)
```

CarloLepelaars: Hmm, still some issues with that. `sklearn.clone` only works on scikit-learn objects so if it is `Mock()` it breaks (`TypeError: 'Mock' object is not iterable`).

If we set `estimator.fit = Mock()` and check `estimator.fit.assert_called_with` we get `AssertionError: expected call not found.` Not sure why that is, but it looks like it is not tracking the cloned object.

Solution I was thinking of:
```python
import sklearn
from unittest.mock import Mock
from unittest.mock import patch

@patch('sklearn.clone')
def test_kwargs(patched_clone, random_xy_dataset_clf):
    """ Test if kwargs are properly passed to an underlying estimator. """
    X, y = random_xy_dataset_clf
    estimator = LinearRegression()
    estimator.fit = Mock()
    patched_clone.return_value = estimator

    sample_weights = np.ones(shape=len(y))
    pipeline = EstimatorTransformer(estimator)
    pipeline.fit(X, y, sample_weight=sample_weights)

    estimator.fit.assert_called_with(X, y, sample_weight=sample_weights)
```

CarloLepelaars: To be honest I'm stuck on properly tracking this cloned object with `Mock`, @MBrouns. Do you have any suggestions on how to proceed?

MBrouns: Okay that was indeed a bit trickier than I figured at first. The following works:

```

@patch('sklego.meta.estimator_transformer.clone')
def test_kwargs(patched_clone, random_xy_dataset_clf):
    """ Test if kwargs are properly passed to an underlying estimator. """
    X, y = random_xy_dataset_clf
    estimator = Mock()
    patched_clone.return_value = estimator

    sample_weights = np.ones(shape=len(y))
    pipeline = EstimatorTransformer(estimator)
    pipeline.fit(X, y, sample_weight=sample_weights)

    # We can't use `assert_called_with` becuase that compares by `==` which is ambiguous
    # on numpy arrays
    np.testing.assert_array_equal(
        X, estimator.fit.call_args.args[0]
    )
    np.testing.assert_array_equal(
        y, estimator.fit.call_args.args[1]
    )
    np.testing.assert_array_equal(
        sample_weights, estimator.fit.call_args.kwargs['sample_weight']
    )
```

The patch I suggested was wrong because in the `estimator_transformer.py` we import clone in to its namespace, rather than doing `import sklearn; sklearn.clone(...)` 

Also, the assert_called_with doesn't work on numpy arrays as it uses `==` by default, so I had to dig in to the `call_args`

CarloLepelaars: Interesting, thank you! This works like a charm! The tests are passing now. Looks like this is resolved.

MBrouns: It seems that that functionality is implemented in 3.8 onwards indeed (https://bugs.python.org/issue21269). Could you try this:

```
  np.testing.assert_array_equal(
        X, estimator.fit.call_args[0][0]
    )
    np.testing.assert_array_equal(
        y, estimator.fit.call_args[0][1]
    )
    np.testing.assert_array_equal(
        sample_weights, estimator.fit.call_args[0]['sample_weight']
    )
```

CarloLepelaars: 👍 Nice you could find the specific Python bugs issue. 

I pushed the code above after getting tests passing locally for Python3.7. I only had to change the last line, because `kwargs` are contained in the 2nd element of `call_args`. Like this:
```python
np.testing.assert_array_equal(
        sample_weights, estimator.fit.call_args[1]['sample_weight']
    )
```

CarloLepelaars: YES, that's it! Thank you, @MBrouns! 

Learned a lot from this.

MBrouns (approved):

Issue #533, opened 2022-09-14
[BUG]  'EstimatorTransformer' object has no attribute 'get_feature_names_out'


CarloLepelaars:
When calling `get_feature_names_out` on `EstimatorTransformer` or a `Pipeline` that contains `EstimatorTransformer` you will get the following error:
`AttributeError: 'EstimatorTransformer' object has no attribute 'get_feature_names_out'`

Minimal reproducible example:
```python
from sklego.meta import EstimatorTransformer
from sklearn.linear_model import LinearRegression
EstimatorTransformer(LinearRegression()).get_feature_names_out(None)
```

I thought this issue was resolved in `scikit-learn >= 1.1` ([get_feature_names_out Available in all Transformers release highlight](https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_1_0.html#get-feature-names-out-available-in-all-transformers)), but apparently a manual implementation of `get_feature_names_out` is still needed for custom scikit-learn transformers.

Proposed solution sketch:
```python
from sklearn.utils.validation import check_is_fitted
class EstimatorTransformer(TransformerMixin, MetaEstimatorMixin, BaseEstimator):
    .
    .
    .
    def fit(X, y, **kwargs):
        .
        .
        .
        # Store how many output columns estimator has
        self.output_len_ = y.shape[1] if self.multi_output_ else 1
        .
        .
    
    def get_feature_names_out(self, feature_names_out=None) -> list:
        """ 
        Get names for output of EstimatorTransformer. 
        Estimator must be fitted first before this function can be called. 
        """
        check_is_fitted(self.estimator_)
        if self.multi_output_:
            feature_names = [f"prediction_{i}" for i in range(self.output_len_)]
        else: 
            feature_names = ["prediction"]
        return feature_names
        
```

Happy to contribute this if you agree with the proposed solution idea. If this a general problem I'm also open to work on implementing `get_feature_names_out` for other transformers in `scikit-lego`.

koaning:
Minor ask: you can attach a language to a code-block to get syntax highlighting. Like so: 

````
```python
import pandas as pd
```
````

[...]

PR #539, opened 2022-09-27
`get_feature_names_out` for `EstimatorTransformer`

CarloLepelaars:
This PR adds `get_feature_names_out` functionality for `EstimatorTransformer`. This is a desirable method for `EstimatorTransformer` because its output is used as input for a subsequent estimator. 

- [x] Implement `get_feature_names_out` for `EstimatorTransformer`.
- [x] Write core tests for implementation.  
- [x] Create tests to ensure `get_feature_names_out` works with a `FeatureUnion` component.
- [x] Research if `get_feature_names_out` implementation is necessary and desired for other Meta estimators.

@koaning I think it is not necessary to implement `get_feature_names_out` for the other estimators in `sklego.meta`, because it seems all these estimators are used as the last component in a `Pipeline`, except for `EstimatorTransformer`. `EstimatorTransformer` will often be a component somewhere in the middle of a `Pipeline` and therefore `get_feature_names_out` is desirable. That said I'm happy to work on other implementations of `get_feature_names_out` in `sklego.meta` if you disagree with this reasoning.

Solves Issue #533


koaning:
Just to check @CarloLepelaars there are _a lot_ of commits here. Is the PR meant to be reviewed? If not, could you add [WIP] in the title to indicate it's a work in progress? It's also fine to only open the PR once it's ready.

sklego/meta/estimator_transformer.py
```diff
@@ -38,6 +39,34 @@ def transform(self, X):
...
         """
-        check_is_fitted(self, "estimator_")
+        check_is_fitted(self)
         output = getattr(self.estimator_, self.predict_func)(X)
         return output if self.multi_output_ else output.reshape(-1, 1)
+
+    def get_feature_names_out(self, feature_names_out=None) -> list:
+        """
+        Defines descriptive names for each output of the (fitted) estimator.
+        :param feature_names_out: Redundant parameter for which the contents are ignored in this function.
```
koaning, line 49: Have you checked if this renders nicely in sphinx? I wouldn't mind some newlines in the docstring.

koaning (approved):
LGTM. 

I left a minor nitpick tho.

CarloLepelaars:
Hey, yes this feature is ready and meant to be reviewed. I'll make sure to add [WIP] if something would be still work in progress

As for the commits there are some commits here that are already merged to scikit-lego from a previous PR, but somehow it still shows when I create a new local branch. Everything up to commit #f06a7af is duplicate. Do you know how we can clean up the commit history?

CarloLepelaars:
@koaning, shall we go ahead and merge this one to `main`?

sklego/meta/estimator_transformer.py
```diff
@@ -38,6 +39,36 @@ def transform(self, X):
         Returns array of shape `(X.shape[0], )` if estimator is not multi output.
         For multi output estimators an array of shape `(X.shape[0], y.shape[1])` is returned.
         """
-        check_is_fitted(self, "estimator_")
```
koaning, line 41: Why are we no longer checking `estimator_`?

sklego/meta/estimator_transformer.py
```diff
@@ -38,6 +39,36 @@ def transform(self, X):
...
+        return feature_names
+
+    def __sklearn_is_fitted(self) -> bool:
+        """
+        Custom additional requirements that need to be satisfied to pass check_is_fitted.
+
+        :return: Boolean indicating if the additional requirements
+        for determining check_is_fitted are satisfied.
+        """
+        has_fit_attr = all(hasattr(self, attr) for attr in ["multi_output_", "output_len_"])
```
koaning, line 72: Is it possible to make `["multi_output_", "output_len_"]` more robust? I'm wondering if we shouldn't check for all properties that end with a single `_`, like `estimators_`.

koaning (changes requested):
Found two nitpicks worth discussing. Sorry for the delay!

sklego/meta/estimator_transformer.py
CarloLepelaars: I moved this call to `__sklearn_is_fitted`, so the behavior hasn't changed.

`check_is_fitted` will call `self.__sklearn_is_fitted` ([sklearn source](https://github.com/scikit-learn/scikit-learn/blob/626b4608d4f840af7c37bff2ccb38fcfd2ef594f/sklearn/utils/validation.py#L1338), which contains `check_is_fitted(self.estimator_)`. 

Not sure if this obscures the behavior too much. Implementing `__sklearn_is_fitted` works well for sklearn transformers, but I can imagine not that many people are familiar with it.

sklego/meta/estimator_transformer.py
CarloLepelaars: Good point, will check if there is a more dynamic way to get all properties. Else I'll add all properties that are created during `.fit`.

sklego/meta/estimator_transformer.py
koaning: I'm cool with the implementation, though I'd prefer to have a comment around explaining it. Should their internals change, we'll have a link.

CarloLepelaars: Sounds good! Added a comment for clarification.

koaning: Could you also add a link to the sklearn source? Also; since this is a hidden method; comments may be better, otherwise tools like sphinx might accidentally document it.

CarloLepelaars: Sure! Turned some lines into comments and added link to exact line where `__sklearn_is_fitted` is called.

Issue #543, opened 2022-10-07
[FEATURE] `get_feature_names_out` for `sklego.preprocessing` transformers.


CarloLepelaars:
`get_feature_names_out` is an important component for interpreting scikit-learn `Pipeline` objects. A `get_feature_names_out` call on a `Pipeline` only works if it is implemented for all components in the pipeline, except the last step (i.e. the Model).

Scikit-learn recently implemented `get_feature_names_out` for all Transformers in their 1.1 release ([Source](https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_1_0.html?highlight=scikit+learn+release#get-feature-names-out-available-in-all-transformers)).

 I think it makes sense to also implement `get_feature_names_out` for all `scikit-lego` Transformers that are not models and are not `TrainOnly`. This leaves most objects in `sklego.preprocessing`.

- [ ] `sklego.preprocessing.ColumnCapper`
- [ ] `sklego.preprocessing.DictMapper`
- [ ] `sklego.preprocessing.IdentityTransformer`
- [ ] `sklego.preprocessing.IntervalEncoder`
- [x] ~~`sklego.preprocessing.OutlierRemover` (TrainOnly)~~
- [ ] `sklego.preprocessing.PandasTypeSelector`
- [ ] `sklego.preprocessing.ColumnSelector`
- [ ] `sklego.preprocessing.PatsyTransformer`
- [ ] `sklego.preprocessing.OrthogonalTransformer`
- [ ] `sklego.preprocessing.InformationFilter`
- [x] ~~`sklego.preprocessing.RandomAdder` (TrainOnly)~~
- [ ] `sklego.preprocessing.InformationFilter`

@koaning and I recently discussed implementing `get_feature_names_out` for `sklego.meta` and ended up implementing it for `EstimatorTransformer` (PR #539). It does not look like objects in `sklego.decomposition`, sklego.mixture` require an implementation of `get_feature_names_out`, because it seems there are mostly used as the last step in a pipeline or wrapped in an `EstimatorTransformer`. 

Since this is such a systemic issue, we can consider adding some additional requirements for people contributing to `sklego.preprocessing`. That is, make sure to implement `get_feature_names_out` for any new preprocessor that is not a train-time only Transformer.

- [ ] make sure that the feature you want is not already supported by sklearn

CarloLepelaars:
[...]

PR #544, opened 2022-10-11
[WIP] `get_feature_names_out` for `sklego.preprocessing`.  

CarloLepelaars:
This PR solves issue #543 and implements `get_feature_names_out` for all relevant transformers in `sklego.preprocessing` (i.e. transformers that do not contain the `TrainOnlyTransformerMixin`). 

Functionality is implemented through adding the `_ClassNamePrefixFeaturesOutMixin` to the class and making sure `self._n_features_out` is defined in `.fit`. This is also generally how `scikit-learn` implements `get_feature_names_out` for many of its transformers ([Example](https://github.com/scikit-learn/scikit-learn/blob/36958fb240fbe435673a9e3c52e769f01f36bec0/sklearn/preprocessing/_data.py#L2126)). Unit tests are added for all new functionality.

P.S. Don't pay attention to the commit history before October 10th. These changes have already been merged into `koaning/scikit-lego/main`, but is still displayed here as commit history. Will try to fix this. Suggestions to remove these redundant commits from the commit history of this `CarloLepelaars/scikit-lego/` fork are welcome.

- [x] ~~Find alternative solution for using [_ClassNamePrefixFeaturesOutMixin](https://github.com/scikit-learn/scikit-learn/blob/626b4608d4f840af7c37bff2ccb38fcfd2ef594f/sklearn/base.py#L868) so it works with scikit-learn on Python 3.7. (Remove Python 3.7. support)~~
- [x] Add implementation of `get_feature_names_out` to contributing guidelines so people implement this for each new preprocessor that is not TrainOnly.
- [x] Remove Python 3.7. GitHub Actions pipelines and update Optional dependencies GitHub Actions pipeline to use Python 3.8.
- [x] Add general unit test that checks if `get_feature_names_out` can be called for all relevant preprocessors and `EstimatorTransformer`.


koaning:
That's a big PR 😅 . 

Let us know when the tests are green and this PR is no longer WIP.

CarloLepelaars:
[_ClassNamePrefixFeaturesOutMixin](https://github.com/scikit-learn/scikit-learn/blob/626b4608d4f840af7c37bff2ccb38fcfd2ef594f/sklearn/base.py#L868) is not available for the `scikit-learn` version on Python 3.7. 😭 
Will find a different solution.

koaning:
@MBrouns for how long do we want to keep Python3.7 support?

MBrouns:
I think I'm okay with dropping 3.7 support. EOL in 8 months is soon enough and if it makes this feature easier to implement I'd say go for it

CarloLepelaars:
@koaning @MBrouns I added contributing guidelines to the README. Idea is that contributors of a new preprocessor will make sure to add `get_feature_names_out` functionality. We might want to consider creating a `CONTRIBUTING.MD` file [like HuggingFace transformer has](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md). 

Unfortunately it will take quite some work to get this working for Python 3.7. since in that case we don't have access to the convenient Mixins that implement `get_feature_names_out`. 

If you are fine with dropping Python 3.7. support then the PR seems ready to be merged.

koaning:
Feel free to remove 3.7 support. Maybe upgrade the github action that checks the optional deps.

koaning:
Is there a general unit test we might write that can check every component if it's implemented?

CarloLepelaars:
Don't think I have permissions to update the optional deps actions pipeline and remove Python 3.7. pipelines. Isn't that only available for maintainers?

> Is there a general unit test we might write that can check every component if it's implemented?

Good point! Will add a general unit test that checks if `get_feature_names_out` can be called for all relevant preprocessors (and `EstimatorTransformer`).

koaning:
You can change the files in .github in your PR right?

Calmcode has a github actions course that gives more details.

CarloLepelaars:
Aha of course, forgot I can adjust files in `.github`. Bumped minimum Python versions from 3.7. to 3.8 in `.github` and `.gitpod.yml`. Also added the general unit test.

@koaning The Github Actions pipeline still shows a `build (3.7.)` job. I'm don't know why this one still shows up.

koaning:
That's a setting that only I can change, but feel free to ignore that.

tests/test_preprocessing/test_interval_encoder.py
```diff
@@ -15,10 +15,6 @@
             general_checks,
             estimator_checks.check_estimators_dtypes,
             estimator_checks.check_fit_score_takes_y,
-            # estimator_checks.check_dtype_object,
-            # estimator_checks.check_sample_weights_pandas_series,
-            # estimator_checks.check_sample_weights_list,
-            # estimator_checks.check_sample_weights_invariance,
```
koaning, line 21: Please don't remove these. By leaving these comments here, we're conscious of the sklearn tests that we've manually removed.

readme.md
```diff
@@ -146,3 +148,11 @@ things before they become added to the project:
...
 
 We automate all of our testing and use pre-commit hooks to keep the code working.
+
+### Implementing a new preprocessor
+
+When creating a completely new transformer in `sklego.preprocessing`, make sure to implement `get_feature_names_out` functionality. 
+
+The preferred method is to add [_ClassNamePrefixFeaturesOutMixin](https://github.com/scikit-learn/scikit-learn/blob/626b4608d4f840af7c37bff2ccb38fcfd2ef594f/sklearn/base.py#L868) as a base class\
+and make sure `self._n_features_out` (i.e. the number of output features) is defined in the `fit` method of the new preprocessor.
+
```
koaning, line 158: @MBrouns I'm wondering. Would it be better to have folks import a hidden mixin, or just to have them implement `get_feature_names_out` themselves? Part of me is uneasy about relying on a non-public part of the sklearn api.

koaning: @CarloLepelaars also, seen [this](https://scikit-lego.netlify.app/contribution.html)?

tests/test_preprocessing/test_interval_encoder.py
CarloLepelaars: Ok, no problem. Added them back in. 

I also noticed some print statements in `test_patsy_transformer.py`. Are they also supposed to be there?

readme.md
MBrouns: interesting that this is part of the private API. Maybe copy the sklearn implementation into one of our own files then and use that instead for the time being?

CarloLepelaars: Fair point, the Mixin seems like the most clean implementation, but agree it is not a good idea to use Mixins from the non-public API. Would be interesting to ask an sklearn core maintainer about this decision. Not sure if they are aware that there are open-source projects like scikit-lego that are tightly built on sklearn. Seems like a good argument to make these kinds of Mixins part of the public API.

@MBrouns, sounds like a good intermediate step. Will put it in `common.py` with the `TrainOnlyTransformerMixin`.

koaning: > That already looks great! Perhaps we can add a link to here in the "New Features" section of the readme?

Sounds good.

koaning: Mhm, yeah, this is tricky territory.

If their internal class has many more links to internal objects it'll be even riskier to rely on it. 

Open to ideas though, there doesn't seem to be an "obvious" solution here.

CarloLepelaars:
Nice find, @MBrouns! I adjusted the general tests to use `check_transformer_get_feature_names_out` and `check_transformer_get_feature_names_out_pandas`.

Unfortunately the both break on the 1st line with ` TypeError: _get_tags() missing 1 required positional argument: 'self'`. Should every estimator have `_get_tags` implemented? If so, why doesn't it seem to be part of `BaseEstimator` or `TransformerMixin`?

CarloLepelaars:
@koaning As discussed we can keep this PR on hold for now and revisit later. The reason is that even when `_ClassNamePrefixFeaturesOutMixin` is made part of the public API, it would still restrict people to using the latest version of sklearn.

Some things that we should still consider: 
1. The PR for `EstimatorTransformer` #539 is done, tests are passing (also for Python 3.7) and it does not use the Mixin. Shall we go ahead and merge that one?
2. Do we still want to drop support for Python 3.7.? I can make a separate PR to adjust the Github Actions pipelines, test configs, etc.
3. Perhaps we can still implement `get_feature_names_out` for `sklego.preprocessing` ourselves? At least we already have the tests ready. If we implement it this way it is likely that the tests would still pass for Python 3.7.

FYI @MBrouns

PR #547, opened 2022-10-31
Selectors : allow results in empty dataframe

eromoe:
Before working on a large PR, please check with @koaning or @MBrouns that they agree with the direction of the PR. This discussion should take place in a Github issue before working on the PR, unless it's a minor change like spelling in the docs. 

# Description

Consider you want to build a semi-auto Pipeline. 
So, the pipeline may looks like: 

```python

import pandas as pd
import numpy as np
from sklearn.pipeline import Pipeline, FeatureUnion
from sklearn.preprocessing import OneHotEncoder, StandardScaler

transformer = Pipeline([
    ('features', FeatureUnion(n_jobs=1, transformer_list=[
        # Part 1
        ('boolean', Pipeline([
            ('selector', PandasTypeSelector(include='bool')),
        ])),  # booleans close
        
        ('numericals', Pipeline([
            ('selector', PandasTypeSelector(include='number')),
            ('scaler', StandardScaler()),
            ('add_pca', FeatureUnion([
                ('orig', IdentityTransformer()),
                ('pca', PCA(2))
            ]))
        ])),  # numericals close
        
        # Part 2
        ('categoricals', Pipeline([
            ('selector', PandasTypeSelector(include='category')),
            ('labeler', StringIndexer()),
            ('encoder', OneHotEncoder(handle_unknown='ignore')),
        ]))  # categoricals close
    
    ])),  # features close
])  # pipeline close
```

There may be boolean, numericals and categoricals variables , but may not exists erther . 
Current behaviour is raise Exception when a Selector return empty DataFrame , I think we can expose a parameter let user choose .

Fixes # (issue)

## Type of change
- [x ] New feature (non-breaking change which adds functionality)

# Checklist:

- [ ] My code follows the style guidelines (flake8)


MBrouns:
Interesting, I think I see how this could be useful for having 'default' pipelines that you can apply without needing to alter too much to your use case. I'm wondering though, how well do the other components (like the `StandardScaler` for example) handle datasets without any columns?

eromoe:
Oh, you are right !  I found the behaviour of sklearn built-in Transformer isn't consistent .

For `numerical` type, there does some transformers would raise error.

For `object` type , transformers accept empty dataframe, I have tested below without exception
- OneHotEncoder
- category_encoders.TargetEncoder

------

# First Let's consider the situations

## No NLP , datasets 

1. contains numericals in common
2. may contain boolean, string, and datetime 

## In typical NLP situation, datasets 

1. only have string in common
2. may contain others in addition

# Second, current mechanism

## manually check

1. We apply a boilerplate pipeline to an dataset 
2. Got the empty dataframe error 
3. Find out which transfomer raised , and remove that type

Not so efficient.

## code check

1. We check dataset columns type at first 
2. Use some python code to generate the pipeline 

You must put the logic out of pipeline 

----

I tend to build several baseline models for common usecases, and like to put mechanism in pipeline which at least looks better for me.
Do you have any suggesttion?

koaning:
It might be preferable to discuss this in an issue instead, but alas. 

Wouldn't it be somewhat more pragmatic to solve this using a function that can apply logic? Something like "only add this to the feature-union if this dtype is present?". That seems like the safest option, but it also seems like it's so specific that I might want to defer that responsibility to the end user.

FBruzzesi:
@koaning would it be reasonable to say that (almost 2 years after this PR) [`skrub.TableVectorizer`](https://skrub-data.org/stable/generated/skrub.TableVectorizer.html#skrub.TableVectorizer) leans towards this direction?

koaning:
The tablevectorizer is still being discussed internally so I want to be careful with making promises on how it will work, but yeah right now it works by checking the types going in and this will result in "empty" outputs sometimes.

FBruzzesi:
Thanks for the details! I am going to close this one as we are not planning to support it anyway for now

PR #550, opened 2022-11-04
CI versions

koaning:
This is a small cleanup of the Python versions in our CI. I also removed some old files that don't really do anything anymore.


koaning:
Good to know. @MBrouns numba doesn't support 3.11 yet. So we should delay upgrading for now.

PR #554, opened 2022-12-19
Fix a bug when using models that don't support sample weights

Garve:
# Description

Not all classifiers/regressors have a sample weight keyword, e.g. KNN. If you use one of these models, the ZIR will fail. This PR circumvents the problem by checking if the classifier/regressor has a `sample_weight` as a keyword and only then passes it on. If the classifier or regressor doesn't support sample weights, a warning is logged.

## Type of change

@koaning


Garve:
See https://github.com/koaning/scikit-lego/issues/555.

Garve:
Weird, after I merged your current main into my PR, it says 0 files were changed. Is it ok to remove this commit [1307b56](https://github.com/koaning/scikit-lego/pull/554/commits/1307b56f5dd68e8ada362d5db231e13982937c8f) again?

Garve:
Wow, I'm stupid, this PR got merged already months ago 😅 I thought I forgot it.

Issue #555, opened 2022-12-19
[BUG] Github Actions fail because boston housing was removed from scikit-learn 


Garve:
Hi!

Just created a PR and noticed that. Seems that we have to change the testing dataset or download it manually.

Best
Robert

koaning:
Ah, yeah. This was found to happen. 

The dataset is still hosted and fetch-able via the [fetch_openml](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_openml.html) function. Thanks for reporting!

PR #558, opened 2023-02-28
Load Boston is finally deprecated!

koaning:
I'm trying to fix the broken tests in CI while also reducing the warnings. 

Will also fix https://github.com/koaning/scikit-lego/issues/555.


koaning:
### Before 

```
=========== 7 failed, 3330 passed, 863 warnings in 198.02s (0:03:18) ===========
```

koaning:
### Halfway 

```
================ 3337 passed, 445 warnings in 223.14s (0:03:43) ================
```

PR #580, opened 2023-09-26
Relax `check_X_y` for `DecayEstimator`

### LINKED ISSUE #573: [BUG] Cannot use GroupedPredictor with DecayEstimator for fitting 1-feature dataframe
## Overview

Running the below works

```py
DecayEstimator(DummyRegressor(),decay=0.9).fit(df[['m']],df['yt'])
```

while this doesn't (expected to work)

```py
_ = (
    GroupedPredictor(DecayEstimator(DummyRegressor(), decay=0.9), groups=["m"])
    .fit(df[['m']], df['yt'])
)
```

While following Vincent's _excellent_ [sklearn crash course](https://www.youtube.com/watch?v=0B5eIE_1vpU) (See relevant screenshot below), I noticed that I wasn't able to use `GroupedPredictor` to fit with a single feature dataframe `df[['m']]` if `DecayEstimator` was used in conjunction. Is this the expected behaviour? I find it odd that calling `DecayEstimator(DummyRegressor(),decay=0.9).fit(df[['m']],df['yt'])` works, but failes when wrapped w `GroupedPredictor`.

As far as I understand, in this example the `DummyRegressor` just calcs the mean of `df['yt']` for the grouped period, so adding the `index` column here or any other cols don't seem to do anything (which looks to be the case when I tried adding a random col in place of `index`). I'm not 100% sure on this, so please correct me if I'm wrong!

Full error description and reproducable snippet below.

image

## Error description

The error: `ValueError: Found array with 0 feature(s) (shape=(1825, 0)) while a minimum of 1 is required by DecayEstimator.`

image

image

## Snippet producing error

Example snippet:

```py
from sklego.meta import GroupedPredictor, DecayEstimator
from sklego.datasets import make_simpleseries
from itertools import combinations

yt = make_simpleseries(seed=1)
dates = pd.date_range("2000-01-01", periods=len(yt))
df = (
    pd.DataFrame({"yt": yt,"date": dates})
    .assign(m=lambda d: d.date.dt.month)
    .reset_index()
)

df['a_useless_col'] = 0

# WORKS: Without wrapping the GroupedPredictor
_ = DecayEstimator(DummyRegressor(), decay=0.9).fit(df[['m']],df['yt']).predict(df[['m']])

# WORKS: Without the DecayEstimator
mod1 = (
    GroupedPredictor(DummyRegressor(), groups=["m"])
    .fit(df[['m']], df['yt'])
)

# WORKS: Adding a useless col when wrapping the GroupedPredictor
mod2 = (
    GroupedPredictor(DecayEstimator(DummyRegressor(), decay=0.9), groups=["m"])
    .fit(df[['m','a_useless_col']], df['yt'])
)

# WORKS: Adding the 'index' as shown in tutorial when wrapping the GroupedPredictor
mod3 = (
    GroupedPredictor(DecayEstimator(DummyRegressor(), decay=0.9), groups=["m"])
    .fit(df[['m','index']], df['yt'])
)

# ERROR: Wrapping the GroupedPredictor with DecayEstimator
mod4 = (
    GroupedPredictor(DecayEstimator(DummyRegressor(), decay=0.9), groups=["m"])
    .fit(df[['m']], df['yt'])
)
```

Additional columns does not seem to be taking any effect on the model output:

```py
# The output predictions of models 2,3 are the same, regardless of the added column in input 'X'
mod2_preds_m = mod2.predict(df[['m']])
mod2_preds_m_useless = mod2.predict(df[['m','a_useless_col']])
mod3_preds_m = mod3.predict(df[['m']])
mod3_preds_m_useless = mod3.predict(df[['m','index']])

array_equal_statements = []
for pair in combinations([
    mod2_preds_m,
    mod2_preds_m_useless,
    mod3_preds_m,
    mod3_preds_m_useless
], 2):
    array_equal_statements.append(np.array_equal(pair[0], pair[1]))

print(f"Are all {len(array_equal_statements)} arrays equal?: {all(array_equal_statements)}")
# OUPUTS: Are all 6 arrays equal?: True
```
## Further comments

From what I  can see in the error msg, it seems like it is reducing the dimensions so that it becomes an empty dataframe. I feel like this could be a result of dropping the `m` column somewhere for some reason when specifying `grouping = ['m']` in the `GroupedPredictor` params.

FBruzzesi:
# Description

Relaxes the check for `X` in `DecayEstimator` by delegating the actual checks to the given estimator. This fixes #573 and allows estimators that do not need `X` features. 
Remark, as discussed in the issue, it may be worth it to completely remove the check.

## Type of change

# Checklist:

### DIFF
--- a/sklego/meta/decay_estimator.py
+++ b/sklego/meta/decay_estimator.py
@@ -16,6 +16,9 @@ class DecayEstimator(BaseEstimator):
     This meta estimator will only work for estimators that have a
     "sample_weights" argument in their `.fit()` method.
 
+    The `fit` method computes the weights to pass to the estimator. All the checks are
+    delegated to the wrapped estimator.
+
     The DecayEstimator will use exponential decay to weight the parameters.
 
     w_{t-1} = decay * w_{t}
--- a/sklego/meta/decay_estimator.py
+++ b/sklego/meta/decay_estimator.py
@@ -18,9 +18,9 @@ class DecayEstimator(BaseEstimator):
 
     The `fit` method computes the weights to pass to the estimator.
 
-    .. warning:: By default all the checks on `X` and `y` are delegated to the wrapped estimator.
+    .. warning:: By default all the checks on the inputs `X` and `y` are delegated to the wrapped estimator.
 
-        To change such behaviour, set `check_X_y` to `True`.
+        To change such behaviour, set `check_input` to `True`.
 
         Remark that if the check is skipped, then `y` should have a `shape` attrbute, which is
         used to extract the number of samples in training data, and compute the weights.
@@ -31,12 +31,12 @@ class DecayEstimator(BaseEstimator):
     """
 
     def __init__(
-        self, model, decay: float = 0.999, decay_func="exponential", check_X_y=False
+        self, model, decay: float = 0.999, decay_func="exponential", check_input=False
     ):
         self.model = model
         self.decay = decay
         self.decay_func = decay_func
-        self.check_X_y = check_X_y
+        self.check_input = check_input
 
     def _is_classifier(self):
         return any(
@@ -57,7 +57,7 @@ class DecayEstimator(BaseEstimator):
         :return: Returns an instance of self.
         """
 
-        if self.check_X_y:
+        if self.check_input:
             X, y = check_X_y(
                 X, y, estimator=self, dtype=FLOAT_DTYPES, ensure_min_features=0
             )
--- a/sklego/meta/decay_estimator.py
+++ b/sklego/meta/decay_estimator.py
@@ -44,8 +44,10 @@ class DecayEstimator(BaseEstimator):
         :param y: array-like, shape=(n_samples,) training data.
         :return: Returns an instance of self.
         """
-        X, y = check_X_y(X, y, estimator=self, dtype=FLOAT_DTYPES)
-        self.weights_ = np.cumprod(np.ones(X.shape[0]) * self.decay)[::-1]
+        X, y = check_X_y(
+            X, y, estimator=self, dtype=FLOAT_DTYPES, ensure_min_features=0
+        )
+        self.weights_ = np.cumprod(np.ones(y.shape[0]) * self.decay)[::-1]
         self.estimator_ = clone(self.model)
         try:
             self.estimator_.fit(X, y, sample_weight=self.weights_)
--- a/sklego/meta/decay_estimator.py
+++ b/sklego/meta/decay_estimator.py
@@ -16,18 +16,27 @@ class DecayEstimator(BaseEstimator):
     This meta estimator will only work for estimators that have a
     "sample_weights" argument in their `.fit()` method.
 
-    The `fit` method computes the weights to pass to the estimator. All the checks are
-    delegated to the wrapped estimator.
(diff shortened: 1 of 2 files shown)


koaning:
I'm also OK with removing the entire check. Mainly because the internal method should already be doing that and this way we'll need to make fewer assumptions.

FBruzzesi:
> I'm also OK with removing the entire check. Mainly because the internal method should already be doing that and this way we'll need to make fewer assumptions.

To be completely pedantic, the `check_X_y` function call _checks_ and _converts_ X and y to arrays. This allows to then access `.shape` attribute to know how many samples are there and compute the decay.

Now both approaches feel wrong and with shortcomings:
- Using _a_ check, it may lead to errors or different results from the check in the estimator (e.g. if the estimator allows for multioutput while the check we are doing doesn't)
- Not having a check doesn't allow to access `.shape` or `len` in the general case (I noticed this by running unit tests) and I would be unsure how to proceed.

Opinions?

koaning:
> Not having a check doesn't allow to access .shape or len in the general case (I noticed this by running unit tests) and I would be unsure how to proceed.

Ah, I forgot about that. That's a great point. I guess the simplest solution is to offer a flag to the end-user and to defer the responsibility of applying the check to them? I think _most_ of the time `X` will be `.shape`-able or `len()`-able so the default could be `True`. Or am I missing something?

FBruzzesi:
> I guess the simplest solution is to offer a flag to the end-user and to defer the responsibility of applying the check to them? 

That's reasonable!

> That said, I think _most_ of the time `X` will be `.shape`-able or `len()`-able so the default could be `True`

With such default the scikit-learn tests would fail. I will add a big warning in the docstring 😁

koaning:
> With such default the scikit-learn tests would fail. I will add a big warning in the docstring 😁

There are a lot of components in this library that fail the standard checks. That's ok. It's not ideal, but given how "non-standard" some of these ideas are ... it's fine to turn a test off sometimes. Adding a warning to the docstring is fair though!

sklego/meta/decay_estimator.py
```diff
@@ -16,15 +16,27 @@ class DecayEstimator(BaseEstimator):
...
+        used to extract the number of samples in training data, and compute the weights.
+
     The DecayEstimator will use exponential decay to weight the parameters.
 
     w_{t-1} = decay * w_{t}
     """
 
-    def __init__(self, model, decay: float = 0.999, decay_func="exponential"):
+    def __init__(
+        self, model, decay: float = 0.999, decay_func="exponential", check_X_y=False
```
koaning, line 34: This is a bit of a nitpick. 

```suggestion
        self, model, decay: float = 0.999, decay_func="exponential", check_input=False
```

My reasoning here is that some estimators may only have a `check_X`. This is the case when the component is a scikit-learn transformer. But it'd be nice to have a consistent name for checking the internal thing. So maybe this namechange? 

Once that change is in/discussed I think we can hit merge :)

koaning (commented):
Left one comment related to the name of a variable

sklego/meta/decay_estimator.py
FBruzzesi: That's very reasonable 👌 let me adjust that!

koaning (approved):
LGTM!

PR #584, opened 2023-10-12
relax check_X_y on EstimatorTransformer

mtrbean:
# Description

Similar to #580, #507 ,  this change allows using an estimator / pipeline that can handle nan/string inputs with `EstimatorTransformer`

Fixes # (issue)

## Type of change

# Checklist:

- [x] My code follows the style guidelines (flake8)

### DIFF
--- a/sklego/meta/estimator_transformer.py
+++ b/sklego/meta/estimator_transformer.py
@@ -7,7 +7,7 @@ from sklearn.base import (
 from sklearn.utils.validation import (
     check_is_fitted,
     check_X_y,
-    FLOAT_DTYPES
+    FLOAT_DTYPES,
 )
 
 
--- a/sklego/meta/estimator_transformer.py
+++ b/sklego/meta/estimator_transformer.py
@@ -17,7 +17,7 @@ class EstimatorTransformer(TransformerMixin, MetaEstimatorMixin, BaseEstimator):
 
     :param estimator: An instance of the estimator that should be used for the transformation
     :param predict_func: The function called on the estimator when transforming e.g. (`predict`, `predict_proba`)
-    :param check_input: Whether to check the input data for NaNs, Infs and non-numeric values
+    :param check_input: Whether or not to check the input data. If False, the checks are delegated to the wrapped estimator.
     """
 
     def __init__(self, estimator, predict_func="predict", check_input=False):
--- a/sklego/meta/estimator_transformer.py
+++ b/sklego/meta/estimator_transformer.py
@@ -7,6 +7,7 @@ from sklearn.base import (
 from sklearn.utils.validation import (
     check_is_fitted,
     check_X_y,
+    FLOAT_DTYPES
 )
 
 
@@ -19,14 +20,16 @@ class EstimatorTransformer(TransformerMixin, MetaEstimatorMixin, BaseEstimator):
     :param check_input: Whether to check the input data for NaNs, Infs and non-numeric values
     """
 
-    def __init__(self, estimator, predict_func="predict", check_input=True):
+    def __init__(self, estimator, predict_func="predict", check_input=False):
         self.estimator = estimator
         self.predict_func = predict_func
         self.check_input = check_input
 
     def fit(self, X, y, **kwargs):
         """Fits the estimator"""
-        X, y = check_X_y(X, y, estimator=self, force_all_finite=self.check_input, dtype=None, multi_output=True)
+
+        if self.check_input:
+            X, y = check_X_y(X, y, estimator=self, dtype=FLOAT_DTYPES, multi_output=True)
 
         self.multi_output_ = len(y.shape) > 1
         self.estimator_ = clone(self.estimator)
--- a/sklego/meta/estimator_transformer.py
+++ b/sklego/meta/estimator_transformer.py
@@ -16,17 +16,17 @@ class EstimatorTransformer(TransformerMixin, MetaEstimatorMixin, BaseEstimator):
 
     :param estimator: An instance of the estimator that should be used for the transformation
     :param predict_func: The function called on the estimator when transforming e.g. (`predict`, `predict_proba`)
-    :param check_X: Whether to check the input data for NaNs, Infs and non-numeric values
+    :param check_input: Whether to check the input data for NaNs, Infs and non-numeric values
     """
 
-    def __init__(self, estimator, predict_func="predict", check_X=True):
+    def __init__(self, estimator, predict_func="predict", check_input=True):
         self.estimator = estimator
         self.predict_func = predict_func
-        self.check_X = check_X
+        self.check_input = check_input
 
     def fit(self, X, y, **kwargs):
         """Fits the estimator"""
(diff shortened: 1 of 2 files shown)


sklego/meta/estimator_transformer.py
```diff
@@ -17,15 +16,18 @@ class EstimatorTransformer(TransformerMixin, MetaEstimatorMixin, BaseEstimator):
 
     :param estimator: An instance of the estimator that should be used for the transformation
     :param predict_func: The function called on the estimator when transforming e.g. (`predict`, `predict_proba`)
+    :param check_X: Whether to check the input data for NaNs, Infs and non-numeric values
     """
 
-    def __init__(self, estimator, predict_func="predict"):
+    def __init__(self, estimator, predict_func="predict", check_X=True):
```
koaning, line 22: [This discussion](https://github.com/koaning/scikit-lego/pull/580#discussion_r1356711192) may be relevant. Once we agree on a direction this PR can follow the same path.

koaning (commented):
Thanks for the PR. It may be preferable to iron out https://github.com/koaning/scikit-lego/pull/580 before we resolve this one. The linked discussion should help explain.

sklego/meta/estimator_transformer.py
mtrbean: sounds good. thanks @koaning

tests/test_meta/test_estimatortransformer.py
```diff
@@ -111,3 +113,27 @@ def test_kwargs(patched_clone, random_xy_dataset_clf):
...
+def test_object_input():
+    """ Test X containing string with check_X=False. """
+    X = np.array([["1", 4], ["7", 3], ["5", 5], ["7", 2], ["5", 7]])
+    y = np.array([1, 0, 1, 0, 1])
+    clf = HistGradientBoostingClassifier()
+    transformer = EstimatorTransformer(clf, check_X=False)
+    transformed = transformer.fit(X, y).transform(X)
+
+    assert transformed.shape == (y.shape[0], 1)
+    assert np.all(transformed == clf.fit(X, y).predict(X))
```
koaning, line 139: It feels like these two tests are _very_ similar and might benefit from `pytest.parametrize`. If you appreciate an example, calmcode has a small tutorial. 

https://calmcode.io/pytest-tricks/parametrize-twice.html

You'd only need to parametrize once for this example, but this video has a nice example.

mtrbean: thanks! I made it a parametrized test

sklego/meta/estimator_transformer.py
mtrbean: done!

sklego/meta/estimator_transformer.py
```diff
@@ -17,15 +16,18 @@ class EstimatorTransformer(TransformerMixin, MetaEstimatorMixin, BaseEstimator):
...
-    def __init__(self, estimator, predict_func="predict"):
+    def __init__(self, estimator, predict_func="predict", check_input=True):
         self.estimator = estimator
         self.predict_func = predict_func
+        self.check_input = check_input
 
     def fit(self, X, y, **kwargs):
         """Fits the estimator"""
-        X, y = check_X_y(X, y, estimator=self, dtype=FLOAT_DTYPES, multi_output=True)
+        X, y = check_X_y(X, y, estimator=self, force_all_finite=self.check_input, dtype=None, multi_output=True)
```
FBruzzesi, line 29: Considering how #580 arise from #573, I personally would lean towards skipping the check entirely (and leaving it to the estimator itself) if `check_input=True` or leave it as it is otherwise. Similar to #580 the default behaviour could be `check_input=False` and delegate the check.

Here a similar example of #573 that would break if using `GroupedPredictor`:

```py
import pandas as pd

from sklego.meta import EstimatorTransformer, GroupedPredictor
from sklego.datasets import make_simpleseries

yt = make_simpleseries(seed=1)
dates = pd.date_range("2000-01-01", periods=len(yt))
df = (
    pd.DataFrame({"yt": yt,"date": dates})
    .assign(m=lambda d: d.date.dt.month)
    .reset_index()
)

# This works fine
_ = GroupedPredictor(DummyRegressor(), groups=["m"]).fit(df[['m']], df['yt']).predict(df[['m']])

# The following would break because GroupedPredictor requires a dataframe while the check would convert it to an array
_ = (EstimatorTransformer(GroupedPredictor(DummyRegressor(), groups=["m"]))
    .fit(df[['m', 'index']], df['yt'])
)
```
I understand that skipping _just_ a specific check is useful to your use case, but delegating it entirely feels like a more standardized solution for meta models

FBruzzesi (commented):
Left a comment with personal preference for the expected behaviour

sklego/meta/estimator_transformer.py
mtrbean: I generally agree with your point about skipping checking entirely for meta models, like what was done with the Thresholder in #507. In this case why do we even bother having a `check_input` param for `GroupedPredictor` / `EstimatorTransformer`?

Among other things, many times I wanted to pass a Pandas DataFrame directly into the underlying estimator (e.g. lightgbm detects categorical variables automatically if X is a pandas dataframe) but `check_array` or `check_X_y` would convert them into vanilla numpy arrays against my wish :/

FBruzzesi: @koaning should we conform all the _meta_ models to skip input check entirely, or at least all of them having a boolean flag to skip it? I wouldn't be against delegating everything to the model/estimator itself

koaning: > should we conform all the meta models to skip input check entirely, or at least all of them having a boolean flag to skip it?

I'm leaning towards adding a flag to all of them. The thing is that _in general_ it is also a good idea that this check is in there. It's just that the meta models pull some advanced/unconventional stuff which is why folks hit their head. I'm not 100% sure if the default should be true or false. Both have their merit. 

The honest story here is that we probably added the `check_X_y` in all these meta models to get some of the base sklearn checks to pass in CI. Not because a user made a request.

mtrbean: Sounds good, I've updated the PR accordingly. Hopefully this is good enough for merging. Thanks @FBruzzesi @koaning

FBruzzesi (commented):
LGTM
I left a tiny comment for the docstring describing the parameter, we should be ready to merge rightafter

koaning:
@FBruzzesi let me know on Slack once this is merged in so that I may prep a release.

mtrbean:
> LGTM I left a tiny comment for the docstring describing the parameter, we should be ready to merge rightafter

@FBruzzesi I don't seem to be able to find your comment - can you let me know what change you'd like to see?

FBruzzesi:
>  I don't seem to be able to find your comment

I can't find it either, not sure what went wrong. Apologies for the inconvenient

> can you let me know what change you'd like to see?

In the docstring for the `check_input` parameter, I would suggest something more generic as:

`:param check_input: Whether or not to check the input data. If False, the checks are delegated to the wrapped estimator.`

FBruzzesi (approved):
LGTM! Thank you 👏🏼

PR #586, opened 2023-10-19
Docstrings standardization

FBruzzesi:
# Description

This PR aims to refactor all the modules docstring in order to standardize them into numpy style. 
Therefore `numpydoc` dependency and extension has been added as well.

**Remark**: The target building tool is **mkdocs-materials** instead of sphinx, therefore some part will not render properly when tested with sphinx. If the decision is to stick with sphinx, then more changes are necessary.

To check how those would render, I added a minimal (?) **mkdocs.yaml** which reads **mkdocs/** folder containing just a page API rendering.

It can be tested by running the following command:
```console
python -m pip install \
    mkdocs==1.5.3 \
    mkdocs-autorefs==0.5.0 \
    mkdocs-material==9.4.5 \
    mkdocs-material-extensions==1.2 \
    mkdocstrings==0.23.0 \
    mkdocstrings-python==1.7.3 
mkdocs serve
```

**Additional information**

- I tried to keep the original docstrings as much as possible
- In some classes there are example usage but not nearly in the most of them. It would be good for future to add a simple example for each class
- All code changes are _solely_ due to black and isort with line length of 120

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -43,7 +43,8 @@ extensions = [
     'sphinx.ext.mathjax',
     'sphinx.ext.viewcode',
     'nbsphinx',
-    'recommonmark'
+    'recommonmark',
+    'numpydoc'
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/mkdocs/api/model-selection.md
+++ b/mkdocs/api/model-selection.md
@@ -2,15 +2,15 @@
 
 :::sklego.model_selection.GroupTimeSeriesSplit
     options:
-        show_root_full_path: false
+        show_root_full_path: true
         show_root_heading: true
 
 :::sklego.model_selection.KlusterFoldValidation
     options:
-        show_root_full_path: false
+        show_root_full_path: true
         show_root_heading: true
 
 :::sklego.model_selection.TimeGapSplit
     options:
-        show_root_full_path: false
+        show_root_full_path: true
         show_root_heading: true
--- a/mkdocs/api/model-selection.md
+++ b/mkdocs/api/model-selection.md
@@ -1,0 +1,16 @@
+# Model Selection
+
+:::sklego.model_selection.GroupTimeSeriesSplit
+    options:
+        show_root_full_path: false
+        show_root_heading: true
+
+:::sklego.model_selection.KlusterFoldValidation
+    options:
+        show_root_full_path: false
+        show_root_heading: true
+
+:::sklego.model_selection.TimeGapSplit
+    options:
+        show_root_full_path: false
+        show_root_heading: true
--- a/sklego/linear_model.py
+++ b/sklego/linear_model.py
@@ -16,24 +16,34 @@ from scipy.optimize import minimize
 from scipy.special._ufuncs import expit
 from sklearn.base import BaseEstimator, RegressorMixin
 from sklearn.linear_model._base import LinearClassifierMixin
-from sklearn.multiclass import OneVsRestClassifier, OneVsOneClassifier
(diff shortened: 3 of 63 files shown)


sklego/linear_model.py
```diff
@@ -121,18 +188,78 @@ def fit(self, X, y):
...
+    $$
+
+    Parameters
+    ----------
+    threshold : float, default=0.3
+        The threshold value for the dead zone effect.
+    relative : bool, default=False
+        If True, the threshold is relative to the target value. Namely the _dead zone effect_ is applied to the
+        relative error between the predicted and target values.
+    effect : Literal["linear", "quadratic"], default="linear"
```
FBruzzesi, line 236: I didn't documented `effect="constant"` because it breaks during `.fit()` (as `deadzone(errors)` would return None). 
I am not sure what the expected behaviour is: I would imagine the loss to be 0 within threshold and some constant value outside of it, but such value is not provided anywhere

mkdocs/api/model-selection.md
```diff
@@ -0,0 +1,16 @@
+# Model Selection
+
+:::sklego.model_selection.GroupTimeSeriesSplit
```
koaning, line 3: Spotted a table rendering issue. 

CleanShot 2023-10-30 at 17 04 32

mkdocs/api/model-selection.md
```diff
@@ -0,0 +1,16 @@
...
+    options:
+        show_root_full_path: false
```
koaning, line 5: CleanShot 2023-10-30 at 17 05 04

params also don't render super well on this one

doc/conf.py
```diff
@@ -43,7 +43,8 @@
     'sphinx.ext.mathjax',
```
koaning, line 43: I guess this is a general observation.

CleanShot 2023-10-30 at 17 05 57

Do we maybe also want to show the import path in these docs?

koaning (commented):
Great job! I only found one page with issues here. That's a ton of work! 

I left one comment about adding the import path. I don't know if it can be automated. But it might be nice to perhaps manually add on top?

mkdocs/api/model-selection.md
FBruzzesi: - For the table, the best solution I found is to make it in code as
image
  Since the table concept is slightly different in mkdocs.
- Thanks for the heads up on the parameters, the ":" at the end has been wrongly copy-pasted in a few places

doc/conf.py
FBruzzesi: It should be enough to change from:

```markdown
:::sklego.<module>.<class>
    options:
        show_root_full_path: false
        show_root_heading: true
```
to 
```markdown
:::sklego.<module>.<class>
    options:
        show_root_full_path: true
        show_root_heading: true
```
in order to render as:
image

However for some submodules it would be "partially wrong" as it would show `sklego.meta.confusion_balancer.ConfusionBalancer` while that can actually be imported as 
`from sklego.meta import ConfusionBalancer`. 

I think that keeping the full path is still better than none, yet the best solution could be to provide a (very tiny) example usage for each class.

koaning: Most of my other open source docs tend to indeed have usage examples partially for this reason. But I'm fine with keeping that open maybe for beginner issues.

koaning (approved):
I couldn't find any more issues from eyeballing this ... so I think it's good to merge :)

doc/conf.py
FBruzzesi: I will open a **(good first) issue** to keep track of it

PR #589, opened 2023-10-21
Docs via mkdocs

FBruzzesi:
# Description

This PR substitutes the sphinx docs in favor of mkdocs style via mkdocs-material and mkdocsrings-python (depends on #586).

In particular it migrates all the notebooks directly to markdown files (_almost_ one-to-one mapping).

## Render locally

To render the documentation locally, you can run the following command from the root of the repository:

```console
make docs
```

Then the documentation page will be available at localhost. 

## Remark

The majority of code displayed in the markdowns and code generate plots in the documentation is generated using the scripts in the `docs/_scripts` folder, and then accessed via the [pymdown snippets](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/) extension.

To generate the plots from scratch it is enough to run the following command from the root of the repository:

```console
cd docs
make generate-all
```

which will run all the scripts and save results in the `docs/_static` folder.

Notice that _not every image_ is code generated!


.gitignore
```diff
@@ -117,7 +117,6 @@ dmypy.json
 # editor
```
koaning, line 117: Just a nit, but the `[dev]` install should update to include `mkdocs` deps.

koaning: This is a cool command though. TIL. 

```
python -k mkdocs get-deps
```

.gitignore
```diff
@@ -117,7 +117,6 @@ dmypy.json
...
 .vscode
```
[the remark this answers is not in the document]
koaning: And I think we can move the Rstudio bit to a place where it's less front-and-center. Not that many folks will use it.

.gitignore
```diff
@@ -117,7 +117,6 @@ dmypy.json
...
 .idea
```
koaning, line 119: CleanShot 2023-10-30 at 07 07 55@2x

Isn't there a mkdoctrings setting to parse this?

koaning (commented):
Left some initial comments. In short: this looks great and thanks for the effort!

.gitignore
FBruzzesi: Yep I forgot to substitute that in both _setup.py_ and _docs-requirements.txt_

.gitignore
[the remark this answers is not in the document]
FBruzzesi: Should _RStudio_ queue on the user guide? At the end of the day it is a tutorial just as the others

.gitignore
FBruzzesi: Yes, you can try with mkdocs.yaml, line 61 `docstring_style: sphinx`. However notice that it would not entirely solve the problem as the docstrings follow a "mixed" style. E.g. `linear_model.BaseScipyMinimizeRegressor` is documented in numpy style.

#586 addresses this by standardizing all the docstring to numpydoc (and rendering with mkdocs)

docs/rstudio.md
```diff
@@ -74,64 +70,75 @@ X['colour'] <- as.numeric(X['colour'] == "Black")
...
+  scoring = list(accuracy=make_scorer(accuracy),
                  pp_score=pp_score('colour')),
   refit="accuracy")
 
 cv$fit(X, y)
 ```
 
-You'll note that we're using `list` and `as.integer` here. For details on why, 
-check out this [documentation page]().
+You'll note that we're using `list` and `as.integer` here. For details on why, check out this [documentation page]().
```
FBruzzesi, line 89: One question I had regarding this tutorial. Here there is a missing link that I am not sure what should refer to 🙃

koaning: This is probably fine: https://rstudio.github.io/reticulate/articles/calling_python.html#lists-tuples-and-dictionaries

koaning:
I just approved the docstrings PR. I think once you agree it's good to merge in we can sync it here and start work on getting this merged in. 👍

koaning:
@FBruzzesi could you fix the optional dependencies CI job? I think runing `mkdocs build` there instead of sphinx should be fine as a check.

.github/workflows/dependencies.yml
```diff
@@ -28,23 +28,23 @@ jobs:
     - name: Run Checks
       run: |
         python tests/scripts/check_pip.py missing cvxpy
-        python tests/scripts/check_pip.py contains scikit-learn
+        python tests/scripts/check_pip.py installed scikit-learn
```
FBruzzesi, line 31: Figured out that `contains` wasn't really checking anything 🙈

mkdocs.yaml
```diff
@@ -1,5 +1,7 @@
 # Project information
 site_name: scikit-lego
+# site_description: tbd
+# site_url: tbd
```
FBruzzesi, line 4: Left both to compile in the future

Makefile
```diff
@@ -34,7 +34,6 @@ clean:
 	rm -rf .pytest_cache
 	rm -rf build
 	rm -rf dist
-	rm -rf docs
```
FBruzzesi, line 37: Got it just in time before running the make command 😁

FBruzzesi (commented):
As a minor comment but very annoying behaviour, I am getting _a lot_ of warnings when building the docs locally.
Couldn't manage to find a fix for both image references and links for autorefers

koaning:
Yeah those warnings are annoying, but as long as there's a test in CI that confirms that it's still able to build then I think we can live with it. 

One detail, I pay for mkdocstrings via their insiders program so if I render the page I can get a big of extra sugar on the methods/classes like [here](https://koaning.github.io/sentence-models/). 

> Preview 

CleanShot 2023-11-22 at 20 34 15

koaning:
@FBruzzesi maybe as a final step, could we add a CI file like this one? 

```yaml
name: Confirm MKdocs can build

on: [push]

jobs:
  build:
    runs-on: ubuntu-latest
    strategy:
      max-parallel: 4
      matrix:
        python-version: [3.9]
        os: [ubuntu-latest]
    steps:
    - uses: actions/checkout@v1
    - name: setup ${{ matrix.python-version }}
      uses: actions/setup-python@v1
      with:
        python-version: ${{ matrix.python-version }}
    - name: install
      run: |
        make install
    - name: build docs
      run: python -m mkdocs build
```

Just to confirm that the docs can build on each PR?

.github/workflows/dependencies.yml
```diff
@@ -28,23 +28,23 @@ jobs:
...
     - name: Run Checks
       run: |
-        python tests/scripts/check_pip.py contains cvxpy scikit-learn
+        python tests/scripts/check_pip.py installed cvxpy formulaic patsy scikit-learn umap-learn
     - name: Docs can Build
       run: |
         sudo apt-get update && sudo apt-get install pandoc
         python -m pip install -e ".[docs]"
-        sphinx-build doc docs
+        mkdocs build
```
FBruzzesi, line 50: > as a final step, could we add a CI file like this one?

@koaning how's that different from the one we have here?

FBruzzesi:
> One detail, I pay for mkdocstrings via their insiders program so if I render the page I can get a big of extra sugar on the methods/classes like [here](https://koaning.github.io/sentence-models/).

That's a nice touch! Indeed you are also the only one to be able to deploy them in any case 😂

BTW the _meth_ for method is just hilarious

mkdocs.yaml
```diff
@@ -109,8 +113,28 @@ extra_javascript:
   - https://polyfill.io/v3/polyfill.min.js?features=es6
   - https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js
 
+validation:
+  absolute_links: ignore
+  unrecognized_links: ignore
```
FBruzzesi, line 118: Found these parameters to have waaaaaaaay less spamming when building the docs

.github/workflows/dependencies.yml
koaning: Ah! I somehow missed that. LGTM then :)

koaning (approved):

PR #594, opened 2023-11-06
[Feature] Decay functions

### LINKED ISSUE #581: [FEATURE] Decay functions
While working on the two tiny PRs #577 and #580 , I noticed that although `DecayEstimator` has a `decay_func` argument, the only implemented decay is exponential.

I have in mind some decay functions that we could support, however it _could_ break the API, which is something that nobody is looking forward to.

The idea is similar to how [shrinkage](https://github.com/koaning/scikit-lego/blob/main/sklego/meta/grouped_predictor.py#L54) works for `GroupedPredictor`:

```py
class DecayEstimator(BaseEstimator):
    def __init__(self, model, decay_func, **decay_kwargs):
        self.model = model
        self.decay_func = decay_func
        self.decay_kwargs = decay_kwargs
   
    def fit(self, X, y):
        ...
        sample_weight = self.decay_func(n_samples, **self.decay_kwargs)
        ...
```
where `decay_func` can be either a string (mapping to the "ready to use" decay functionalities) or a callable _that follows a given protocol_. By which I mean that, for example, the first argument should be the number of samples - which is enough to compute the exponential decay (and many others).

Here the list of decay functions I have implemented if we move forward with this (example with sample size of 500):
decays

Edit: In my mind, all of them can accept a `min_value` and a `max_value` which scale the shape of the decay (so that 0 is never actually reached)

FBruzzesi:
# Description

Introduces built-in decay functions as discussed in #581

## Type of change
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected): Fixes #581 

# Checklist:


sklego/meta/_decay_utils.py
```diff
@@ -0,0 +1,272 @@
...
+        - If `n_steps` less than 0 or greater than the number of samples.
+
+    TypeError
+        - If `n_steps` is not an integer.
+        - If `step_size` is not an integer.
+    """
+
+    def __init__(self, n_steps=None, step_size=None, min_value=0.0, max_value=1.0):
+        self.n_steps = n_steps
+        self.step_size = step_size
```
FBruzzesi, line 144: Not a huge fan of this implementation, looking for advice 😊

sklego/meta/_decay_utils.py
```diff
@@ -0,0 +1,272 @@
...
+        if self.min_value > self.max_value:
+            raise ValueError("`min_value` must be less than or equal to `max_value`")
+
+        if self.growth_rate is not None and (
+            self.growth_rate <= 0 or self.growth_rate >= 1
+        ):
+            raise ValueError("`growth_rate` must be between 0. and 1.")
+
+        n_samples = y.shape[0]
+        growth_rate = self.growth_rate or 10 / n_samples
```
FBruzzesi, line 264: This tend to be a good default in my experience

tests/test_meta/test_decay_estimator.py
```diff
@@ -16,43 +16,59 @@
 
 
 @pytest.mark.parametrize("test_fn", flatten([general_checks, regressor_checks]))
-def test_estimator_checks_regression(test_fn):
-    trf = DecayEstimator(LinearRegression(), check_input=True)
+@pytest.mark.parametrize("decay_func", ["exponential", "linear", "sigmoid"])
+def test_estimator_checks_regression(test_fn, decay_func):
+    trf = DecayEstimator(LinearRegression(), decay_func=decay_func, check_input=True)
     test_fn(DecayEstimator.__name__, trf)
```
FBruzzesi, line 22: I couldn't manage to change default values in ln22 and ln 29 as well.

sklego/meta/_decay_utils.py
```diff
@@ -0,0 +1,272 @@
+import numpy as np
```
FBruzzesi, line 1: Should we add check for `min_value` to be strictly positive? Estimators will handle that anyway, I would say it is redundant

sklego/meta/_decay_utils.py
FBruzzesi: Moved validation to `__init__`

koaning (commented):
I guess the main thing that's missing here is perhaps a docs page with plots of all of these decay functions?

FBruzzesi:
> I guess the main thing that's missing here is perhaps a docs page with plots of all of these decay functions?
> 
> I'll leave it up to @FBruzzesi to determine if it's best to add that later given the big push on the docs. But I'd imagine without those docs people may omit this new feature, which would be a shame.

In order we could:
- Merge doc PRs (whenever ready)
- Add proper user guide/documentation for this feature in this PR
- Finally merge this PR as well

koaning:
@FBruzzesi just merged the docs. Feel free to continue on this thread again. I'll try to figure out a nice docs deployment setup from my machine in the meantime.

FBruzzesi:
Ready for review!

- Switched from classes to functions to have a unified "api" when passing and calling a custom function
- Added section called **Decay Functions** in the **Meta Models** user guide
- Added API section for the functions as their signature may be useful to anyone who wants to tinker with values

koaning (approved):

Issue #307, opened 2020-03-16
[FEATURE] CV solution for anomaly detection without outliers during training


janvdvegt:
With anomaly detection, if you have labeled outliers there are two types of models. The first one requires the outliers, although regularly unlabeled. Isolation forests fall in this category. One class SVM specifically works better without the outliers in there. Properly evaluating this model does require the outliers though. The current sklearn setup does not allow for this case (I believe). It would be nice to have a way to do this easily. 

One possible approach would be to use a different type of validation iterator, that returns only negative sample indices in the training fold but both in the validation fold.

koaning:
Just to confirm. The proposal is to create a new CV method that accepts an outlier detector part of it's initialisation?

I certainly see some merit to this idea. Got an example of what the API might look like?

janvdvegt:
I think I mean something slightly different. There are outlier detection methods that only work when there are no outliers in the training data, in that sense, they are more like novelty detection. Of course, these outliers are very important for properly evaluating the hyperparameters and performance in general. So let's say we have `X` which contains our features and we have `y` that contains whether they are considered to be an outlier or not. `y` is important for evaluation but we don't use it during training. However, in these novelty algorithms we want to throw out the positive samples in the training set, so in our CV loop.

Let's say we have the following dataset:

```
X   y (anomalous)
0   0
1   1
2   0
3   0
4   0
5   1
6   0
```

If the first four samples are in the training split we want to remove sample `1` because we don't want outliers in our training set but in the validation split we do not want to remove `5` because we need it for proper evaluation.

With regard to a possible implementation, I'm not super familiar with the types of arguments available. I know that CV iterators return indices, so if `y` is available it could just be a CV iterator that filters out the positive indices in the training set but keeps them in the evaluation set.

koaning:
Just for confirmation. This is the situation?

```
X   y (anomalous)   y (to predict)  split
0   0               0               A
1   1               1000            A
2   0               2               A
3   0               3               B
4   0               4               B
5   1               10000           B
6   0               6               B
```

1. We first generate a split, this gives us `A, B`.
2. First `A` is the training set, we keep the outlier? We remove it in `B` before it is passed to another pipeline? 
3. Then `B` is the training set, we keep the outlier? We remove it in `A` before it is passed to another pipeline? 

[...]

PR #595, opened 2023-11-07
[Feature] `WithoutLiersCV` model selection

FBruzzesi:
# Description

Introduces `WithoutLiersCV` as discussed in #307. To be able to follow different cross validation _strategies_, the idea is to take a CV object as input and exclude the anomalous samples from the training indexes. All the splitting logic is delegated to the cv object.

## Type of change

# Checklist:

### DIFF
--- a/sklego/model_selection.py
+++ b/sklego/model_selection.py
@@ -643,3 +643,105 @@ class GroupTimeSeriesSplit(_BaseKFold):
         # create a mapper to set every group to the right group_id
         mapper = dict(zip(df["index"], df["group"]))
         return np.vectorize(mapper.get)(groups)
+
+
+class WithoutLiersCV:
+    """A custom cross-validation splitter that filters out data points labeled as anomalies during the splitting process
+    to exclude them from the training sets.
+
+    The anomaly label is specified by the `anomalous_label` parameter. Data points with this label will be excluded
+    during the splitting process from the training sets, but will be included in the test sets.
+
+    Parameters
+    ----------
+    cv : CV Splitter instance
+        The base cross-validation splitter used to split the data. It must have a `split` method that returns
+        a generator of (train_index, test_index) tuples.
+    anomalous_label : int, default=-1
+        The label used to identify anomalous data points in the target labels.
+        Data points with this label will be excluded during the splitting process from the training sets.
+
+    Example
+    -------
+    ```py
+    import numpy as np
+    from sklearn.model_selection import KFold
+    from sklego.model_selection import WithoutLiersCV
+
+    np.random.seed(1)
+    X = np.random.randn(100, 3)
+    y = np.random.randn(100) > 1.5  # 7% of the data is labeled as anomalous
+
+    cv = WithoutLiersCV(
+        cv=KFold(n_splits=3),
+        anomalous_label=1
+    )
+
+    for train_index, test_index in cv.split(X, y):
+        print(f"Train samples: {len(train_index)}", f"Test samples: {len(test_index)}", sep="\n")
+
+    '''
+    Train samples: 62
+    Test samples: 34
+    Train samples: 60
+    Test samples: 33
+    Train samples: 64
+    Test samples: 33
+    '''
+    ```
+
+    Note
+    ----
+    The `WithoutLiersCV` class is designed to work in conjunction with standard cross-validation techniques and
+    evaluate anomaly detection estimators that do not work well with outliers in the training dataset.
+    Such class of estimators are typically trained on inliers only, and therefore, the training set should not
+    contain any outliers
+    (cf. [Novelty Detection](https://scikit-learn.org/stable/modules/outlier_detection.html#novelty-detection)).
+    """
+
+    def __init__(self, cv, anomalous_label=-1):
+        self.cv = cv
+        self.anomalous_label = anomalous_label
+
+    def split(self, X, y, groups=None):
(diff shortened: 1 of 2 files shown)


sklego/model_selection.py
```diff
@@ -643,3 +643,105 @@ def _regroup(self, groups):
...
+    from sklego.model_selection import WithoutLiersCV
+
+    np.random.seed(1)
+    X = np.random.randn(100, 3)
+    y = np.random.randn(100) > 1.5  # 7% of the data is labeled as anomalous
+
+    cv = WithoutLiersCV(
+        cv=KFold(n_splits=3),
+        anomalous_label=1
+    )
```
koaning, line 678: I think I'd want @MBrouns to weight in on the name 😅 just to make sure. 

But I'm also wondering if it's perhaps easier to the enduser to not require an anomalous label ... wouldn't it perhaps be better to pass in an outlier model? this outlier model could then internally train on X and determine which items are outliers. Or am I overthinking?

FBruzzesi: From the conversation in the issue my understanding is slightly different. The goal of the CV is to validate anomaly detectors that do not train with different labels, namely the novelty detection ones. Therefore passing a novelty detection model would not be possible in the first place.

Now I agree that the name would suit both implementations 😁

FBruzzesi: @koaning Potentially we could have two CV strategies:
- `WithoutLiersCV`: takes _any_ outlier detection model, train on `X`, and excludes outliers from `train_indexes`
- `NoveltyDetectorCV`: what's in this PR, to be able to train a novelty detection algorithm on non-anomalous labels and evaluate on both anomalous and not.

koaning:
@FBruzzesi it's been a few years, is this something you still want in? Is this something you'd argue that you have a practical use-case for?

FBruzzesi:
Hey @koaning 👋🏼 

> is this something you still want in?

I don't have a strong opinion to be honest. I am ok if we close this

> Is this something you'd argue that you have a practical use-case for?

_Actually_ I now have a use case for it: as I moved to my current company and we do a bunch of anomaly detection, there are definitely cases in which I check the difference in training with and without anomalous points.

koaning:
If you have a use-case for it, why not add it in?

FBruzzesi:
> If you have a use-case for it, why not add it in?

I got the wrong impression that you just wanted to close this 😅

---

Let me rework this a tiny bit as it's been ~3 years and I learned a few things in the meanwhile. First thing I notice is that the class is not inheriting from anything in scikit, let me double check that

koaning:
I am cool with adding it, just to be clear, but on reflecting I think the state right now is more of a thing of "hey what's the state of the library?". It could be that we take it more in maintenance mode or that we say "nah we should celebrate creativity of components" now that we can have agents help out. 

Either way is totally fine.

FBruzzesi:
> I am cool with adding it, just to be clear, but on reflecting I think the state right now is more of a thing of "hey what's the state of the library?". It could be that we take it more in maintenance mode or that we say "nah we should celebrate creativity of components" now that we can have agents help out.
> 
> Either way is totally fine.

It's true that there has not been much active development, and we are acting in maintenance mode. That said, I would like to keep celebrating the original philosophy of creative freedom and "low stake" experimental components.

koaning:
We always celebrated creative freedom, but we also celebrated the constraint of "should be genuinely useful". As long as both apply, I would love to see more :)

FBruzzesi:
Fully agree Vincent 👌🏼

Anyway, I added a commit (https://github.com/koaning/scikit-lego/pull/595/commits/cb4bd1efb229308474b60e183f311ffbeac3cc93) and PR is ready for review

sklego/model_selection.py
```diff
@@ -902,3 +905,110 @@ def _regroup(self, groups):
         # create a mapper to set every group to the right group_id
         mapper = dict(zip(df["index"], df["group"]))
         return np.vectorize(mapper.get)(groups)
+
+
+class WithoutLiersCV(BaseCrossValidator):
+    r"""Cross-validation splitter that filters out data points labeled as anomalies during the splitting process
+    to exclude them from the training sets.
```
koaning, line 912: Do all the outliers appear in the test sets now? I think they don't but it could be good to clarify explicitly.

sklego/model_selection.py
```diff
@@ -902,3 +905,110 @@ def _regroup(self, groups):
...
+        Yields
+        ------
+        train_index : np.ndarray
+            Indices of the training set, excluding anomalous data points.
+        test_index : np.ndarray
+            Indices of the test set, including anomalous data points.
+        """
+        y_arr = np.asarray(y)
+        for train_index, test_index in self.cv.split(X, y, groups):
+            inlier_index = np.where(y_arr[train_index] != self.anomalous_label)[0]
```
koaning: Does it make sense to have a setting here to also exclude it from the test set?

docs/api/model-selection.md
```diff
@@ -15,6 +15,11 @@
         show_root_full_path: true
         show_root_heading: true
 
+:::sklego.model_selection.WithoutLiersCV
```
koaning, line 18: Maybe we should add a small segment to explain when you'd want this new component? What specific problem it solves. We don't need to have a full dataset/example here but something that makes it feel tangible/plausible of when to use could help people toy with it.

koaning (changes requested):
mostly some small doc suggestions

PR #602, opened 2023-12-13
Relative docs paths

koaning:
I think this fixes most of the non-rendering images on Github Pages.


FBruzzesi (approved):
Added a few more that I found in fairness and meta user guides.

It should be good to go 🤞🏼

Issue #601, opened 2023-12-12
[FEATURE] Switch to `ruff`


koaning:
Might also be good to see what's up with the CI. It seems like the style checks are skipped?

PR #603, opened 2023-12-22
Ruff, Github Actions and other refactoring

FBruzzesi:
# Description

This is another large PR I guess, here the list of changes:

- Migrate from setup.py to pyproject.toml
  - keeping setup.py just to distribute using the same process but could even be removed
- Move away from black and flake8 to adopt ruff both as linter and formatter (solve #601)
- Cleanup some github actions/workflows
- Every `sklego/*.py` and `tests/*.py` edit is due to ruff
- Only exception being `sklego/__init__.py` 

# Checklist:


.github/dependabot.yaml
```diff
@@ -0,0 +1,7 @@
+version: 2
```
FBruzzesi, line 1: Periodically checks for github action updates - are we interested in this?

.github/workflows/test.yml
```diff
@@ -0,0 +1,37 @@
...
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
FBruzzesi, line 13: Moved all os into a single workflow

.github/workflows/test.yml
```diff
@@ -0,0 +1,37 @@
...
+        python-version: [
+          # "3.6", Unsupported on ubuntu-latest (22.04)
+          # "3.7", Long series of scikit-learn failures
+          "3.8",
+          "3.9",
+          "3.10",
+          "3.11",
+          # "3.12", # Numba not available for Python 3.12: https://github.com/numba/numba/issues/9197
```
FBruzzesi, line 21: Non-working version are commented with their reason for breaking

pyproject.toml
```diff
@@ -0,0 +1,103 @@
...
+    "scikit-learn>=1.0",
+    "importlib-metadata >= 1.0; python_version < '3.8'",
+]
+
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
FBruzzesi, line 34: Tagged only the version for which workflows run without errors

sklego/__init__.py
```diff
@@ -1 +1,9 @@
-__version__ = "0.7.0"
+import sys
+
+if sys.version_info >= (3, 8):
+    from importlib import metadata
+else:
+    import importlib_metadata as metadata
+
+__title__ = "sklego"
+__version__ = metadata.version("scikit-lego")
```
FBruzzesi, line 9: This took me sooo long to debug due to the different naming between distribution and module

sklego/decomposition/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .pca_reconstruction import PCAOutlierDetection
-from .umap_reconstruction import UMAPOutlierDetection
+from sklego.decomposition.pca_reconstruction import PCAOutlierDetection
+from sklego.decomposition.umap_reconstruction import UMAPOutlierDetection
```
FBruzzesi, line 2: I made all imports absolute, maybe unnecessary

pyproject.toml
```diff
@@ -0,0 +1,103 @@
...
+license = {file = "LICENSE"}
+readme = "readme.md"
+requires-python = ">=3.6"
+authors = [
+    {name = "Vincent D. Warmerdam"},
+    {name = "Matthijs Brouns"},
+]
+
+maintainers = [
+    {name = "Francesco Bruzzesi"}
```
FBruzzesi, line 19: Kinda sneak myself in there 😁

.github/workflows/test.yml
koaning: It think it's fine to remove both 3.6 and 3.7 at this point though. Both have reached end of life officially.

.github/dependabot.yaml
koaning: I'm willing to try it. But I'm in favor of removing it the moment it starts producing noise.

pyproject.toml
koaning: Rightly so!

pyproject.toml
```diff
@@ -0,0 +1,103 @@
...
+    "matplotlib>=3.0.2",
+    "jupyter>=1.0.0",
+    "jupyterlab>=0.35.4",
+]
+
+dev = [
+    "scikit-lego[all,test,docs]",
+    "pre-commit>=1.18.3",
+    "ruff>=0.1.6",
+    "hatch",
```
koaning, line 81: I might be missing something, but where are we using hatch? I don't think I mind, but we should add a minimum version I guess.

koaning (commented):
I left some minor comments and a question about hatch. These feel relatively minor though so I'm assuming I'll just merge it once the answers ping back. 

Good work though, this feels like a pretty big cleanup!

sklego/__init__.py
koaning: Yeah, that's certainly a bit awkward.

pyproject.toml
FBruzzesi: Oh nice catch! I was trying to completely move away from using setup.py using another building tool. For some reason I couldn't fully manage to do so, but forgot to remove `hatch`.

I will fix it as soon as I log into a laptop 😁

.github/dependabot.yaml
FBruzzesi: In my experience it shouldn't be too noisy as we are not using very fast changing github actions

koaning (approved):

PR #604, opened 2023-12-28
WIP: unittests speedup

FBruzzesi:
# Description

I was playing around for some features and while iterating I was getting bored of waiting for tests.
Thus running `pytest . --durations 30 --disable-warnings` I could get the most time consuming tests.

The code changes keep the same level of coverage but lower the runtime.

I opened a [PR in my own fork](https://github.com/FBruzzesi/scikit-lego/actions/runs/7349185428) to quantify the difference and it is quite nice (4m 14s vs last merge here of 7m 37s). 

I didn't investigate much further, but since it was a bit relevant for me I wanted to share it.

### DIFF
--- a/.github/workflows/schedule-dependencies.yml
+++ b/.github/workflows/schedule-dependencies.yml
@@ -28,4 +28,4 @@ jobs:
     - name: Test with pytest
       run: |
         pip install -e ".[test]"
-        pytest -n 2 tests
+        make test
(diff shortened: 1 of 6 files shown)


koaning (approved):
Hah! Interesting, I remember getting to a similar conclusion a few years ago when I tried making these tests faster. I wrote [pytest duration insights](https://github.com/koaning/pytest-duration-insights) for a use-case at Rasa which I then ran on this codebase. But yeah, making UMAP tests a bunch faster is an easy accept in my book.

FBruzzesi:
Love to hear that! 
To be fair, the best improvement comes from not generating the splits in the grouped time series cv, but running its "raising warning" method as standalone

PR #605, opened 2023-12-29
Un-pandas `GroupedTimeSeriesSplit`

FBruzzesi:
# Description

In #604 there was a tiny hint at this 😂 

In this PR I tried to remove the use of pandas from `GroupedTimeSeriesSplit` by moving on numpy backend. The only method that still needs a "dataframe"-like object is `.summary()` (which it's even a nice to have).

In a way this is an alternative way to achieve #597 when a dataframe is just a nice abstraction but not mandatory. The only function that has to work to proceed forward and run `.split()` method is `indexable` (from sklearn.utils.validation).

### DIFF
--- a/sklego/model_selection.py
+++ b/sklego/model_selection.py
@@ -131,7 +131,7 @@ class TimeGapSplit:
 
         if len(X) != len(X_index_df):
             raise AssertionError(
-                "X and X_index_df are not the same length, " "there must be some index missing in 'self.date_serie'"
+                "X and X_index_df are not the same length, there must be some index missing in 'self.date_serie'"
             )
 
         date_min = X_index_df["__date__"].min()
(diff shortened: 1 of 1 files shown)


sklego/model_selection.py
```diff
@@ -485,8 +472,10 @@ def _iter_test_indices(self, X=None, y=None, groups=None):
         self._first_split_index, self._last_split_index = self._calc_first_and_last_split_index(groups=groups)
         self._best_splits = self._get_split_indices()
         groups = self._regroup(groups)
+        group_indices = tuple(np.where(groups == i)[0] for i in range(self.n_splits + 1))
+
         for i in range(self.n_splits):
-            yield np.where(groups == i)[0], np.where(groups == i + 1)[0]
+            yield group_indices[i], group_indices[i + 1]
 
     def _calc_first_and_last_split_index(self, X=None, y=None, groups=None):
```
FBruzzesi, line 480: In a way, the changes in `_calc_first_and_last_split_index` are the only hard core difference

FBruzzesi:
Following all the effort put into adopting Narwhals, it would be nice to revamp this as well. I will close the PR and come back to the topic!

PR #606, opened 2023-12-31
[DOCS] Fix relative paths

FBruzzesi:
# Description

In the deployed docs in github pages, all relative links are broken. I would imagine that the issue is similar to what happened in #602 .

This PR tries to patch that. 

Most annoying fact is that locally both strategies work.

Edit: Adding screenshot as example:

Clicking the link at https://koaning.github.io/scikit-lego/#usage to open the user guide

image

maps to https://koaning.github.io/user-guide/datasets (which is 404) instead of https://koaning.github.io/scikit-lego/user-guide/datasets/


readme.md
```diff
@@ -1,16 +1,16 @@
 [![Build status](https://github.com/koaning/scikit-lego/workflows/Unit%20Tests/badge.svg)](https://github.com/{github_id}/{repository}/workflows/{workflow_name}/badge.svg)
```
FBruzzesi, line 1: Not sure if `Build status` is still relevant, I couldn't find a passing status though

PR #609, opened 2024-01-19
Fix slep006 issue in CI cronjob

koaning:
If we want to use sklearns unit tests then we should just import them directly because this ensures that all the required fixtures are around. Opening a file and checking for global variables is a dirty way of going about it so I figured it best to delete these tests for now. CI is complaining that it cannot find a fixture and it's getting to the point where I'm ignoring the daily email, which is a habbit that needs to go away.


koaning:
@FBruzzesi unless you have strong opinions, I'd prefer to just remove these tests for now. We have plenty of other tests that are better checks than the stuff in this file.

FBruzzesi (approved):
Sounds good to me. I admit I adopted the same habit of just ignoring such email first time in the morning 🙈

koaning:
> Sounds good to me. I admit I adopted the same habit of just ignoring such email first time in the morning 🙈

Yeah, and _that_ is a dangerous thing. These emails should be taken serious and once they are ignored they stop providing utility.

PR #611, opened 2024-01-22
[Feature]: Meta `OrdinalClassifier` estimator

### LINKED ISSUE #607: [FEATURE] Meta Ordinal Classification
# Description
 
So I was looking for implementation(s) of ordinal classification in python (possibly scikit-learn compatible).

There exists a library called [mord](https://pythonhosted.org/mord/) that implements a few strategies. And then I stumbled upon [a simple approach to ordinal classification](https://link.springer.com/chapter/10.1007/3-540-44795-4_13) (*) which is also discussed in a "needs decision" [scikit-learn issue](https://github.com/scikit-learn/scikit-learn/issues/23324).

This feature is not currently supported by scikit-learn, but it will potentially be in the future. I am not sure what the stance is towards this kind of situations. Considering the different speed at which we can move it would be a nice addition in my opinion.

## (*) The paper

The idea of the paper is quite simple (and _very_ meta): turn a ordinal classification problem with K classes, into K-1 binary problems. 
Example in image:

image

FBruzzesi:
# Description

Fix #607 by introducing `OrdinalClassifier` (meta) estimator.

## Checklist

- [X] Add docstrings
- [x] Add documentation section
- [x] Add unit tests


sklego/meta/ordinal_classification.py
```diff
@@ -0,0 +1,185 @@
...
+        ------
+        ValueError
+            If `X` has a different number of features than the one seen during `fit`.
+        """
+        check_is_fitted(self, ["estimators_", "classes_"])
+
+        if X.shape[1] != self.n_features_in_:
+            raise ValueError(f"X has {X.shape[1]} features, expected {self.n_features_in_} features.")
+
+        raw_proba = np.array([estimator.predict_proba(X)[:, 1] for estimator in self.estimators_.values()]).T
```
FBruzzesi, line 153: Some food for thought: I also wonder if we want to expose a method that returns these probabilities directly

sklego/meta/ordinal_classification.py
```diff
@@ -0,0 +1,185 @@
...
+
+        Returns
+        -------
+        fitted_model : scikit-learn compatible classifier
+            The fitted binary classifier.
+        """
+        y_bin = (y <= y_label).astype(int)
+        fitted_model = clone(self.estimator).fit(X, y_bin)
+        if self.use_calibration:
+            return CalibratedClassifierCV(fitted_model, cv="prefit").fit(X, y_bin)
```
FBruzzesi, line 183: This is something I would rather have, but open to discuss if actually needed (and eventually which default to have) or only suggested as good practice for the user. I added a _warning_ section in the class docstring to explain why

koaning: Wouldn't the sklearn API prefer this:

```
pipeline = CalibratedClassifierCV(OrdinalClassifier(estimator))
```

It feels more inline with the API to have users do this themselves, no?

sklego/meta/ordinal_classification.py
koaning: Makes sense to me!

tmp/ordinal_classification_demo.ipynb
```diff
@@ -2,16 +2,16 @@
  "cells": [
```
koaning, line 2: I think we'd want to remove this file?

tests/test_meta/test_ordinal_classification.py
```diff
@@ -0,0 +1,51 @@
...
+    if exc_info:
+        assert err_msg in str(exc_info.value)
+
+
+@pytest.mark.parametrize("n_jobs", [-2, -1, 2, None])
+@pytest.mark.parametrize("use_calibration", [True, False])
+def test_can_fit_param_combination(random_xy_ordinal, n_jobs, use_calibration):
+    X, y = random_xy_ordinal
+    ord_clf = OrdinalClassifier(estimator=LogisticRegression(), n_jobs=n_jobs, use_calibration=use_calibration)
+    _ = ord_clf.fit(X, y)
```
koaning, line 46: Nitpick, but we don't need to store this in `_`, right? 
```suggestion
    ord_clf.fit(X, y)
```

docs/user-guide/meta-models.md
```diff
@@ -381,6 +381,62 @@ The `OutlierClassifier` can be combined with any classification model in the `St
 
 --8<-- "docs/_static/meta-models/outlier-classifier-stacking.html"
 
+## Ordinal Classification
+
+Ordinal classification (sometimes also referred to as Ordinal Refression) involves predicting an ordinal target variable, where the classes have a meaningful order.
```
koaning, line 386:
```suggestion
Ordinal classification (sometimes also referred to as Ordinal Regression) involves predicting an ordinal target variable, where the classes have a meaningful order.
```

docs/user-guide/meta-models.md
```diff
@@ -381,6 +381,62 @@ The `OutlierClassifier` can be combined with any classification model in the `St
...
+Examples of this kind of problem are: predicting customer satisfaction on a scale from 1 to 5, predicting the severity of a disease, predicting the quality of a product, etc.
+
+The [`OrdinalClassifier`][ordinal-classifier-api] is a meta-model that can be used to transform any classifier into an ordinal classifier by fitting N-1 binary classifiers, each handling a specific class boundary, namely: $P(y <= 1), P(y <= 2), ..., P(y <= N-1)$.
+
```
koaning, line 390: Bit of a nit, but might it make sense to add this diagram ...

CleanShot 2024-01-26 at 15 57 39

... somewhere over here? ... 

CleanShot 2024-01-26 at 15 56 56

tmp/ordinal_classification_demo.ipynb
FBruzzesi: I kept it so that you could see the results on monotonicity. I am going to remove it as soon as we confirm the rest

sklego/meta/ordinal_classification.py
```diff
@@ -178,8 +205,29 @@ def _fit_binary_estimator(self, X, y, y_label):
...
+            The training data.
+        y : array-like of shape (n_samples,)
+            The target values.
+
+        Returns
+        -------
+        score : float
+            Accuracy score of self.predict(X) wrt. y.
+        """
+        return accuracy_score(y, self.predict(X))
```
koaning, line 228: Doesn't the `ClassifierMixin` provide this already? If we want to do something custom here, perhaps it's better to re-use the `score()` method of the underlying classifier?

koaning (changes requested):
Nice work! It feels like this is about 90% done, but I did find some things to address. Relatively minor things though!

docs/user-guide/meta-models.md
FBruzzesi: Makes 110% sense! Collapsed admonition should do the trick

sklego/meta/ordinal_classification.py
```diff
@@ -0,0 +1,233 @@
...
+
+        if not hasattr(self.estimator, "predict_proba"):
+            raise ValueError("The estimator must implement `.predict_proba()` method.")
+
+        X, y = check_X_y(X, y, estimator=self)
+
+        self.classes_ = np.sort(np.unique(y))
+        self.n_features_in_ = X.shape[1]
+
+        if self.n_classes_ < 2:
```
koaning, line 136: I wonder ... don't we need at least three classes for ordinal regression to make sense? Any binary classification with two classes is automatically ordinal, no?

sklego/meta/ordinal_classification.py
FBruzzesi: Oh you are right! I am removing it for now 😊
Not sure what something custom and default would be.

I am personally not paying much attention to accuracy in the ordinal problem I am dealing with, but don't want to put anything very opinionated, so let's go with the hierarchical implementation from `ClassifierMixin`

sklego/meta/ordinal_classification.py
FBruzzesi: Also very true! I cheated for the scikit-learn tests 😂 
Let me figure out how to deal with those

tests/test_meta/test_ordinal_classification.py
```diff
@@ -0,0 +1,63 @@
...
+@pytest.mark.parametrize(
+    "test_fn",
+    select_tests(
+        flatten([general_checks, classifier_checks]),
+        exclude=[
+            "check_dict_unchanged",
+            "check_fit2d_1feature",
+            "check_classifier_data_not_an_array",
+            "check_classifiers_classes",
+            "check_classifiers_train",
```
FBruzzesi, line 27: This is a fair amount of skipping, yet:
- coverage is still same
- the internal estimator will most likely deal directly with these cases (e.g. data not array)

koaning: Yeah, seems fair.

koaning (approved):

Issue #616, opened 2024-02-07
[BUG] Error when calling predict_proba with GroupedPredictor using shrinkage and global model


cerlymarco:
When trying to generate probabilities calling `predict_proba` from a fitted instance of `GroupedPredictor` with `shrinkage` not None and `use_global_model=True` the following error is generated:

<img width="582" alt="image" src="https://github.com/koaning/scikit-lego/assets/36955807/9db77ec8-17f2-449e-8092-835cbfc32ea6">

Code to reproduce the error:

<img width="358" alt="image" src="https://github.com/koaning/scikit-lego/assets/36955807/373d12bf-ce45-4fe8-af48-c201a6c9e33e">

```
import sklearn
import sklego
from sklego.datasets import load_hearts
from sklearn.linear_model import LogisticRegression
from sklego.meta import GroupedPredictor

## sklego.__version__   0.7.4
## sklearn.__version__  1.0.2

df_hearts = load_hearts(as_frame=True)
model = GroupedPredictor(
    LogisticRegression(max_iter=1000), groups=['sex'],
    use_global_model=True, shrinkage="relative",
).fit(df_hearts.drop(columns=['thal','target']), df_hearts['target'])

preds = model.predict_proba(df_hearts.drop(columns=['thal','target']))
```

FBruzzesi:
Hey @cerlymarco, thanks for reporting this bug. I can replicate the issue and will take a closer look.

My first impression is that shrinkage was intended for regression problems only, as running `.predict(...)` with the same data you provided will yield values that are not classes.

cerlymarco:
Hi @FBruzzesi, thanks for your reply! That seems reasonable. If that is the case, an error/warning may be raised when a classifier (an estimator with predict_proba method) is provided with shrinkage enabled. Otherwise, a proper shrinkage method for probabilities could be implemented

FBruzzesi:
Hey @cerlymarco , as you can see this issue was mentioned in a few places. Just to add some description to each of those:

- [`GroupedPredictor` refactoring](https://github.com/koaning/scikit-lego/pull/618) was a first attempt to fix the issue reported here, as well as expanding the functionalities. Since this was getting a bit messy we decided to follow another approach, which led us to the following two points.
- [`GroupedPredictor` patch](https://github.com/koaning/scikit-lego/pull/619) follows your suggestion and raises an error if shrinkage is used in combination with a classification task. This is a somewhat breaking change which is already in the `main` branch.
[...]

Issue #579, opened 2023-09-26
[BUG] GroupedPredictor inconsistency for predict_proba having different classes per group


fabioscantamburlo:
Hello scikit-lego users.
While using predict_proba paired with GroupedPredictor and a classifier on a df with different labels per group, the final matrix is collapsed to the left, without caring about the label order.
This yields inconsistencies in the final output, especially while using an high number of labels.
To achieve a sound result every label should appear at least once in every group, that is somehow unrealistic.

Here a snippet of code:
``` python
import pandas as pd
import numpy as np

from sklego.meta import GroupedPredictor
from sklearn.linear_model import LogisticRegression


np.random.seed(43)

group_size = 5
n_groups = 2
df = pd.DataFrame({
    "group": ["A"] * group_size + ["B"] * group_size,
    "x": np.random.normal(size=group_size * n_groups),
    "y": np.hstack([
        np.random.choice([0, 1, 2], size=group_size),
        np.random.choice([0, 2], size=group_size),
        ])
})

print(df.groupby('group').agg({'y': set}))


X, y = df[["x", "group"]], df["y"]
model = GroupedPredictor(LogisticRegression(), groups=["group"])
_ = model.fit(X, y)
y_prob = model.predict_proba(X)

print(y_prob.round(2))

```

Outputs:
``` python

>>>                y
>>> group           
>>> A      {0, 1, 2}
>>> B         {0, 2}

>>> [[0.45 0.19 0.36]#grp A
>>> [0.3  0.23 0.47]
>>> [0.37 0.21 0.42]
>>> [0.35 0.22 0.44]
>>> [0.53 0.16 0.31]
>>> [0.79 0.21  nan]#grp B
>>> [0.8  0.2   nan]
>>> [0.81 0.19  nan]
>>> [0.81 0.19  nan]
>>> [0.79 0.21  nan]]

# Expected:
>>> [[0.45 0.19 0.36]#grp A
>>> [0.3  0.23 0.47]
>>> [0.37 0.21 0.42]
>>> [0.35 0.22 0.44]
>>> [0.53 0.16 0.31]
>>> [0.79 nan  0.21]#grp B
>>> [0.8  nan   0.2]
>>> [0.81 nan  0.19]
>>> [0.81 nan  0.19]
>>> [0.79 nan  0.21]]

```

FBruzzesi:
Hey @fabioscantamburlo, thanks for reporting the bug. 
[...]

PR #618, opened 2024-02-13
`GroupedPredictor` refactoring

FBruzzesi:
# Description

This is a first attempt to refactor `GroupedPredictor` class (personally one of my favorite features 😁) following the issue raised in #616  .

While working on this I noticed another set of issues with the implementation:
- Following #579, decision function will yield wrong results. This is due to the following fact: if the two groups are different classification problems, namely binary and multiclass (as it happens in the unittests), the result of `.decision_function(...)` will be a 1d or 2d array (resp.), which when concatenated lead to wrong behaviour.
  - This PR falls short in fixing that, yet I have a couple of ideas that needs test/triage
- Consider the following example when multiple groups are present:  let `groups=["a", "b"]` with values `(0, 0)`, `(0, 1)` and `(1, 0)`. Currently if at prediction time we encounter `(a=0, b=2)` for which we don't have a trained model, we are falling back to the global one. However I would argue that we should fallback to the model trained on `a=0`.
  - The PR implements this behaviour, however I would like to have support to decide how to implement it in a non-breaking manner 

## Type of change
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

## TODOs

- [ ] The current tests will break due to some API changes
- [ ] Some steps could be quite unclear


sklego/meta/_grouped_utils.py
```diff
@@ -88,7 +91,46 @@ def _check_grouping_columns(X_group, **kwargs) -> pd.DataFrame:
...
     # Only check missingness in object columns
     if X_group.select_dtypes(exclude="number").isnull().any(axis=None):
-        raise ValueError("X has NaN values")
+        raise ValueError("Group columns contain NaN values")
 
     # The grouping part we always want as a DataFrame with range index
     return X_group.reset_index(drop=True)
+
+
+def _get_estimator(estimators, grp_values, grp_names, return_level, fallback_method):
```
FBruzzesi, line 100: The point of this function is to determine which estimator to use to predict.
- if `fallback_method ="raise"`, we have to have the model for the group we are predicting
- if `fallback_method ="next"`, we check for recursively for the first available parent
- if `fallback_method ="global"`, we summon the global model.

The point of returning a `return_level` is a trick to know how far back we went, and used to slice an array afterwards (more comments where this happens)

sklego/meta/grouped_predictor.py
```diff
@@ -212,112 +122,32 @@ def fit(self, X, y=None):
...
         """
-        X_group, X_value = _split_groups_and_values(
-            X, self.groups, min_value_cols=0, check_X=self.check_X, **self._check_kwargs
-        )
-
-        X_group = self.__add_shrinkage_column(X_group)
 
-        if y is not None:
-            y = check_array(y, ensure_2d=False)
+        # TODO: Validate class params?
```
FBruzzesi, line 126: Open question

sklego/meta/grouped_predictor.py
```diff
@@ -212,112 +122,32 @@ def fit(self, X, y=None):
...
 
-        if self.shrinkage is not None:
-            self.__set_shrinkage_function()
+        if is_classifier(self.estimator):
+            self.classes_ = np.sort(np.unique(y))  # TODO: Must be sequential for the rest of the code to work
```
FBruzzesi, line 129: If `y` has classes in random order we fall short, should we enforce that?

sklego/meta/grouped_predictor.py
```diff
@@ -212,112 +122,32 @@ def fit(self, X, y=None):
...
+            self.n_classes_ = len(self.classes_)
 
-        # List of all hierarchical subsets of columns
-        self.group_colnames_hierarchical_ = expanding_list(X_group.columns, list)
+        self.groups_ = as_list(self.groups)
 
-        self.fallback_ = None
+        # TODO: __grouped_predictor_target_value__?
+        frame = pd.DataFrame(X).assign(__target_value__=np.array(y)).reset_index(drop=True)
```
FBruzzesi, line 135: `__target_value__` (and `__global_model__` just right after) could be "safe enough" column names, but we can even prefix with `__grouped_predictor` or `__sklego_grouped_predictor`  to be extra safe that someone is not using those as column names 😂

sklego/meta/grouped_predictor.py
```diff
@@ -397,15 +208,160 @@ def decision_function(self, X):
...
+        if X_group_num.shape[1]:
+            check_array(X_group_num, **self._check_kwargs)
+
+        # Only check missingness in object columns
+        if X_groups.select_dtypes(exclude="number").isnull().any(axis=None):
+            raise ValueError("Group columns contain NaN values")
+
+        return self
+
+    def __set_fit_levels(self):
```
FBruzzesi, line 235: Send help here 😁

sklego/meta/grouped_predictor.py
```diff
@@ -397,15 +208,160 @@ def decision_function(self, X):
...
+    def __fit_estimators(self, frame):
+        """Fit one estimator per level of the group column(s)"""
+        estimators_ = {}
+        for grp_names in self.fitted_levels_:
+            for grp_values, grp_frame in frame.groupby(grp_names):
+                _X = grp_frame.drop(columns=self.groups_ + ["__target_value__"])
+                _y = grp_frame["__target_value__"]
+
+                if _y.isnull().all():
+                    estimators_[grp_values] = clone(self.estimator).fit(_X)  # TODO: or .fit(_X, None) ?
```
FBruzzesi, line 270: Not sure what the convention is for outlier detectors. In scikit-learn and pyod it seems that we can pass `y` and it will just get ignored anyway.
If we stand to that convention we don't need to distinguish the two cases

sklego/meta/grouped_predictor.py
```diff
@@ -397,15 +208,160 @@ def decision_function(self, X):
...
-            return self.__predict_shrinkage_groups(X_group, X_value, method="decision_function")
+            raise ValueError(f"`fallback_method` should be one of {self._ALLOWED_FALLBACK}, not {self.fallback_method}")
+
+        return levels_
+
+    def __fit_estimators(self, frame):
+        """Fit one estimator per level of the group column(s)"""
+        estimators_ = {}
+        for grp_names in self.fitted_levels_:
+            for grp_values, grp_frame in frame.groupby(grp_names):
```
FBruzzesi, line 265: Example here could be:
```py
groups = ["a", "b"]
fitted_levels_ = [["__global_model__"], ["__global_model__", "a"], ["__global_model__", "a", "b"]]
```
hence `grp_names` goes from outer to inner and  `grp_values` are the unique values identifying the group

sklego/meta/grouped_predictor.py
```diff
@@ -397,15 +208,160 @@ def decision_function(self, X):
...
+            result = self.shrinkage(group_lengths)
+        except Exception as e:
+            raise ValueError(f"Caught an exception while checking the shrinkage function: {str(e)}") from e
+        else:
+            if not isinstance(result, np.ndarray):
+                raise ValueError(f"shrinkage_function({group_lengths}) should return an np.ndarray")
+            if result.shape != expected_shape:
+                raise ValueError(f"shrinkage_function({group_lengths}).shape should be {expected_shape}")
+
+    def __predict_estimators(self, X, method_name):
```
FBruzzesi, line 333: This is a bit of a headache but allows for a unified approach:

- `preds` is a 3d array of `(n_samples, n_levels, n_classes)` where for regression or od `n_classes` can be considered to be 1. This will be populated with a prediction for each level
- `shrinkage` is a 2d array `(n_samples, n_levels)` as shrinkage factor is already one per level. This will get the shrinkage to use for each sample. Since we are doing predictions from outer to inner most level, we are overwriting when needed.
- The `_get_estimator` returning the level allows to select which model shrinkage to use for that particular prediction.
- Finally we multiply `preds` and `shrinkage` and sum over all levels. 
  - If shrinkage is none, the array will be only zeros and ones, therefore it is equivalent of selecting the model to use
  - If shrinkage is not none, then it is equivalent over averaging with shrinkage factors.

sklego/meta/_grouped_utils.py
```diff
@@ -45,13 +46,15 @@ def _split_groups_and_values(
     _shape_check(X, min_value_cols)
 
     try:
+        lgroups = as_list(groups)
```
koaning, line 49: I guess we should set `group_sizes` to also be a non-list type? Bit of a nit this one. 

Also: maybe `groups_list` instead of `lgroups`.

sklego/meta/_grouped_utils.py
```diff
@@ -88,7 +91,46 @@ def _check_grouping_columns(X_group, **kwargs) -> pd.DataFrame:
...
+        try:
+            return estimators[grp_values], return_level
+        except KeyError:
+            if len(grp_values) == 1:
+                raise KeyError(
+                    f"No fallback/parent estimator found for the given group values: {grp_names}={grp_values}"
+                )
+            return _get_estimator(estimators, grp_values[:-1], grp_names[:-1], return_level - 1, fallback_method)
+
+    else:  # fallback_method == "global"
```
koaning, line 132: Nit: technically the `else` isn't needed anymore because the function would've returned otherwise. Might be nicer to confirm at the start of the function that a correct fallback method is chosen.

PR #623, opened 2024-02-29
Hierarchical predictors

### LINKED ISSUE #620: `HierarchicalPredictor` and `HierarchicalTransformer`
# Description

As noted in #616 and discussed on private channels, `GroupedPredictor` class has its shortcomings.

I am opening this issue to keep track on how we could expand from what we learned.
Here a list of features that `HierarchicalPredictor` and `HierarchicalTransformer` should have in my opinion:

1. Given an input group, all levels should be fitted, including a global model
    - Example: `groups = ["a", "b"]`, groups to fit are: `["global"], ["global", "a"], ["global", "a", "b"]`
    - This implies that now the order _always_ matters, not only when shrinkage is not None.
    - This is quite opinionated, but also guarantees that 3.1 (see later) always works.
2. Shrinkage available for **both** regression and classification tasks (tbd how it could be extended to outlier detection)
3. Fallback strategy to be either `"next"` or `"raise"`, meaning:
    - `"next"`: if a group value is not found at prediction/transformation time, it fallbacks to the first available group in the hierarchy.
    - `"raise"`: if a group value is not found at prediction/transformation time, an error is raised.
    - ~Disclaimer: better value instead of "next" can be found 😂~ what about `"parent"`?
4. Allows for extra params in fit (e.g. `sample_weight`), which is currently not possible.

A first draft of the implementation was developed in #618, however trying to keep the current behavior of `GroupedPredictor` as well as expanding its functionalities was going to become a headache both to develop and maintain. This is why an implementation from scratch would make sense.

## Edit

A few considerations regarding the hypothetical `HierarchicalTransformer`:

- Shrinkage doesn't really make sense, or at least not for all transformers. In fact the current `GroupedTransformer` doesn't support it.
- Should the group columns be returned as they are from the `.transform(X)` operation?
- Should the `.transform(X)` output maintain the input type? Hence if pandas dataframe maintain the index  as well?

FBruzzesi:
# Description

- Introduces `HierarchicalPredictor`, `HierarchicalClassifier` and `HierarchicalRegressor` - Partially fixes #620 
- Introduces `ShrinkageMixin` to abstract such shared functionalities from `HierarchicalPredictor` and `GroupedPredictor`
- Adds a new shrinkage built-in function `equal_shrinkage`

## Type of change

## Checklist:

- [X] I have made corresponding changes to the documentation (also to the readme.md)


docs/api/shrinkage-functions.md
```diff
@@ -0,0 +1,33 @@
+# Shrinkage
```
FBruzzesi, line 1: Deserve their own page, at least in API

docs/user-guide/meta-models.md
```diff
@@ -177,6 +177,50 @@ This transformer also has use-cases beyond fairness. You could use this transfor
 
 For example, for predicting house prices, using the surface of a house relatively to houses in the same neighborhood could be a more relevant feature than the surface relative to all houses.
 
+## Hierarchical Prediction
+
+!!! info "New in version 0.8.0"
```
FBruzzesi, line 182: Again, totally made up version number. Need to be fixed accordingly

sklego/meta/_shrinkage_utils.py
```diff
@@ -0,0 +1,210 @@
...
+    ----------
+    group_sizes : array-like
+        The number of observations in each group, must implement the `__len__` method.
+
+    Returns
+    -------
+    np.ndarray
+        The weights for each group.
+    """
+    return np.ones(len(group_sizes))
```
FBruzzesi, line 86: Not sure if we want this to exist, but it felt like an easy one to add

sklego/meta/_shrinkage_utils.py
```diff
@@ -0,0 +1,210 @@
...
+    def _set_shrinkage_function(self):
+        """Set the shrinkage function and validate it if it is a custom callable"""
+        if isinstance(self.shrinkage, str) and self.shrinkage in self._ALLOWED_SHRINKAGE.keys():
+            shrinkage_function_ = self._ALLOWED_SHRINKAGE[self.shrinkage]
+
+        elif callable(self.shrinkage):
+            self.__check_shrinkage_func()
+            shrinkage_function_ = self.shrinkage
+
+        elif self.shrinkage is None:
```
FBruzzesi, line 118: We discussed about adding proper comments in here. Please let me know if it is reasonable and clear enough

tests/test_meta/test_hierarchical_predictor.py
```diff
@@ -0,0 +1,144 @@
...
+def test_shrinkage(meta_cls, base_estimator, task, metric, shrinkage):
+    """Tests that the model performance is better than the base estimator when predicting with different shrinkage
+    methods.
+    """
+    X, y, groups = make_hierarchical_dataset(task)
+
+    meta_model = meta_cls(estimator=clone(base_estimator), groups=groups, **shrinkage).fit(X, y)
+    base_model = clone(base_estimator).fit(X.drop(columns=groups), y)
+
+    assert metric(y, base_model.predict(X.drop(columns=groups))) <= metric(y, meta_model.predict(X))
```
FBruzzesi, line 144: This is some sort of behavioural test

docs/user-guide/meta-models.md
```diff
@@ -177,6 +177,50 @@ This transformer also has use-cases beyond fairness. You could use this transfor
...
+
+    !!! warning
+        As a consequence of this, the **order of groups matters** and potentially a combinatoric number of estimators are fitted, one for each unique combination of group values and each level, including a global one.
+
+2. `HierarchicalClassifier` is meant to properly handle shrinkage for classification tasks - however this requires that the base estimator implements a `.predict_proba()` method.
+3. While `GroupedPredictor` is meant to be used directly, `HierarchicalPredictor` is meant to be used as a base class for other estimators such as `HierarchicalClassifier` and `HierarchicalRegressor`.
+
+## Shrinkage Functions
+
+scikit-lego provides a set of shrinkage functions that can be used to shrink the predictions of a model in [Grouped Prediction] and [Hierarchical Predictions].
```
koaning, line 202:
```suggestion
Scikit-lego provides a set of shrinkage functions that can be used to shrink the predictions of a model in [Grouped Prediction] and [Hierarchical Predictions].
```

docs/user-guide/meta-models.md
```diff
@@ -177,6 +177,50 @@ This transformer also has use-cases beyond fairness. You could use this transfor
...
+
+Such function should takes a list of group sizes and returns an array of the same size with the weights (positive values) for each group.
+
+```py title="Custom shrinkage function"
+import numpy as np
+
+def random_shrinkage_function(group_sizes):
+    """A custom random shrinkage function"""
+    a = np.random.uniform(low=0., high=1., size=len(group_sizes))
+    return a / a.sum()
```
koaning, line 221: Maybe something with exponential smoothing instead? A random number as a shrinkage function doesn't feel like it's tangible or real-life enough.

FBruzzesi:
Forgot to mention a couple of ideas:

- `estimators_` attribute is something like the following dictionary:

    ```terminal
    {
        (1,): LogisticRegression(),
        (1, 'A'): LogisticRegression(),
        (1, 'B'): LogisticRegression(),
        (1, 'A', 'X'): LogisticRegression(),
        (1, 'B', 'Y'): LogisticRegression(),
    }
    ```
    which is ok-ish, but not super interpretable. One idea could be to have a namedtuple:
    ```python
    from collections import namedtuple
   
   groups, grp_values = ["col1", "col2"], [1,2]
    estimator = namedtuple("estimator", groups)

    estimator(*grp_values)
    # estimator(col1=1, col2=2)
    ```
    This hits edge cases if a group has a "_" prefix (maybe we can trim that), but the end result would be:

   ```terminal
    {
        estimator(sklego_global_estimator=1): LogisticRegression(),
        estimator(sklego_global_estimator=1, col1="A"): LogisticRegression(),
        estimator(sklego_global_estimator=1, col1="B"): LogisticRegression(),
        estimator(sklego_global_estimator=1, col1="A", col2="X"): LogisticRegression(),
        estimator(sklego_global_estimator=1, col1="B", col2="Y"): LogisticRegression(),
    }
    ```

- Possibility of accessing an estimator from its dictionary directly, namely being able to do `hierarchical_predictor[(1, "A", "X")]` in place of `hierarchical_predictor.estimators_[(1, "A", "X")]`
- Related to both the above, as adding the 1 for the global model is not very ergonomic, we can create shortcuts by aliasing externally what is in the first point as:
   ```terminal
    {
        "__sklego_global_estimator__": LogisticRegression(),
        ('A', ): LogisticRegression(),
        ('B', ): LogisticRegression(),
        ('A', 'X'): LogisticRegression(),
        ('B', 'Y'): LogisticRegression(),
    }
    ```
    and all the associated "conversions"  from the first and/or second point, if accepted

koaning:
At the moment I'd be fine with this: 

```
{
    (1,): LogisticRegression(),
    (1, 'A'): LogisticRegression(),
    (1, 'B'): LogisticRegression(),
    (1, 'A', 'X'): LogisticRegression(),
    (1, 'B', 'Y'): LogisticRegression(),
}
```

But it would be preferable to have this documented in the code with comments. I agree that it's  merely "ok", but it feels clear enough if the comment is there, no?

FBruzzesi:
> But it would be preferable to have this documented in the code with comments. I agree that it's merely "ok", but it feels clear enough if the comment is there, no?

Just added a few comments on those. Here is how it looks like:
image

koaning:
Ah yeah, that's even nicer. Just in the comments would've been sufficient but adding it in the docs for sure is a nice touch.

tests/test_meta/test_hierarchical_predictor.py
koaning: I guess there's an opportunity for one more test that leverages the [DummyRegressor](https://scikit-learn.org/stable/modules/generated/sklearn.dummy.DummyRegressor.html). 

```python
df = pd.DataFrame([
  {"grp": "a", "x": 1, "y": 0},
  {"grp": "a", "x": 1, "y": 0},
  {"grp": "a", "x": 1, "y": 0},
  {"grp": "b", "x": 1, "y": 2},
  {"grp": "b", "x": 1, "y": 2},
  {"grp": "b", "x": 1, "y": 2},
])
```

We should be able to confirm that the parent model would have a model that always predicts 1. Then depending on the shrinkage factor we might be able to have some expectation on predictions for group a and b, no?

koaning (changes requested):
My main observation is that we might want to add an extra test that uses a dummy model to help predict the values that we'd expect.

FBruzzesi:
> My main observation is that we might want to add an extra test that uses a dummy model to help predict the values that we'd expect. Other than that; this looks great! Good work :)

I am thinking out loud here but...maybe the easiest way to check this is to use a deterministic/fake predictions model. Does that sound reasonable?

koaning:
Oh that was totally in line with what I had in mind. I've used Dummy models for this in the past but you're also free to pick another method. As long as we just have a test that our assumptions on how we shrink play out.

tests/test_meta/test_hierarchical_predictor.py
FBruzzesi: For some reason I didn't see this comment. I ended up with a somewhat similar approach, namely: a dataset that could be used for both regression and classification with dummy estimators (see `make_hierarchical_dummy`) used in `test_expected_output` function.

There are comments on why the expected outputs are the way they are. I hope it is good enough. Let me know what you think

koaning (approved):
LGTM

PR #624, opened 2024-03-05
`GroupedTransformer` with transformers that use `y`

### LINKED ISSUE #459: [BUG] GroupedTransformer can't deal with transformers that use `y`
Hi! First of all thanks for your package, it's awesome! (Same as calmcode and other resources you built).

I'm trying to perform a LeaveOneOutEncoder transformation for each group of my dataset and I think it's not possible (or I can't get it), the problem is that within `sklego/meta/grouped_transformer.py` the fit is called without `y` and it throws an error: `TypeError: fit() missing 1 required positional argument: 'y'`.

I post here a dummy example that has the same functionality that the thing I'm trying to do in my code:

```python
import pandas as pd
import numpy as np

from sklego.datasets import load_heroes
from sklego.meta import GroupedTransformer
from category_encoders import LeaveOneOutEncoder


df_heroes = load_heroes(as_frame=True).replace([np.inf, -np.inf], np.nan).dropna()

# Dummy example
X = df_heroes[['attack_type', 'health', 'attack_spd']]
y = df_heroes['attack']

GroupedTransformer(LeaveOneOutEncoder(), groups=['attack_type']).fit_transform(X, y)
```

Thansk in advance!

FBruzzesi:
# Description

Fixes #459 by passing `y` when fitting the global transformer as well.

## Type of change

## Checklist:

### DIFF
--- a/sklego/meta/grouped_transformer.py
+++ b/sklego/meta/grouped_transformer.py
@@ -31,10 +31,11 @@ class GroupedTransformer(BaseEstimator, TransformerMixin):
 
     _check_kwargs = {"accept_large_sparse": False}
 
-    def __init__(self, transformer, groups, use_global_model=True):
+    def __init__(self, transformer, groups, use_global_model=True, check_X=True):
         self.transformer = transformer
         self.groups = groups
         self.use_global_model = use_global_model
+        self.check_X = check_X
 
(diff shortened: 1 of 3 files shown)


koaning (approved):
LGTM

PR #625, opened 2024-03-11
0.8.0 Release

FBruzzesi:
# Description

Preparing for future release.


koaning:
This makes me wonder ... should the docs maybe list a changelog? Similar maybe to what we did for [Prodigy](https://prodi.gy/docs/changelog)?

FBruzzesi:
That would be really neat indeed! How hard would it be to automate?

koaning:
I think you'll always need some manual work, if only to link to relevant places in the docs.

PR #647, opened 2024-04-05
Update this.py

FBruzzesi:
# Description

`this.py` contribution - no issue to fix, just some swag!


koaning:
Made a tiny modification, bit of spelling and another edit to keep the original final line the final line.

PR #649, opened 2024-04-08
docs: `ClusterFoldValidation` api

FBruzzesi:
# Description

While spinning the documentation locally with the latest changes, I noticed that `ClusterFoldValidation` didn't make it into the API section since classes are manually added there.

I also bumped the lib version, as this change in the documentation should appear only once the feature is released.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "scikit-lego"
-version = "0.9.0"
+version = "0.8.2"
 description="A collection of lego bricks for scikit-learn pipelines"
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "scikit-lego"
-version = "0.8.1"
```
koaning, line 7: Technically ... we're not introducing breaking changes here, right? So we might also consider 0.8.2? Or am I missing something?

FBruzzesi: True, let me change it in both places

koaning (approved):

PR #651, opened 2024-04-19
feat: numpy 2.0rc support

FBruzzesi:
# Description

There are 3 folds on this PR:

- Numpy changes needed for v2 which are compatible with v1:
  - `np.lib.pad` to `np.pad`
  - `np.NaN` to `np.nan`
- Pytest marks
- Constraining `numpy<2.0` for cvxpy and umap extra dependencies: it is out of our control for them to support latest numpy version.
  - This last bit led me to separate the cron gha into base and extra

Edit: [Manual trigger of cron gha is successful](https://github.com/FBruzzesi/scikit-lego/actions/runs/8751350431)

### DIFF
--- a/.github/workflows/schedule-dependencies.yml
+++ b/.github/workflows/schedule-dependencies.yml
@@ -58,7 +58,7 @@ jobs:
     - name: Install dependencies
       run: |
         uv pip install wheel --system
-        uv pip install ${{ matrix.pre-release-dependencies }} -e ".[test]" --system
+        uv pip install ${{ matrix.pre-release-dependencies }} -e ".[test-all]" --system
         uv pip freeze
     - name: Test with pytest
       run: pytest -n auto --disable-warnings --cov=sklego -m "cvxpy or formulaic or umap"
--- a/.github/workflows/schedule-dependencies.yml
+++ b/.github/workflows/schedule-dependencies.yml
@@ -7,7 +7,7 @@ on:
   
 
 jobs:
-  cron:
+  cron-base:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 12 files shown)


tests/test_estimators/test_umap_reconstruction.py
```diff
@@ -5,6 +5,8 @@
 from sklego.decomposition import UMAPOutlierDetection
 from tests.conftest import general_checks, nonmeta_checks, outlier_checks, select_tests
 
+pytestmark = pytest.mark.umap
```
koaning, line 8: I've never used this feature before, just to check my understanding ... does this set the marker on every test in this file?

.github/workflows/schedule-dependencies.yml
```diff
@@ -30,9 +30,35 @@ jobs:
...
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
       run: |
-        uv pip install -e ".[test]" --system
-        make test
+        uv pip install wheel --system
+        uv pip install ${{ matrix.pre-release-dependencies }} -e ".[test-all]" --system
+        uv pip freeze
+    - name: Test with pytest
+      run: pytest -n auto --disable-warnings --cov=sklego -m "cvxpy or formulaic or umap"
```
koaning, line 64: I wonder ... is the `-m` part needed here? Don't we test for everything by default?

koaning (commented):
Looks great overall, just wanted to double-check two things.

tests/test_estimators/test_umap_reconstruction.py
FBruzzesi: Yes correct, it _marks_ the entire module/file!

.github/workflows/schedule-dependencies.yml
FBruzzesi: Oh true, by default everything is tested, with such flag just the _extras_. I was thinking doing a sort of "delta".

koaning: That can still be a good idea. I guess we could first do a run without those marked tests and then run a step for each of them. Just make sure that you always run each step.

FBruzzesi: Split it using matrix, it gets a little spammy but very detailed ([example](https://github.com/FBruzzesi/scikit-lego/actions/runs/8757546995))

.github/workflows/schedule-dependencies.yml
```diff
@@ -30,9 +30,36 @@ jobs:
     - name: Install dependencies
       run: |
         uv pip install wheel --system
-        uv pip install ${{ matrix.pre-release-dependencies }} scikit-lego --system
+        uv pip install ${{ matrix.pre-release-dependencies }} --system -e ".[test-dep]"
         uv pip freeze
     - name: Test with pytest
+      run: pytest -n auto --disable-warnings --cov=sklego -m "not cvxpy and not formulaic and not umap"
```
FBruzzesi, line 36: As far as I know, there is no easy way to ask only for those tests without a "registered" mark.
I tried to achieve that dynamically via [pytest_collection_modifyitems](https://docs.pytest.org/en/latest/reference/reference.html#pytest.hookspec.pytest_collection_modifyitems), but it's a bit of a headache as `@pytest.mark.parametrize` and any other pytest marker count as a mark in the items

koaning: As-is seems perfectly fine :)

koaning (approved):

Issue #653, opened 2024-04-26
`decomposition.umap_reconstruction.UMAPOutlierDetection`


FBruzzesi:

FBruzzesi:
@anopsy created separate issues to link directly PRs

anopsy:
Thanks! I'll also repeat my question:

**what would be a more elegant example - applying them on ndrrays or a dataframe?**

I noticed that some examples in sklego are based on dataframes and some on numpy arrays or ndarrays. I noticed that scikit-learn mostly uses arrays in their examples. Let me know what do you think, maybe I'm overthinking that

FBruzzesi:
I would suggest to use arrays as well wherever possible. Dataframes may be more interpretable but they are more in the direction of skrub philosophy

anopsy:
Coolio! That's more of an idea for the future, but when all examples are added, maybe it would be good to base them all on arrays then. Just sayin'

FBruzzesi:
Closed by #654

PR #654, opened 2024-04-27
docs: PCA and UMAP OutlierDetection docstring examples

### LINKED ISSUE #652: `decomposition.pca_reconstruction.PCAOutlierDetection`

anopsy:
# Docs

Added usage examples to:
decomposition.umap_reconstruction.UMAPOutlierDetection
decomposition.pca_reconstruction.PCAOutlierDetection

Fixes #652 and #653 

## Type of change

- [ x] New feature (non-breaking change which adds functionality)


anopsy:
image
Wondering what happend here

FBruzzesi (commented):
Just wondering, how did you come up with these array values?
Wouldn't make sense to have some randomly generated arrays (with fixed seed) and then tinker around some value to make them outliers? Point is: currently I cannot spot outliers just by looking at the arrays and everything seems a bit magical.

**On the Github action failing**: Getting a sense that macos-latest runner is on macos 14, and it is breaking most of github actions.
I can't deep dive today, but will try in the coming week.

Anyway, certainly unrelated from the docstring changes

anopsy:
Oh gee, there are even more failed tests now.😮

About examples
I copied the approach with the arrays from scikit-learn docs on PCA (check ss) 
image

and I tinkered around with the values to check how the detectors work for different values in the arrays, n_components and thresholds. I tried to add values that would be clearly an outlier because of the quantiles for example [-100, 99, -99] but the PCA/UMAPOutlierDetectors "classified" them as inliers. I also saw in User Guide  that the values classified by PCA/UMAPOutlierDetectors as outliers, don't look like quantiles based outliers -so you can't  spot them just by looking. If that makes any sense. 

image

FBruzzesi:
Yep the doc page is using iris dataset, which I would not expect to have any particular outlier.
We have one [obvious example](https://github.com/koaning/scikit-lego/blob/main/tests/test_estimators/test_umap_reconstruction.py#L42) in the test suite.

@koaning thoughts on this? In my opinion, it could be worth it to change dataset in the user guide as well. It seems a bit confusing

anopsy:
Sure, I'll do it the way it's done in the test suite.

FBruzzesi (approved):
Let's ship it ✨

PR #655, opened 2024-05-05
feat: make ColumnDropper dataframe-agnostic

MarcoGorelli:
Hey 👋 I discussed a bit with Francesco, who discussed it with Vincent. So, I was keen to get the ball rolling

# Description

Making a start towards dataframe-agnosticism. For now, this only makes `ColumnDropped` dataframe-agnostic. It requires adding an extremely lightweight dependecy (Narwhals), but the result is that computation can happen natively for pandas/Polars/modin/cuDF (and any other library which may want to become Narwhals-compliant), without any data conversion.

pandas is still a required dependency, as it's required for the rest of scikit-lego. But, it doesn't seem that far-fetched to be able to make all of scikit-lego dataframe-agnostic. If you're open to this, we could do it in stages?

There is no impact on current users, other than that Narwhals would be a required dependency (though it's really lightweight, and without any dependencies itself, so there's no risk for conflicts)

## Type of change

## Checklist:

Not sure about some of these items so I haven't checked them.

---

## Demo

pandas users, they can keep using `scikit-lego`, without needing Polars to be installed
```console
$ pip uninstall polars -y
Found existing installation: polars 0.20.23
Uninstalling polars-0.20.23:
  Successfully uninstalled polars-0.20.23
$ cat t.py 
import pandas as pd
from sklego.preprocessing import ColumnDropper

df = pd.DataFrame({
    "name": ["Swen", "Victor", "Alex"],
    "length": [1.82, 1.85, 1.80],
    "shoesize": [42, 44, 45]
})
print(ColumnDropper(["name"]).fit_transform(df))
$ python t.py 
   length  shoesize
0    1.82        42
1    1.85        44
2    1.80        45
```

But, Polars users can use this, without it doing any conversion to pandas. In particular, Polars LazyFrame is supported (and stays lazy, but here's I'm calling `.collect` to show you the output):
```console
$ cat t.py 
import polars as pl
from sklego.preprocessing import ColumnDropper

df = pl.LazyFrame({
    "name": ["Swen", "Victor", "Alex"],
    "length": [1.82, 1.85, 1.80],
    "shoesize": [42, 44, 45]
})
result = ColumnDropper(["name"]).fit_transform(df)
print(result.explain())
print(result.collect())
$ python t.py 
DF ["name", "length", "shoesize"]; PROJECT 2/3 COLUMNS; SELECTION: "None"
shape: (3, 2)
┌────────┬──────────┐
│ length ┆ shoesize │
│ ---    ┆ ---      │
│ f64    ┆ i64      │
╞════════╪══════════╡
│ 1.82   ┆ 42       │
│ 1.85   ┆ 44       │
│ 1.8    ┆ 45       │
└────────┴──────────┘
```

### DIFF
--- a/sklego/preprocessing/pandastransformers.py
+++ b/sklego/preprocessing/pandastransformers.py
@@ -1,3 +1,4 @@
+import narwhals as nw
 import pandas as pd
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.utils.validation import check_is_fitted
@@ -6,7 +7,7 @@ from sklego.common import as_list
 
 
 class ColumnDropper(BaseEstimator, TransformerMixin):
-    """The `ColumnDropper` transformer allows dropping specific columns from a pandas DataFrame by name.
+    """The `ColumnDropper` transformer allows dropping specific columns from a DataFrame by name.
     Can be useful in a sklearn Pipeline.
 
     Parameters
@@ -18,6 +19,15 @@ class ColumnDropper(BaseEstimator, TransformerMixin):
     ----------
     feature_names_ : list[str]
         The names of the features to keep during transform.
+
+    Notes
+    -----
+    Supported dataframes are:
+
+    - pandas
+    - Polars (eager or lazy)
+    - Modin
+    - cuDF
 
     Examples
     --------
@@ -81,14 +91,14 @@ class ColumnDropper(BaseEstimator, TransformerMixin):
 
         Checks:
 
-        1. If input is a `pd.DataFrame` object
+        1. If input is a supported DataFrame
         2. If column names are in such DataFrame
 
         Parameters
         ----------
-        X : pd.DataFrame
+        X : DataFrame
             The data on which we apply the column selection.
-        y : pd.Series, default=None
+        y : Series, default=None
             Ignored, present for compatibility.
 
         Returns
@@ -99,25 +109,25 @@ class ColumnDropper(BaseEstimator, TransformerMixin):
         Raises
         ------
         TypeError
-            If `X` is not a `pd.DataFrame` object.
+            If `X` is not a supported DataFrame.
         KeyError
             If one or more of the columns provided doesn't exist in the input DataFrame.
         ValueError
             If dropping the specified columns would result in an empty output DataFrame.
         """
         self.columns_ = as_list(self.columns)
-        self._check_X_for_type(X)
+        X = nw.from_native(X)
         self._check_column_names(X)
-        self.feature_names_ = X.columns.drop(self.columns_).tolist()
+        self.feature_names_ = [x for x in X.columns if x not in self.columns_]
         self._check_column_length()
         return self
 
     def transform(self, X):
-        """Returns a pandas DataFrame with only the specified columns.
+        """Returns a DataFrame with only the specified columns.
 
         Parameters
         ----------
-        X : pd.DataFrame
+        X : DataFrame
             The data on which we apply the column selection.
 
         Returns
@@ -131,10 +141,10 @@ class ColumnDropper(BaseEstimator, TransformerMixin):
             If `X` is not a `pd.DataFrame` object.
         """
         check_is_fitted(self, ["feature_names_"])
-        self._check_X_for_type(X)
+        X = nw.from_native(X)
         if self.columns_:
-            return X.drop(columns=self.columns_)
-        return X
+            return nw.to_native(X.drop(self.columns_))
+        return nw.to_native(X)
 
     def get_feature_names(self):
         """Alias for `.feature_names_` attribute"""
@@ -150,12 +160,6 @@ class ColumnDropper(BaseEstimator, TransformerMixin):
         non_existent_columns = set(self.columns_).difference(X.columns)
         if len(non_existent_columns) > 0:
             raise KeyError(f"{list(non_existent_columns)} column(s) not in DataFrame")
-
-    @staticmethod
-    def _check_X_for_type(X):
-        """Checks if input of the Selector is of the required dtype"""
-        if not isinstance(X, pd.DataFrame):
-            raise TypeError("Provided variable X is not of type pandas.DataFrame")
 
 
 class PandasTypeSelector(BaseEstimator, TransformerMixin):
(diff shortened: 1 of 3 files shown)


sklego/preprocessing/pandastransformers.py
```diff
@@ -151,12 +152,6 @@ def _check_column_names(self, X):
         if len(non_existent_columns) > 0:
             raise KeyError(f"{list(non_existent_columns)} column(s) not in DataFrame")
 
-    @staticmethod
-    def _check_X_for_type(X):
-        """Checks if input of the Selector is of the required dtype"""
-        if not isinstance(X, pd.DataFrame):
-            raise TypeError("Provided variable X is not of type pandas.DataFrame")
```
MarcoGorelli, line 158: I'm removing this here, as `nw.from_native` already raises a similar error. For example, if you tried passing in a numpy array here, you'd get:
```python
TypeError: Expected pandas-like dataframe, Polars dataframe, or Polars lazyframe, got: <class 'numpy.ndarray'>
```

sklego/preprocessing/pandastransformers.py
```diff
@@ -19,6 +20,15 @@ class ColumnDropper(BaseEstimator, TransformerMixin):
     feature_names_ : list[str]
         The names of the features to keep during transform.
 
+    Notes
+    -----
+    Supported dataframes are:
```
FBruzzesi, line 25: Consider adding a link to narwhals docs where the list of supported dataframe libraries is mentioned.
Currently I was able to spot it in [extending section](https://marcogorelli.github.io/narwhals/extending/) only.

tests/test_preprocessing/test_columndropper.py
```diff
@@ -1,6 +1,8 @@
 import pandas as pd
+import polars as pl
```
FBruzzesi, line 2: Just some food for thoughts (I can work on this later), we may parametrize the fixture in order to parametrize tests

FBruzzesi (commented):
Honestly it looks quite neat! I left some considerations for _better sync_ with `narwhals` in the long term.

Edit: Regarding macos tests, we are getting a lot of fails since latest upgraded to macos-14. We definitely need to get a closer look at those, but completely unrelated to the PR

sklego/preprocessing/pandastransformers.py
```diff
@@ -6,7 +7,7 @@
 
 
 class ColumnDropper(BaseEstimator, TransformerMixin):
-    """The `ColumnDropper` transformer allows dropping specific columns from a pandas DataFrame by name.
+    """The `ColumnDropper` transformer allows dropping specific columns from a DataFrame by name.
```
koaning, line 10: Maybe good to mention that we use narwhals under the hood here?

MarcoGorelli: thanks for your review! Sure, I've mentioned it in the "notes" section

koaning: I was wondering if it makes sense to mention earlier, but yeah, if it's in the notes it's probably fine.

FBruzzesi:
@koaning I finally took a look at the CI...and you are right 😁

- Moving back to pip doesn't break when building wheel ([proof on my fork](https://github.com/FBruzzesi/scikit-lego/pull/9))
- uv has some related issue when [building wheels from scratch](https://github.com/astral-sh/uv/issues/3308) and [qdldl](https://pypi.org/project/qdldl/#files)  has no built distribution for macos 10+ **and** arm architecture.

As a side note: even with pip, the `--pre` flag break as of `scikit-learn==1.5.0rc1` ([proof on my fork]( https://github.com/FBruzzesi/scikit-lego/actions/runs/8975596456/job/24650556520)).

tests/test_preprocessing/test_columndropper.py
FBruzzesi: @MarcoGorelli I took the freedom to refactor this test suite via parametrization

FBruzzesi:
@koaning another food for thought is that... all these transformers lie in a `../pandastransformers.py` module, however this should be transparent to the user as they are import-able from `sklego.preprocessing` directly. I would feel fairly safe to change the filename once all of them are dataframe agnostic

koaning:
@FBruzzesi any objections to moving back to `pip` for now? I prefer to wait until `uv` is a bit more reliable before relying our CI on it. 

I made [an issue for that v1.5rc01 test failure on the sklearn side](https://github.com/scikit-learn/scikit-learn/issues/28966), feels like it might be breaking so upstream should know about it.

> I would feel fairly safe to change the filename once all of them are dataframe agnostic. 

Totally!

FBruzzesi (approved):
I am biased as I added a few commits myself 😂

koaning:
Do we want to release early with a micro release by the way or would we prefer to make a bigger splash? There's something to be said for both, but given that we are introducing a dependency ... might be good to make a larger release?

anopsy:
If everybody is fine with it, I could finish pandastransformers.py, or at least have a shot at ColumnDropper and see how it goes

anopsy:
I meant ColumnSelector

FBruzzesi:
@koaning I would personally opt for one _somewhat big_ release. We can wait to merge the PRs related to Narwhals migration to have some space for bug fixes if needed.

@anopsy Of course feel free to develop that. I will create an issue to track progress 😇

MarcoGorelli:
Just got a question about

> We can wait to merge the PRs related to Narwhals migration to have some space for bug fixes if needed.

No objections to waiting for a bigger release to include this, I'm just worried that it's going to become quite hard to review if everything is done in a single PR, or if PRs build on top of other open PRs

May I suggest that you (i.e. a scikit-lego committer) make a `narwhals` branch in `scikit-lego`, we make PRs to that branch, and then if/when you're ready, you merge the `narwhals` branch into `main` and include it in the release?

FBruzzesi:
@MarcoGorelli that's very reasonable! Here you have it: [`narwhals-development` branch](https://github.com/koaning/scikit-lego/tree/narwhals-development).

koaning:
FWIW I think pushing to main is fine too. It's more of a "oh, we have one new feature, usually that means deploy". But we can also collect a bunch of things to main ... this is the main thing we're going to work on short-term.

PR #659, opened 2024-05-08
Agnostic ColumnSelector

### LINKED ISSUE #658: [FEATURE] Narwhals migration for dataframe-agnostic codebase
# Description

Creating this issue to keep track of which classes/function could benefit from adopting Narwhals.

| Class/Function/Module | Status | Related | Notes |
| - | - | - | - |
|  `preprocessing.ColumnDropper` | ✅ | Solved in #655 | |
|  `preprocessing.ColumnSelector` | ✅  | Solved in #659 | | 
|  `preprocessing.PandasTypeSelector` | ✅ | Solved in #670  |  Consider creating another class `TypeSelector` and deprecate this one |
|  `common.TrainOnlyTransformerMixin` |🔲 |  | Uses specific pandas hashing functionality. I wonder how crucial is to hash the index as well. If it's not we could just use `.to_numpy()` and hash the array data? |
|  `model_selection.TimeGapSplit` | ✅ | Solved in #668  |  |
|  `model_selection.GroupedTimeSeriesSplit` | 🔲|  | Related to #605  |
|  `projections.InformationFilter` | ✅ | Solved in [commit](https://github.com/koaning/scikit-lego/commit/0773db9325e87041519aed79371ee30955e0028f) | |
|  `meta.RegressionOutlierDetector` | ✅| Solved in #665  | |
|  `meta.hierarchical_predictor.py` | ✅| Solved in #667  |  |
|  `meta.grouped_transformer.py` | ✅| Solved in #667  |  |
|  `meta.grouped_predictor.py` | ✅| Solved in #667  | |
|  `linear_models._FairClassifier` | ✅ | Solved in #669  |  |
|  `pandas_utils.py` | 🚧| Partially #661  | (*) |
|  `datasets.py` | 🚫|  | It would require `read_csv` function |

Personally I would wait to have at least `preprocessing.pandastransformers.py` entire migration before bumping to v0.9.0.

cc @MarcoGorelli @anopsy 

## (*) Regarding `pandas_utils`

Changing `log_step` to narwhals is fairly easy ([~4 lines of code](https://github.com/koaning/scikit-lego/blob/feat/narwhals-utils/sklego/pandas_utils.py#L95)), however as this decorator is supposed to work for _any_ function that operates on pandas, doing so would limit its functionality. It could be reasonable to add another one which although restricted to narwhals methods, it can interoperate with all its compatible dataframes.

**Legend**

✅ Done 
🚧 WIP 
🔲 Not Started 
🚫 Won't do

anopsy:
# Description

Narwhalyfied ColumnSelector, I added an example with polars df to the docstrings and adjusted the test_columnselector.py
I tried to test the docstrings, but it doesn't work. Maybe we could add those to sklego in the future?

Fixes #658 - partly

## Type of change

## Checklist:

### DIFF
--- a/sklego/preprocessing/pandastransformers.py
+++ b/sklego/preprocessing/pandastransformers.py
@@ -321,11 +321,14 @@ class ColumnSelector(BaseEstimator, TransformerMixin):
     -----
     Native cross-dataframe support is achieved using
     [Narwhals](https://narwhals-dev.github.io/narwhals/){:target="_blank"}.
+
     Supported dataframes are:
+
     - pandas
     - Polars (eager or lazy)
     - Modin
     - cuDF
+
     See [Narwhals docs](https://narwhals-dev.github.io/narwhals/extending/){:target="_blank"} for an up-to-date list
     (and to learn how you can add your dataframe library to it!).
 
--- a/sklego/preprocessing/pandastransformers.py
+++ b/sklego/preprocessing/pandastransformers.py
@@ -372,16 +372,16 @@ class ColumnSelector(BaseEstimator, TransformerMixin):
 
     ColumnSelector(["length", "shoesize"]).fit_transform(df_pl)
     '''
-        shape: (3, 2)
-        ┌────────┬──────────┐
-        │ length ┆ shoesize │
-        │ ---    ┆ ---      │
-        │ f64    ┆ i64      │
-        ╞════════╪══════════╡
-        │ 1.82   ┆ 42       │
-        │ 1.85   ┆ 44       │
-        │ 1.8    ┆ 45       │
-        └────────┴──────────┘
+    shape: (3, 2)
+    ┌────────┬──────────┐
+    │ length ┆ shoesize │
+    │ ---    ┆ ---      │
+    │ f64    ┆ i64      │
+    ╞════════╪══════════╡
+    │ 1.82   ┆ 42       │
+    │ 1.85   ┆ 44       │
+    │ 1.8    ┆ 45       │
+    └────────┴──────────┘
     '''
 
 
--- a/sklego/preprocessing/pandastransformers.py
+++ b/sklego/preprocessing/pandastransformers.py
@@ -309,13 +309,25 @@ class PandasTypeSelector(BaseEstimator, TransformerMixin):
 
 
 class ColumnSelector(BaseEstimator, TransformerMixin):
-    """The `ColumnSelector` transformer allows selecting specific columns from a pandas DataFrame by name.
+    """The `ColumnSelector` transformer allows selecting specific columns from a DataFrame by name.
     Can be useful in a sklearn Pipeline.
 
     Parameters
     ----------
     columns : str | list[str]
         Column name(s) to be selected.
+
+    Notes
+    -----
+    Native cross-dataframe support is achieved using
+    [Narwhals](https://narwhals-dev.github.io/narwhals/){:target="_blank"}.
+    Supported dataframes are:
+    - pandas
+    - Polars (eager or lazy)
+    - Modin
+    - cuDF
+    See [Narwhals docs](https://narwhals-dev.github.io/narwhals/extending/){:target="_blank"} for an up-to-date list
+    (and to learn how you can add your dataframe library to it!).
 
     Attributes
     ----------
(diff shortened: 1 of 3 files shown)


sklego/preprocessing/pandastransformers.py
```diff
@@ -329,52 +344,68 @@ class ColumnSelector(BaseEstimator, TransformerMixin):
...
+    import polars as pl
+    from sklego.preprocessing import ColumnSelector
+
+    df_pl = pl.DataFrame({
+        "name": ["Swen", "Victor", "Alex"],
+        "length": [1.82, 1.85, 1.80],
+        "shoesize": [42, 44, 45]
+    })
+
+    ColumnSelector(["length", "shoesize"]).fit_transform(df_pl)
```
MarcoGorelli, line 371: what you show below here doesn't look like Polars output, I think that may need updating?

sklego/preprocessing/pandastransformers.py
```diff
@@ -404,40 +435,41 @@ def fit(self, X, y=None):
...
         ValueError
-            If dropping the specified columns would result in an empty output DataFrame.
+            If provided list of columns to select is empty and would result in an empty output DataFrame.
         """
         self.columns_ = as_list(self.columns)
-        self._check_X_for_type(X)
-        self._check_column_length()
+        X = nw.from_native(X)
         self._check_column_names(X)
+        self.feature_names_ = [x for x in X.columns if x in self.columns_]
```
MarcoGorelli, line 447: Why is this line needed?

sklego/preprocessing/pandastransformers.py
```diff
@@ -404,40 +435,41 @@ def fit(self, X, y=None):
...
         TypeError
-            If `X` is not a `pd.DataFrame` object.
+            If `X` is not a supported DataFrame.
         """
-        self._check_X_for_type(X)
+        X = nw.from_native(X)
         if self.columns:
-            return X[self.columns_]
-        return X
+            return nw.to_native(X.select(nw.col(self.columns_)))
```
MarcoGorelli, line 471: bit simpler, you can just do `X.select(self._columns_)`

MarcoGorelli (commented):
that was fast! just left some initial comments ahead of maintainers' reviews

sklego/preprocessing/pandastransformers.py
MarcoGorelli: I think you don't need `self.feature_names_` at all, it's not in the current code

sklego/preprocessing/pandastransformers.py
anopsy: Let me fix that

anopsy: image
So I used polars output but it renders this way in mkdocs:

image

MarcoGorelli: ah I think it's just a font that needs setting, the default one isn't exactly monospace

sklego/preprocessing/pandastransformers.py
anopsy: indeed it's not in Attributes list, I see now that the fit performs only checks, cool thank you

sklego/preprocessing/pandastransformers.py
anopsy: Is that something I can fix or not really @FBruzzesi ?

FBruzzesi: I never rendered that! @MarcoGorelli send help 😂!

MarcoGorelli: 😄 dunno, in Narwhals I just have
```
theme:
  name: material
  font: false
```

koaning: I recall doing stuff like that in the past as well, it's a monofont isnt always mono kind of a thing. Especially when you add emojis or fancy kinds of unicode to the mix.

anopsy: image
looks great now! I'll push it then

FBruzzesi (approved):
Thanks a lot! I start to like this a lot!

NiceGIF


P.S. Test do not run because it is not a PR to `main` branch, but I can confirm that it is all good (locally):
image

anopsy:
Dude I've run those tests like 13 times, and also full tests twice (btw you can make lunch while they're running, it takes so long) . Thank you team for the support! I'm so excited

anopsy:
Could we also want some doctests? Or that's too boujee 😅

PR #660, opened 2024-05-09
chore: improving sklearn compatibility via `parametrize_with_checks`

FBruzzesi:
# Description

So the story goes as the following:

- The CI/CD fails for scikit-learn==1.5rc1 because of a change in the `check_estimator` internals
- In the [scikit-learn issue](https://github.com/scikit-learn/scikit-learn/issues/28966) I got a better picture of how to run test for compatible components
- In particular, in [rolling your own estimator](https://scikit-learn.org/dev/developers/develop.html#rolling-your-own-estimator) suggests to use [`parametrize_with_checks`](https://scikit-learn.org/dev/modules/generated/sklearn.utils.estimator_checks.parametrize_with_checks.html#sklearn.utils.estimator_checks.parametrize_with_checks), and of course I thought "that is a great idea to avoid dealing manually with each test"
- Say no more, I enter a rabbit hole to refactor all our tests - which would be fine
- Except that these tests failures helped me figure out a few missing parts in the codebase
- Therefore I started to address those missing bits, and that's how you end up with ~60+~ 70 files changes

## Current status of tests

- In `estimators`:
  - `DemographicParityClassifier` and `EqualOpportunityClassifier`: `n_features_in_` is missing (added those test to the skipping set)
  - `QuantileRegression(..., solver='L-BFGS-B')` fails when some sample_weights are zero.
- In `meta`:
  - `OrdinalClassifier` fails tests because we check that `y` has at least 3 classes, and I could not find a way to flag that to scikit-learn tests (skipping all tests)
  - `Thresholder` ~needs some more love~ added love, now it passes tests.


pyproject.toml
```diff
@@ -45,9 +45,9 @@ issue-tracker = "https://github.com/koaning/scikit-lego/issues"
 documentation = "https://koaning.github.io/scikit-lego/"
 
 [project.optional-dependencies]
-cvxpy = ["cmake", "osqp", "cvxpy>=1.1.8", "numpy<2.0"]
+cvxpy = ["cmake", "osqp", "cvxpy>=1.1.8"]
 formulaic = ["formulaic>=0.6.0"]
-umap = ["umap-learn>=0.4.6", "numpy<2.0"]
+umap = ["umap-learn>=0.4.6"]
```
FBruzzesi, line 50: That's the good news! They are both working with numpy 2.0rc

sklego/decomposition/pca_reconstruction.py
```diff
@@ -93,13 +93,9 @@ def fit(self, X, y=None):
         )
         self.pca_.fit(X, y)
         self.offset_ = -self.threshold
-        return self
 
-    def transform(self, X):
```
FBruzzesi, line 98: This method was throwing off the checks as they dynamically check for method names

sklego/decomposition/umap_reconstruction.py
```diff
@@ -44,6 +47,12 @@ class UMAPOutlierDetection(BaseEstimator, OutlierMixin):
         The offset used for the decision function.
     """
 
+    _parameter_constraints: dict = {
```
FBruzzesi, line 50: From my understanding this is the official way of validating arguments. We are not there yet, but for `n_components` there was no other way of forcing param setting

sklego/dummy.py
```diff
@@ -124,3 +124,6 @@ def allowed_strategies(self):
             DeprecationWarning,
         )
         return self._ALLOWED_STRATEGIES
+
+    def _more_tags(self):
+        return {"poor_score": True, "non_deterministic": True}
```
FBruzzesi, line 129: I love the `poor_score` tag 😂😂

sklego/linear_model.py
```diff
@@ -491,14 +492,19 @@ def fit(self, X, y):
...
         X, y = check_X_y(X, y, accept_large_sparse=False)
 
         sensitive = X[:, self.sensitive_col_idx_]
         if not self.train_sensitive_cols:
             X = np.delete(X, self.sensitive_col_idx_, axis=1)
+
         X = self._add_intercept(X)
+        self.n_features_in_ = (
+            X.shape[1] - self.fit_intercept
+        )  # + (1 - int(self.train_sensitive_cols)) * len(self.sensitive_col_idx_)
```
FBruzzesi, line 507: This is not necessarily true

sklego/meta/ordinal_classification.py
```diff
@@ -93,11 +93,13 @@ class OrdinalClassifier(MultiOutputMixin, ClassifierMixin, MetaEstimatorMixin, B
...
 
-    def __init__(self, estimator, *, n_jobs=None, use_calibration=False, **calibrarion_kwargs):
+    is_multiclass = True
+
+    def __init__(self, estimator, *, n_jobs=None, use_calibration=False, **calibration_kwargs):
         self.estimator = estimator
         self.n_jobs = n_jobs
         self.use_calibration = use_calibration
-        self.calibrarion_kwargs = calibrarion_kwargs
+        self.calibration_kwargs = calibration_kwargs
```
FBruzzesi, line 102: I am ashamed 🙈

sklego/decomposition/umap_reconstruction.py
koaning: ahh this does feel a bit nasty though, do we really need to import `Interval` from a hidden module? If so, are we 100% sure that this is the way sklearn wants to expose these tools?

tests/test_estimators/test_lad_regression.py
```diff
@@ -80,20 +78,3 @@ def test_fit_intercept_and_copy(coefs, intercept):
 def test_lad(test_fn):
     regr = LADRegression()
     test_fn(LADRegression.__name__, regr)
-
-
-@pytest.mark.parametrize("positive", [True, False])
```
FBruzzesi, line 85: It's already tested in `QuantileRegression`, no need to duplicate them

tests/test_estimators/test_quantile_regression.py
```diff
@@ -17,8 +18,9 @@
 
 def _create_dataset(coefs, intercept, noise=0.0):
     np.random.seed(0)
-    X = np.random.randn(10000, coefs.shape[0])
```
FBruzzesi, line 20: going from `10_000` to `1000` was a big achievement to reduce runtime of the solvers

sklego/meta/_grouped_utils.py
```diff
@@ -6,23 +6,22 @@
...
-
 
 def _split_groups_and_values(
     X, groups, name="", min_value_cols=1, check_X=True, **kwargs
 ) -> Tuple[pd.DataFrame, np.ndarray]:
     _data_format_checks(X, name=name)
-    _shape_check(X, min_value_cols)
+    check_array(X, ensure_min_features=min_value_cols, dtype=None)
 
+    print(groups)
```
koaning, line 16: Unneeded print

sklego/meta/ordinal_classification.py
koaning: oh my 😅

sklego/naive_bayes.py
```diff
@@ -264,6 +265,7 @@ def fit(self, X, y) -> "BayesianGaussianMixtureNB":
                 ).fit(subset_x[:, i].reshape(-1, 1), subset_y)
                 for i in range(X.shape[1])
             ]
+        self.n_iter_ = sum(sum(gmm.n_iter_ for gmm in gmm_c) for gmm_c in self.gmms_.values())
```
koaning, line 268: Isn't `np.sum` a bit faster here?

koaning (commented):
All the changes look sensible sofar, but I am wondering if we're perhaps a bit early when it comes to implementing all of this ... the more tags stuff feels a bit experimental. If you don't mind the experimental nature of it then that is fine, but it also feels fine to ignore for now.

sklego/naive_bayes.py
FBruzzesi: Honestly for such small numbers (`n_classes` x `n_features`) it's not worth it

sklego/linear_model.py
```diff
@@ -1128,15 +1131,16 @@ def __init__(
...
+                np.where(X @ params < y, self.quantile, 1 - self.quantile) * np.abs(y - X @ params),
+                weights=sample_weight,
             ) + self._regularized_loss(params)
 
         def grad_quantile_loss(params):
             return (
                 -(sample_weight * np.where(X @ params < y, self.quantile, 1 - self.quantile) * np.sign(y - X @ params))
                 @ X
-                / X.shape[0]
+                / sample_weight.sum()
```
FBruzzesi, line 1143: Hey @Garve, sorry to drag you into this mess.

No need for you to read all the changes. TL;DR is that in this PR I added tests using [scikit-learn parametrize_with_checks](https://scikit-learn.org/dev/modules/generated/sklearn.utils.estimator_checks.parametrize_with_checks.html#sklearn.utils.estimator_checks.parametrize_with_checks)  to check compatibility.

As some of the tests were failing I took a closer look and adjusted the formulas to take into account `sample_weight`. However we still have tests failing for _just_ one of the solvers. I wonder if you have any clue or insight on why this is the case.

sklego/mixture/bayesian_gmm_detector.py
```diff
@@ -162,11 +165,11 @@ def score_samples(self, X):
         if len(X.shape) == 1:
             X = np.expand_dims(X, 1)
 
-        return self.gmm_.score_samples(X) * -1
+        return self.gmm_.score_samples(X)
 
     def decision_function(self, X):
         # We subtract self.offset_ to make 0 be the threshold value for being an outlier:
-        return self.score_samples(X) + self.likelihood_threshold_
+        return self.score_samples(X) - self.offset_
```
FBruzzesi, line 172: Both this detector and gmm one required some (breaking changes) to comply to the tests and how scikit expects `score_samples`, `decision_function` and `predict` to interoperate

sklego/meta/thresholder.py
```diff
@@ -33,21 +38,35 @@ class Thresholder(BaseEstimator, ClassifierMixin):
...
     def _handle_refit(self, X, y, sample_weight=None):
         """Only refit when we need to, unless `refit=True` is present."""
         if self.refit:
-            self.estimator_ = clone(self.model)
-            self.estimator_.fit(X, y, sample_weight=sample_weight)
+            self._handle_unfitted(X, y, sample_weight)
         else:
             try:
-                _ = self.estimator_.predict(X[:1])
+                check_is_fitted(self.estimator_)
```
FBruzzesi, line 67: From the docs:

> If None, estimator is considered fitted if there exist an attribute that ends with a underscore and does not start with double underscore.

I think it's a good proxy. The reason for this is that for non-array data, `X is not subscriptable`

sklego/linear_model.py
koaning: If it's really just one solver I would just not sweat it. We should document it for sure but we can always patch that later.

tests/test_estimators/test_lad_regression.py
FBruzzesi: Edit: Actually the entire test can be considered a duplicate

sklego/linear_model.py
FBruzzesi: Apparently, for ubuntu and windows 'TNC' as well is failing, but only for `quantile=0.3` and `fit_intercept=False`, not in the other cases.
I am skipping that as well and improving the warning in the class description

tests/test_estimators/test_quantile_regression.py
```diff
@@ -17,30 +18,38 @@
...
+@parametrize_with_checks(
+    [
+        QuantileRegression(**dict(zip(["quantile", "positive", "fit_intercept", "method"], args)))
+        for args in product([0.5, 0.3], [True, False], [True, False], ["SLSQP", "TNC", "L-BFGS-B"])
+    ]
+)
+def test_sklearn_compatible_estimator(estimator, check):
+    if (
+        check.func.__name__ == "check_sample_weights_invariance"
+        # and signature(check.func.__dict__["__wrapped__"]).parameters["kind"] == "kind='zeros'"
```
FBruzzesi, line 37: I tried to dynamically find only the zero case, but I couldn't manage?!

FBruzzesi:
@koaning actually umap is failing on --pre with numpy==2.0rc**2**
[proof](https://github.com/FBruzzesi/scikit-lego/actions/runs/9096983379/job/25003746569)

koaning:
Can't we add that numpy req to the umap requirements for the time being?

FBruzzesi:
@koaning I pinned back `numpy<2.0` for the umap extra dependency.
Yesterday I learned that it's not supposed to be compatible yet anyway ([comment](https://github.com/lmcinnes/pynndescent/pull/242#issuecomment-2112733917))

koaning (approved):
lgtm!

PR #665, opened 2024-05-11
Make RegressionOutlier dataframe-agnostic

MarcoGorelli:
# Description

Fairly simple 👌 

Fixes #(issue)

## Type of change

## Checklist:

### DIFF
--- a/sklego/meta/regression_outlier_detector.py
+++ b/sklego/meta/regression_outlier_detector.py
@@ -131,7 +131,7 @@ class RegressionOutlierDetector(BaseEstimator, OutlierMixin):
         """
         X = nw.from_native(X, eager_only=True, strict=False)
         self.idx_ = np.argmax([i == self.column for i in X.columns]) if isinstance(X, nw.DataFrame) else self.column
-        X = check_array(X, estimator=self)
+        X = check_array(nw.to_native(X, strict=False), estimator=self)
         if not self._is_regression_model():
             raise ValueError("Passed model must be regression!")
         X, y = self.to_x_y(X)
--- a/sklego/meta/regression_outlier_detector.py
+++ b/sklego/meta/regression_outlier_detector.py
@@ -43,7 +43,7 @@ class RegressionOutlierDetector(BaseEstimator, OutlierMixin):
     Supported dataframes are:
 
     - pandas
-    - Polars (eager or lazy)
+    - Polars (eager)
     - Modin
     - cuDF
 
@@ -129,7 +129,7 @@ class RegressionOutlierDetector(BaseEstimator, OutlierMixin):
         ValueError
             If the `model` is not a regression estimator.
         """
-        X = nw.from_native(X, strict=False)
+        X = nw.from_native(X, eager_only=True, strict=False)
         self.idx_ = np.argmax([i == self.column for i in X.columns]) if isinstance(X, nw.DataFrame) else self.column
         X = check_array(X, estimator=self)
         if not self._is_regression_model():
--- a/sklego/meta/regression_outlier_detector.py
+++ b/sklego/meta/regression_outlier_detector.py
@@ -1,5 +1,5 @@
+import narwhals as nw
 import numpy as np
-import pandas as pd
 from sklearn.base import BaseEstimator, OutlierMixin
 from sklearn.utils.validation import check_array, check_is_fitted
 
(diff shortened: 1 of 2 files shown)


sklego/meta/regression_outlier_detector.py
```diff
@@ -112,7 +129,8 @@ def fit(self, X, y=None):
         ValueError
             If the `model` is not a regression estimator.
         """
-        self.idx_ = np.argmax([i == self.column for i in X.columns]) if isinstance(X, pd.DataFrame) else self.column
+        X = nw.from_native(X, strict=False)
```
MarcoGorelli, line 132: the question of how expensive this is came up on the livestream

on my laptop, it's about 1-2 microseconds

FBruzzesi: Can we pitch that as it is significantly less than converting to pandas 😁?

sklego/meta/regression_outlier_detector.py
```diff
@@ -112,7 +129,8 @@ def fit(self, X, y=None):
         ValueError
             If the `model` is not a regression estimator.
         """
-        self.idx_ = np.argmax([i == self.column for i in X.columns]) if isinstance(X, pd.DataFrame) else self.column
+        X = nw.from_native(X, eager_only=True, strict=False)
+        self.idx_ = np.argmax([i == self.column for i in X.columns]) if isinstance(X, nw.DataFrame) else self.column
         X = check_array(X, estimator=self)
```
FBruzzesi, line 134: How does `check_array` behave on narwhals frame? Converts it to numpy array?

FBruzzesi (approved):
Easy peasy! 👌
I left a question for my understanding of the (internal) behavior

sklego/meta/regression_outlier_detector.py
MarcoGorelli: oh yeah! even for just a small <1000 row dataframe, it takes 700-1200 times longer (depending on whether you use pyarrow)

and that's without counting that converting to pandas would break polars' optimisations, use much more memory, "force" users to have pandas+pyarrow as dependencies...I think we can sell this well :)

sklego/meta/regression_outlier_detector.py
MarcoGorelli: That's a good point actually, thanks for having asked! I just digged into it, but admittedly I should have checked it more carefully in the first place

Yes, just like for pandas/polars input directly, it converts to a numpy array. BUT - they have some pandas-specific logic in there. So, we can just pass `nw.to_native(X, strict=False)`, and then we can be sure there'll be no difference with respect to the status quo

sklego/meta/regression_outlier_detector.py
```diff
@@ -112,8 +129,9 @@ def fit(self, X, y=None):
         ValueError
             If the `model` is not a regression estimator.
         """
-        self.idx_ = np.argmax([i == self.column for i in X.columns]) if isinstance(X, pd.DataFrame) else self.column
-        X = check_array(X, estimator=self)
+        X = nw.from_native(X, eager_only=True, strict=False)
+        self.idx_ = np.argmax([i == self.column for i in X.columns]) if isinstance(X, nw.DataFrame) else self.column
+        X = check_array(nw.to_native(X, strict=False), estimator=self)
```
FBruzzesi, line 134: I assume now that input is force to be anything compatible with `check_array` itself - I have no clue if Modin and CuDF make the cut.

MarcoGorelli: let's check:

 ✅ modin:
```python
In [8]: import modin.pandas as pd

In [9]: pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]}).__array__()
UserWarning: Distributing <class 'dict'> object. This may take some time.
Out[9]:
array([[1, 4],
       [2, 5],
       [3, 6]])
```

🚫 cuDF
cuDF:
```python
df = cudf.DataFrame({'a': [1,2,3], 'b': [4,5,6]})
check_X_y(df, df['a'])
```
```
TypeError: Implicit conversion to a host NumPy array via __array__ is not allowed, To explicitly construct a GPU matrix, consider using .to_cupy()
To explicitly construct a host matrix, consider using .to_numpy().
```

I just assumed they would work, but surprised that cuDF disallows it - thanks for doing due diligence here, I've removed that one from the docstring

FBruzzesi:
Grazie Marco, amazing work! 
Sorry for delegating the checks but I don't have a GPU setup ready 😁

PR #666, opened 2024-05-11
patch: `ZeroInflatedRegressor` fix training indices

### LINKED ISSUE #664: [BUG] zero_inflated_regressor.py
### The Bug
- The [ZeroInflatedRegressor](https://github.com/koaning/scikit-lego/blob/main/sklego/meta/zero_inflated_regressor.py) class consists of a classifier and a regressor and it says in the documentation that "The regressor is only trained on examples **where the target is non-zero**, which makes it easier for it to focus." In the code, however, the regressor is actually trained on examples **where the classifier-predicted target is non-zero**.

### How to Fix the Bug
- Just replace line 111 with the following code:
```
non_zero_indices = np.where(y != 0)[0]
```

FBruzzesi:
# Description

Issue reported in #664 is very reasonable. I tried to take a look at when `ZeroInflatedRegressor` was introduced in #442, but I couldn't find any related discussion on such choice.

Fix #664 

## Type of change

### DIFF
--- a/sklego/meta/zero_inflated_regressor.py
+++ b/sklego/meta/zero_inflated_regressor.py
@@ -108,7 +108,7 @@ class ZeroInflatedRegressor(BaseEstimator, RegressorMixin):
                 logging.warning("Classifier ignores sample_weight.")
                 self.classifier_.fit(X, y != 0)
 
-        non_zero_indices = np.where(self.classifier_.predict(X) == 1)[0]
+        non_zero_indices = np.where(y != 0)[0]
 
         if non_zero_indices.size > 0:
             try:


sklego/meta/zero_inflated_regressor.py
```diff
@@ -108,7 +108,7 @@ def fit(self, X, y, sample_weight=None):
                 logging.warning("Classifier ignores sample_weight.")
                 self.classifier_.fit(X, y != 0)
```
FBruzzesi, line 109: Just as a side note after taking a look at the internals: as `classifier_` is supposed to classify zeros, wouldn't make sense to  fit on `y==0`? If so, that's a total of 3 lines that need changes. Not big deal anyway

koaning (approved):

PR #667, opened 2024-05-11
feat: Make grouped and hierarchical dataframe-agnostic

FBruzzesi:
# Description

Addresses Grouped* and Hierarchical, following up to the comment I added in the [thread](https://github.com/koaning/scikit-lego/issues/658). I will paste it for easier visibility:

> I am having a hard time to think of another way of implementing meta estimators that use pandas just for it's grouping functionalities. Namely `HierarchicalPredictor`s and `Grouped*` are not necessarily taking a dataframe input, but use pandas groupby's to iterate over different batch of data.
>
> - Would this be possible using pure numpy so that we don't need the pandas dependency? yes
> - Would this be significantly more inconvenient to maintain? absolutely yes
>
> My opinion/suggestion is to leave them as they are in their logic.
>
> If anything, we can allow arbitrary dataframe types as input, and use Narwhals to convert them to pandas and let the rest flow as is.


pyproject.toml
```diff
@@ -63,6 +63,7 @@ docs = [
 
 test-dep = [
     "narwhals[polars]",
+    "pyarrow",
```
FBruzzesi, line 66: Required to convert from polars to pandas

sklego/meta/hierarchical_predictor.py
```diff
@@ -252,20 +253,17 @@ def fit(self, X, y=None):
...
-        #     else
-        # )
 
         # If invalid shrinkage, will raise ValueError (before fitting all the estimators!)
         self.shrinkage_function_ = self._set_shrinkage_function()
 
+        X = nw.from_native(X, strict=False, eager_only=True)
         frame = (
-            pd.DataFrame(X)
+            (X.to_pandas() if isinstance(X, nw.DataFrame) else pd.DataFrame(X))
```
FBruzzesi, line 266: These parenthesis are **very** important, everything breaks otherwise 😂

tests/test_meta/test_hierarchical_predictor.py
```diff
@@ -38,14 +42,18 @@ def make_hierarchical_dataset(task):
...
-        g_1=["A"] * (n_samples // 2) + ["B"] * (n_samples // 2),
-        g_2=["X"] * (n_samples // 4) + ["Y"] * (n_samples // 2) + ["Z"] * (n_samples // 4),
+    X_ = (
+        pd.DataFrame(X, columns=[f"x_{i}" for i in range(X.shape[1])])
+        .assign(
+            g_0=1,
+            g_1=["A"] * (n_samples // 2) + ["B"] * (n_samples // 2),
+            g_2=["X"] * (n_samples // 4) + ["Y"] * (n_samples // 2) + ["Z"] * (n_samples // 4),
+        )
+        .pipe(frame_func)
```
FBruzzesi, line 52: @MarcoGorelli here is where I encountered the [`with_columns` issue](https://github.com/narwhals-dev/narwhals/issues/141). My first intention was to do it as:

```py
    (
     frame_func(dict(zip([f"x_{i}" for i in range(X.shape[1])]), X.T)
     .pipe(nw.from_native)
     .with_columns(
            g_0=1,
            g_1=["A"] * (n_samples // 2) + ["B"] * (n_samples // 2),
            g_2=["X"] * (n_samples // 4) + ["Y"] * (n_samples // 2) + ["Z"] * (n_samples // 4),
        )
        .pipe(nw.to_native)
```

MarcoGorelli:
🤔 are you sure about adding pyarrow as required? I think we should be able to solve this one

>  but use pandas groupby's to iterate over different batch of data.

Narwhals' `DataFrame.GroupBy` also lets you iterate over groups - I'll take a closer look at what's required, but this seems feasible

FBruzzesi:
@MarcoGorelli oh I see what you mean! Regarding hierarchical I am 100% sure it is doable, I can do it tomorrow myself.

For Grouped, internals are a bit messy, but I will give it a try.

tests/test_meta/test_hierarchical_predictor.py
MarcoGorelli: I think Polars doesn't do what you're expecting it to here 😉 
```python
In [34]: frame = pl.DataFrame({"a": [1,2,3]})

In [35]: frame.with_columns(b=['a', 'b', 'c'])
Out[35]:
shape: (3, 2)
┌─────┬─────────────────┐
│ a   ┆ b               │
│ --- ┆ ---             │
│ i64 ┆ list[str]       │
╞═════╪═════════════════╡
│ 1   ┆ ["a", "b", "c"] │
│ 2   ┆ ["a", "b", "c"] │
│ 3   ┆ ["a", "b", "c"] │
└─────┴─────────────────┘
```

I _think_ what we need is a Series constructor. Something like
```python
df.with_columns(
    g_1=nw.Series(
        ["A"] * (n_samples // 2) + ["B"] * (n_samples // 2),
        implementation=nw.get_implementation(X),
    )
)
```
?

FBruzzesi: Maybe we should move this conversation to the [issue](https://github.com/narwhals-dev/narwhals/issues/141) 😇

> I think Polars doesn't do what you're expecting it to here 😉

You are right, but it does with numpy.

> We need to pass something to nw.Series, otherwise it doesn't know which library's Series to create

Doesn't `with_columns` has some context (from self)? (As mentioned on Discord I need to catch up a lot with the internals, black magic to me)

MarcoGorelli: >  it does with numpy.

Just pushed out a release which can handle numpy 😎 It may take a few minutes to appear

> Wouldn't with_columns has some context (from self)?

Yeah, I'm just trying to figure out how to stay compatible with the Polars API. Because if you want to create a new column with `[1, 2, 3]` in Polars, you'd do
```
df.with_columns(d=pl.Series([1,2,3]))
```
So I think we need to instruct the user to construct a `nw.Series`, in such a way that the series can exist in a self-standing way without the `df.with_columns` context. I'm not really sure here...will sleep on it

But the numpy array case is easy, so let's start with that 😇

sklego/meta/grouped_transformer.py
```diff
@@ -1,9 +1,10 @@
+import narwhals as nw
```
FBruzzesi, line 1: @MarcoGorelli I reframed this to rely solely on narwhals.
There are a few missing bits still:

- in `_grouped_utils` there are some useful checks (again on datatypes)
- ln 110 is somewhat problematic in the tests because it is passing a native series without converting, but it would require some more logic for the None case (kinda related to [narwhals#144](https://github.com/narwhals-dev/narwhals/pull/144))

Finally, `__transform_groups` method, I am not sure there is an alternative way without using some sort of index trick, due to the fact that the original ordering needs to be maintained.

Let me know what you think, but I would say that we are quite close to a solution

MarcoGorelli: wow, amazing! exciting stuff

I'll add the equivalent of `X_group.select_dtypes(include="number")` and `X_group.select_dtypes(exclude="number")` then, anything else in Narwhals you'd like prioritising please let me know

FBruzzesi: @MarcoGorelli matter of fact, I don't see why we only check for numbers. I solved this as:

```py
X.select(nw.col(groups).is_null().any()).to_numpy().squeeze().any()
```

FBruzzesi:
I am really pushing the boundaries for `GroupedPredictor`, I think within another swipe I can finish this.

Current status is:

- `GroupedTransformer` passes all the tests, except one in which the group index is provided as negative (honestly, who would do that 😂)
- `HierarchicalPredictor` is failing only one test as well (`tests/test_meta/test_hierarchical_predictor.py::test_expected_output`), but only for the polars case, which could be due to the fact that the estimator cannot deal with polars?
- `GroupedPredictor` needs some more patches to close the deal.

sklego/meta/hierarchical_predictor.py
```diff
@@ -332,11 +350,11 @@ def _predict_estimators(self, X, method_name):
 
     def _fit_single_estimator(self, grp_frame):
         """Shortcut to fit an estimator on a single group"""
-        _X = grp_frame.drop(columns=self.groups_ + [self._TARGET_NAME])
-        _y = grp_frame[self._TARGET_NAME]
+        _X = nw.to_native(grp_frame.drop([*self.groups_, self._TARGET_NAME]))
+        _y = nw.to_native(grp_frame.select(self._TARGET_NAME))
         return clone(self.estimator).fit(_X, _y)
```
FBruzzesi, line 357: Here I am ending up with the following issue in case of polars.

If _X is empty (e.g. the dummy case  in `test_expected_output` test), while for pandas dropping all the columns, still maintains the original shape (I guess because index is preserved?), for polars we end up with `_X.shape` being (0, 0).

Which then leads to the estimator raising:

> ValueError: Found input variables with inconsistent numbers of samples

We could say this is kind of an edge case, but users use this as I remember a related issue (and Vincent tutorial on it).

@MarcoGorelli do you have any ideas on how to deal with this?

MarcoGorelli: just to have more context, do you happen to remember which issue it was? or the tutorial? I searched "hierarchical empty" here and came out with nothing

numpy does allow for things like `np.zeros((6, 0))`, and I presume things are getting converted to numpy at some point here anyway? So maybe, making an empty numpy array with a given shape somewhere can be a solution?

(I haven't yet looked into this code so sorry if this doesn't make sense)

koaning: I may also need a reminder 😅 what tutorial are you recalling here?

FBruzzesi: In issue #573 someone reported that specific case

FBruzzesi:
Fixed all the issues I had in the conversion and marking this as ready to review. I am quite happy with the results 🙌🏼🙌🏼

The only **breaking change** from previous version/implementation is the unfeasibility of providing negative indices in `groups`.

cc: @koaning @MarcoGorelli

FBruzzesi:
That's a nice trick! Thanks Marco! Pushing it now

FBruzzesi:
This should be ready for review. Once we merge into `narwhals-development` then I would say that we covered large majority of #658 and we should do a release. @koaning WDYT?

koaning (approved):
I had been glancing these changes over the last two weeks but have not seen anything that stands out, or at least ... I recall making some comments but these have all since been adressed.

I saw we push!

PR #668, opened 2024-05-12
WIP Make Timegapsplit dataframe-agnostic

MarcoGorelli:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.

Fixes #(issue)

## Type of change

## Checklist:


sklego/model_selection.py
```diff
@@ -167,31 +183,28 @@ def split(self, X, y=None, groups=None):
...
-                    < current_date + self.train_duration + self.valid_duration + self.gap_duration
-                )
-            ]
+            X_train_df = X_index_df.filter(
+                nw.col("__date__") >= start_date, nw.col("__date__") < current_date + self.train_duration
+            )
+            X_valid_df = X_index_df.filter(
+                nw.col("__date__") >= current_date + self.train_duration + self.gap_duration,
+                nw.col("__date__") < current_date + self.train_duration + self.valid_duration + self.gap_duration,
+            )
```
MarcoGorelli, line 192: I find this diff so pleasing 😍

sklego/model_selection.py
```diff
@@ -82,11 +99,7 @@ def __init__(
         if (train_duration is not None) and (train_duration <= gap_duration):
             raise ValueError("gap_duration is longer than train_duration, it should be shorter.")
 
-        if not date_serie.index.is_unique:
-            raise ValueError("date_serie doesn't have a unique index")
```
MarcoGorelli, line 86: this is checked as part of `nw.maybe_align_index`

FBruzzesi (approved):
Another very satisfying migration 🙌🏼
I left only one pedantic question 🙃

sklego/model_selection.py
```diff
@@ -65,6 +68,20 @@ class TimeGapSplit:
...
+    Notes
+    -----
+    Native cross-dataframe support is achieved using
+    [Narwhals](https://narwhals-dev.github.io/narwhals/){:target="_blank"}.
+    Supported dataframes are:
+
+    - pandas
+    - Polars (eager)
+    - Modin
+
```
FBruzzesi, line 81: Is CuDF breaking for some reason? It seems we are not doing any checking on X (`check_array` I am looking at you)

MarcoGorelli: again, thanks for asking - indeed, cuDF should be in the list, as `to_numpy()` is called explicitly here, and they do support that https://docs.rapids.ai/api/cudf/nightly/user_guide/api_docs/api/cudf.dataframe.to_numpy/

it's `__array__` (which gets triggered when you call `np.asarray` on an object) which they don't support they don't support https://github.com/koaning/scikit-lego/pull/665#discussion_r1597450090 . So, indeed, they should be included here

PR #676, opened 2024-06-02
perf: make sure X.schema is only calculated once per dataframe in TypeSelector

MarcoGorelli:
# Description

Refactors `TypeSelector` to use Narwhals selectors, as opposed to hard-coding types

This is preferable because, for Polars LazyFrames, getting the schema isn't a free operation (especially if the original dataframe is on the cloud - `LazyFrame.schema` may get renamed to `LazyFrame.collect_schema` for this reason)

I've also expanded the tests

## Type of change

## Checklist:

### DIFF
--- a/sklego/preprocessing/pandastransformers.py
+++ b/sklego/preprocessing/pandastransformers.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
-import functools
 import warnings
+from typing import Any
 
 import narwhals as nw
 from narwhals.dependencies import get_pandas
@@ -11,20 +11,31 @@ from sklearn.utils.validation import check_is_fitted
 from sklego.common import as_list
 
 
-def _nw_match_dtype(selection):
+def _nw_match_dtype(dtype, selection):
     if selection == "number":
-        return nw.selectors.numeric()
+        return dtype in (
+            nw.Int64,
+            nw.Int32,
+            nw.Int16,
+            nw.Int8,
+            nw.UInt64,
+            nw.UInt32,
+            nw.UInt16,
+            nw.UInt8,
+            nw.Float64,
+            nw.Float32,
+        )
     if selection == "bool":
-        return nw.selectors.boolean()
+        return dtype == nw.Boolean
     if selection == "string":
-        return nw.selectors.string()
+        return dtype == nw.String
     if selection == "category":
-        return nw.selectors.categorical()
+        return dtype == nw.Categorical
     msg = f"Expected {{'number', 'bool', 'string', 'category'}}, got: {selection}, which is not (yet!) supported."
     raise ValueError(msg)
 
 
-def _nw_select_dtypes(df, include: str | list[str], exclude: str | list[str]):
+def _nw_select_dtypes(df, include: str | list[str], exclude: str | list[str], schema: dict[str, Any]):
     if not include and not exclude:
         raise ValueError("Must provide at least one of `include` or `exclude`")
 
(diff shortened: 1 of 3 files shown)


sklego/preprocessing/pandastransformers.py
```diff
@@ -42,17 +32,17 @@ def _nw_select_dtypes(df, include: str | list[str], exclude: str | list[str]):
...
+        include = [_nw_match_dtype(x) for x in include]
+    else:
+        include = [nw.selectors.all()]
+    if exclude:
+        exclude = [_nw_match_dtype(x) for x in exclude]
+    else:
+        exclude = []
+    if exclude:
+        return df.select(functools.reduce(lambda x, y: x | y, include) - functools.reduce(lambda x, y: x | y, exclude))
+    return df.select(functools.reduce(lambda x, y: x | y, include))
```
MarcoGorelli, line 45: unfortunately Polars doesn't have an empty selector, else this could've been even simpler

sklego/preprocessing/pandastransformers.py
```diff
@@ -10,26 +11,15 @@
 from sklego.common import as_list
 
 
-def _nw_match_dtype(dtype, selection):
+def _nw_match_dtype(selection):
```
FBruzzesi, line 14: Big +1 here 👌

sklego/preprocessing/pandastransformers.py
FBruzzesi: That would have been so much cleaner!

FBruzzesi (commented):
LGTM!

I have one concern thought: I realized that polars selectors _select_ in the order they are given and do not maintain the original order.

Consider the following example:

```py
import polars
import polars.selectors as cs

df = pl.DataFrame({
    "a": list("abc"),
    "x": [1,2,3],
    "z": [True, False, True],
})

df.select(cs.numeric() | cs.string())
```
```
shape: (3, 2)
┌─────┬─────┐
│ x   ┆ a   │
│ --- ┆ --- │
│ i64 ┆ str │
╞═════╪═════╡
│ 1   ┆ a   │
│ 2   ┆ b   │
│ 3   ┆ c   │
└─────┴─────┘
```

Is this something we should warn the user or take into consideration?
cc: @koaning

MarcoGorelli:
> I have one concern thought: I realized that polars selectors select in the order they are given and do not maintain the original order.

that's a good point, thanks - it should be possible to preserve the order. `df.schema` needs calling here anyway - i'll update, making sure that `df.schema` is called no more than once per dataframe

sklego/preprocessing/pandastransformers.py
```diff
@@ -43,16 +44,22 @@ def _nw_select_dtypes(df, include: str | list[str], exclude: str | list[str]):
     if isinstance(exclude, str):
         exclude = [exclude]
 
-    include = include or ["string", "number", "bool", "category"]
```
MarcoGorelli, line 46: I don't think this logic was right to begin with. If `include` isn't passed, it should keep everything except for what's in `exclude` - not just the string/number/bool/category cols not in `exclude`...
I've added a test which covers this

sklego/preprocessing/pandastransformers.py
```diff
@@ -34,7 +35,7 @@ def _nw_match_dtype(dtype, selection):
     raise ValueError(msg)
 
 
-def _nw_select_dtypes(df, include: str | list[str], exclude: str | list[str]):
+def _nw_select_dtypes(include: str | list[str], exclude: str | list[str], schema: dict[str, Any]):
```
MarcoGorelli, line 38: rather than passing `df` and repeatedly calling `df.schema`, we can just calculate `schema` once upfront and pass that

sklego/preprocessing/pandastransformers.py
```diff
@@ -377,14 +382,17 @@ def transform(self, X):
...
                 raise ValueError(
                     f"Column dtypes were not equal during fit and transform. Fit types: \n"
                     f"{self.X_dtypes_}\n"
                     f"transform: \n"
                     f"{X.schema}"
                 )
-            transformed_df = _nw_select_dtypes(X, include=self.include, exclude=self.exclude)
+            transformed_df = X.select(
+                _nw_select_dtypes(include=self.include, exclude=self.exclude, schema=X_schema)
+            ).pipe(nw.to_native)
```
MarcoGorelli, line 395: looks like we (well, probably I 😳 ) forgot the `to_native` part at the end

I've added a test which asserts that the output is actually in the native class 👍

MarcoGorelli (commented):
thanks for your review!

I don't think it's worth using selectors here, especially if you want to preserve the order of the columns and if `X.schema` needs calculating anyway

So instead, I've refactored the code to make sure that `X.schema` is only ever calculated once per dataframe, and is passed to `_nw_select_dtypes`. This is better for LazyFrames

I've also expanded test coverage

FBruzzesi (approved):
I probably lost the "ready for review" notification! Well... I finally made it 😁
It looks amazing! Thanks @MarcoGorelli 👌

PR #679, opened 2024-06-15
patch: FairClassifier regularization and fit intercept

### LINKED ISSUE #674: [BUG] Fairness regularization
# Description

While investigating a failing test on #671, I notice the following lines in `_FairClassifier`:

```py
if self.penalty == "l1":
    log_likelihood -= cp.sum((1 / self.C) * cp.norm(theta[1:]))
```
However the default norm computed from cvxpy is p2 ([docs](https://www.cvxpy.org/api_reference/cvxpy.atoms.other_atoms.html#norm)).
Additionally `cp.norm` already sums the values over.

My suggestion to clean the API a bit is to:

- Allow "l2" regularization (currently cannot be specified)
- Correct the formula for "l1" regularization
- "none" should be None (?)

Finally, the `theta[1:]` should only be valid for `fit_intercept=True`, otherwise we should use `theta` "entirely".

Am I missing something?

FBruzzesi:
# Description

Addresses both problems presented in #674, namely:

- Regularization implementation
- Fit intercept not returning if false

Fixes #674 

Opening as draft since the issue has not been discussed

## Type of change

## Checklist:

### DIFF
--- a/sklego/linear_model.py
+++ b/sklego/linear_model.py
@@ -502,15 +502,25 @@ class _FairClassifier(BaseEstimator, LinearClassifierMixin):
 
         self.sensitive_col_idx_ = self.sensitive_cols
         X = nw.from_native(X, eager_only=True, strict=False)
+
         if isinstance(X, nw.DataFrame):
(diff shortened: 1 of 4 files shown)


koaning (approved):
LGTM!

PR #680, opened 2024-06-15
feat: `ZeroInflatedRegressor.score_samples(...)`

### LINKED ISSUE #585: [FEATURE] Implement predict proba in ZeroInflatedRegressor()
Please explain clearly what you'd like to see added. 

- [x] convince us of the use-case, we're open to many suggestions but we prefer to solve problems with pipelines that are at least somewhat general:

In any risk estimation task usually estimation of risk is calculated as probability of the situation happening x expected value of the situation happening. For instance, one can calculate the probability of a natural disaster (tornado) happening and the economical damage that this tornado will do i.e (1M$). Usually these kind of problems are solved via 2 part models: **https://www.sciencedirect.com/science/article/pii/S2212094718301385**. It is not the same a predicted property damage of 100000$ with a probability of tornado of 0.96 that a predicted property damage of 401000$ with a probability of 0.1.

- [x] add a screenshot if applicable (ML stuff is hard to explain with words, pictures say 1000 words)

Currently when predicting with the ZeroInflatedRegressor:

image

The prediction is made via:

`np.where(self.classifier_.predict(X))[0]`

I propose that we implement a predict_proba() for the classifier and then an argument to the predict() function to choose if we want absolute predictions or risk predictions (probability * expected impact).

This will actually make compatible the ZeroInflatedRegressor with the new [CalibratedClassifierCV](https://scikit-learn.org/stable/modules/generated/sklearn.calibration.CalibratedClassifierCV.html)

- [x] make sure that the feature you want is not already supported by sklearn

I guess that the alternative will be to pass the calibrated classifier already fitted to the ZeroInflatedRegressor.

FBruzzesi:
# Description

Implements `score_samples` for `ZeroInflatedRegressor` as discusses in #585 

Fixes #585 

## Type of change

## Checklist:

### DIFF
--- a/docs/user-guide/meta-models.md
+++ b/docs/user-guide/meta-models.md
@@ -388,9 +388,14 @@ ZIR (RFC+RFR) r²: 0.8579468997736154
 RFR r²: 0.7691291933110612
 ```
 
-If the underlying classifier is able to predict the _probability_ of a sample to be zero (i.e. it implements a `predict_proba` method), then the `ZeroInflatedRegressor` can be used to predict the probability of a sample being non-zero times the expected value of such sample.
+If the underlying classifier is able to predict the _probability_ of a sample to be zero (i.e. it implements a `predict_proba` method), then the `ZeroInflatedRegressor` can be used to predict the probability of a sample being non-zero _times_doc the expected value of such sample.
 
 This quantity is sometimes called _risk estimate_ or _expected impact_, however, to adhere to scikit-learn convention, we made it accessible via the  `score_samples` method.
+
+!!! warning "About `predict_proba`"
+    The `predict_proba` method of the classifier does not always return actual probabilities.
+
+    For this reason if you want to use the `score_samples` method, it is recommended to train with a classifier wrapped by the [`CalibratedClassifierCV`][calibrated-classifier-api] class from scikit-learn to calibrate the probabilities.
 
 ```py title="score_samples"
 --8<-- "docs/_scripts/meta-models.py:zero-inflated-score-samples"
--- a/docs/user-guide/meta-models.md
+++ b/docs/user-guide/meta-models.md
@@ -374,8 +374,8 @@ Sure, you can get regions where you are close to zero, but modelling an output o
 
 What we can do circumvent these problems is the following:
 
-1. Train a classifier to tell us whether the target is zero, or not.
-2. Train a regressor on all samples with a non-zero target.
+1. Train a **classifier** to tell us whether the target is zero, or not.
+2. Train a **regressor** on all samples with a non-zero target.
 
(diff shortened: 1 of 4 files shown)


tests/test_meta/test_zero_inflated_regressor.py
```diff
@@ -81,3 +82,47 @@ def test_wrong_estimators_exceptions():
...
+    X = np.random.randn(1_000, 4)
+    y = ((X[:, 0] > 0) & (X[:, 1] > 0)) * np.abs(X[:, 2] * X[:, 3] ** 2)
+
+    zir = ZeroInflatedRegressor(
+        classifier=RidgeClassifier(),
+        regressor=ExtraTreesRegressor(max_depth=20, random_state=0, n_jobs=-1),
+    ).fit(X, y)
+
+    with pytest.raises(AttributeError, match="This 'ZeroInflatedRegressor' has no attribute 'score_samples'"):
+        zir.score_samples(X)
```
FBruzzesi, line 127: Wondering if this is the best error message to give to the user - automatically generated from `available_if` decorator

docs/user-guide/meta-models.md
```diff
@@ -384,8 +384,21 @@ By putting these two together in an obvious way, we get the [`ZeroInflatedRegres
...
 ```console
-ZIR (RFC+RFR) r²: 0.8992404366385873
-RFR r²: 0.8516522752031502
+ZIR (RFC+RFR) r²: 0.8579468997736154
+RFR r²: 0.7691291933110612
+```
+
+If the underlying classifier is able to predict the _probability_ of a sample to be zero (i.e. it implements a `predict_proba` method), then the `ZeroInflatedRegressor` can be used to predict the probability of a sample being non-zero times the expected value of such sample.
+
+This quantity is sometimes called _risk estimate_ or _expected impact_, however, to adhere to scikit-learn convention, we made it accessible via the  `score_samples` method.
```
FBruzzesi, line 393: cc: @edgBR 

Is this description reasonable to you? Any input is more than welcomed

koaning: @edgBR reminder ping

edgBR: Hi guys, the description looks clear to me. However I will add that it can be called risk only if the classifier is calibrated.

FBruzzesi: I will add a warning to mention that classifier should be calibrated

koaning (approved):
Did not spot anything out of the ordinary, feels cool to merge to me. 

@FBruzzesi let me know if you merge this because it feels like it would be ready for a new release as well.

FBruzzesi:
I was out on vacation for the weekend. Merging now 😁

PR #687, opened 2024-07-11
feat: `RepeatingBasisFunction.inverse_transform`

Alex-Cremers:
# Description

This PR implements `get_feature_names_out()` for `_RepeatingBasisFunction`, and `RepeatingBasisFunction` (which inherits feature names from the former).

This PR also implements an inverse_transform for `_RepeatingBasisFunction` in passing. I did not include a more general implementation for `RepeatingBasisFunction` (as the one I use requires importing pandas), but the inverse_transform() can be accessed from the fitted transformer via `.pipeline_.named_transformers_['repeatingbasis'].inverse_transform()`). It's a rare use case, but it shouldn't affect other uses in any way, so I figured I'd include it. Note that the transformation is only invertible if the original values are in the input_range (upper bound excluded). Otherwise, the reconstructed values are only equal modulo the width of the range.

Two tests have been added to test_repeatingbasisfunction.py:

  - test that set_output(transform='pandas') works properly with `RepeatingBasisFunction`
  - test that the new `inverse_transform()` for `_RepeatingBasisFunction` truly recovers the original values (as long as they fall within the input_range).

Note that for feature names, ClassNamePrefixFeaturesOutMixin could also be used if `self.n_periods` was renamed to `self._n_features_out`, but I didn't see a simple way of keeping the original column name as prefix, so I adopted a solution perhaps more idiosyncratic.

It's a minor PR overall, so I didn't ping before submitting. I hope it's okay.

Progress on Issue #543

## Type of change

## Checklist:

- [X] My code follows the style guidelines (ruff)
- [X] I have made corresponding changes to the documentation (also to the readme.md)

### DIFF
--- a/sklego/preprocessing/repeatingbasis.py
+++ b/sklego/preprocessing/repeatingbasis.py
@@ -65,6 +65,10 @@ class RepeatingBasisFunction(TransformerMixin, BaseEstimator):
         self.n_periods = n_periods
         self.input_range = input_range
         self.width = width
+    
+    def get_feature_names_out(self, input_features=None):
+      feature_names = self.pipeline_.get_feature_names_out()
+      return feature_names
 
     def fit(self, X, y=None):
         """Fit `RepeatingBasisFunction` transformer on input data `X`.
@@ -82,6 +86,7 @@ class RepeatingBasisFunction(TransformerMixin, BaseEstimator):
         self : RepeatingBasisFunction
             The fitted transformer.
         """
+        
         self.pipeline_ = ColumnTransformer(
             [
                 (
@@ -95,6 +100,7 @@ class RepeatingBasisFunction(TransformerMixin, BaseEstimator):
                 )
             ],
             remainder=self.remainder,
+            verbose_feature_names_out = False
         )
 
         self.pipeline_.fit(X, y)
@@ -116,6 +122,31 @@ class RepeatingBasisFunction(TransformerMixin, BaseEstimator):
         """
         check_is_fitted(self, ["pipeline_"])
         return self.pipeline_.transform(X)
+      
+    def inverse_transform(self, X):
+        """Transform RBF features back to the input range. Outputs a numpy array. 
+
+        Parameters
+        ----------
+        X : 2D array-like of shape (n_samples, n_features)
+            Can be either a pandas.DataFrame or a numpy array.
+            The RBF columns to transform back must appear first.
+
+        Returns
+        -------
+        X_transformed : array-like with the reconstruction of the original values
+                        in input_range in the first column. Will be of the same type as X.
+        """
+        
+        check_is_fitted(self, ["pipeline_"])
+        
+        if isinstance(X,np.ndarray):
+          Xarr = check_array(X[:,:self.n_periods], estimator=self, ensure_2d=True)
+          new_x = self.pipeline_.named_transformers_['repeatingbasis'].inverse_transform(Xarr)
+          Xarr = np.hstack((new_x.reshape(-1, 1),X[:,self.n_periods:]))
+          return Xarr
+        else:
+          raise TypeError("X must be a numpy array.")
 
 
 class _RepeatingBasisFunction(TransformerMixin, BaseEstimator):
@@ -147,7 +178,10 @@ class _RepeatingBasisFunction(TransformerMixin, BaseEstimator):
         self.n_periods = n_periods
         self.input_range = input_range
         self.width = width
-
+    
+    def get_feature_names_out(self, input_features=None):
+      return [f"{input_features[0]}_rbf{str(i)}" for i in range(self.n_periods)]
+    
     def fit(self, X, y=None):
         """Fit the transformer to the input data and compute the basis functions.
 
@@ -163,6 +197,7 @@ class _RepeatingBasisFunction(TransformerMixin, BaseEstimator):
         self : _RepeatingBasisFunction
             The fitted transformer.
         """
+        
         X = check_array(X, estimator=self)
 
         # find min and max for standardization if not given explicitly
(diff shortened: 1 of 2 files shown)


Alex-Cremers:
Requesting review and approval for the tests, @koaning

koaning:
I am open to the inverse transform, but I think the implementation may change in the future in favor our scikit-learn pipelines. This transformer was added before the `SplineTransformer` was added to core scikit-learn and I might recommend folks to use that going forward. The `extrapolation="periodic"` setting basically covers the use-case here. We might just refer to that under the hood. 

@FBruzzesi figured I'd check with you, do you have any comments/opinions on this one? I am open to adding it, just wondering what might be most pragmatic given the sklearn estimator that now exists.

Alex-Cremers:
Thanks, I had missed the possibility to set SplineTransformer to periodic! The result is pretty similar indeed.

Now I see that sklearn also has an open issue regarding inverse transform on Spline transformer: https://github.com/scikit-learn/scikit-learn/issues/28551

FBruzzesi:
Hey @Alex-Cremers , thanks for the PR, it is much appreciated!

I can take a close look during the weekend, from a sneakpeak I would say that some linting is required. Every other technical comment will follow in the weekend 😂

@koaning 

> just wondering what might be most pragmatic given the sklearn estimator that now exists.

I am generally in favor to keep what we have bug free but not expanding (and maintaining) new features **if** they are already in scikit-learn - which seems not to be the case for `inverse_transform` , therefore I am open to have it implemented here

sklego/preprocessing/repeatingbasis.py
```diff
@@ -116,6 +122,31 @@ def transform(self, X):
         """
         check_is_fitted(self, ["pipeline_"])
         return self.pipeline_.transform(X)
+      
+    def inverse_transform(self, X):
+        """Transform RBF features back to the input range. Outputs a numpy array. 
+
```
FBruzzesi, line 128: Can we add what you mention in the description in the docstring Notes of the method?

In particular I am referring to:

> Note that the transformation is only invertible if the original values are in the input_range (upper bound excluded). Otherwise, the reconstructed values are only equal modulo the width of the range.

sklego/preprocessing/repeatingbasis.py
```diff
@@ -116,6 +122,31 @@ def transform(self, X):
...
+        Returns
+        -------
+        X_transformed : array-like with the reconstruction of the original values
+                        in input_range in the first column. Will be of the same type as X.
+        """
+        
+        check_is_fitted(self, ["pipeline_"])
+        
+        if isinstance(X,np.ndarray):
+          Xarr = check_array(X[:,:self.n_periods], estimator=self, ensure_2d=True)
```
FBruzzesi, line 144: I would be more comfortable with something along the following lines:

If line 144 becomes 

```py
Xarr = check_array(X, estimator=self, ensure_2d=True, ensure_min_features=self.n_periods)[:, :self.n_periods]
```

which also convert to array (some) dataframe-like objects and maybe it's even possible to avoid checking for array instance.

tests/test_preprocessing/test_repeatingbasisfunction.py
```diff
@@ -50,3 +50,20 @@ def test_when_rbf_helper_receives_more_than_one_col_raises_value_error(df):
...
+    Z = tf.fit(X, y).transform(X)
+    assert isinstance(Z, pd.core.frame.DataFrame)
+
+def test_inverse_transform(df):
+    X, y = df[["a", "b", "c", "d"]], df[["e"]]
+    tf = RepeatingBasisFunction(column=0, n_periods=4, input_range=(0,7), remainder="drop")
+    Z = tf.fit(X, y).transform(X)
+    assert np.allclose(
+      X["a"], 
+      tf.pipeline_.named_transformers_['repeatingbasis'].inverse_transform(Z),
```
FBruzzesi, line 67: This feels a bit too nested to me.

Why don't we 
```suggestion
      tf.inverse_transform(Z), 
```
?

They should suppose to return the same or am I missing something?

sklego/preprocessing/repeatingbasis.py
```diff
@@ -116,6 +122,31 @@ def transform(self, X):
...
+          new_x = self.pipeline_.named_transformers_['repeatingbasis'].inverse_transform(Xarr)
+          Xarr = np.hstack((new_x.reshape(-1, 1),X[:,self.n_periods:]))
```
FBruzzesi, line 146: Why are we returning the input as well? kind of related to the comment in tests

sklego/preprocessing/repeatingbasis.py
```diff
@@ -208,7 +243,27 @@ def transform(self, X):
 
         # apply rbf function to series for each basis
         return self._rbf(base_distances)
-
+    
+    def inverse_transform(self, X):
```
FBruzzesi, line 247: This is awesome ✨

FBruzzesi (commented):
Thanks again for the PR. I think it is a nice feature to have in here!

As first comment, please run the following to format and lint the code:

```py
python -m pip install ruff 
make lint
```

I think I am missing something on why `RepeatingBasisFunction` re-implements some logic in the `inverse_transform` and does not just call

```py
self.pipeline_.named_transformers_['repeatingbasis'].inverse_transform(X)
```

I feel like we are separating the logic while there is no real need for it - in fact they return different shape objects and I would not be expecting that.

Alex-Cremers:
> Thanks again for the PR. I think it is a nice feature to have in here!
> 
> As first comment, please run the following to format and lint the code:
> 
> ```python
> python -m pip install ruff 
> make lint
> ```
> 
> I think I am missing something on why `RepeatingBasisFunction` re-implements some logic in the `inverse_transform` and does not just call
> 
> ```python
> self.pipeline_.named_transformers_['repeatingbasis'].inverse_transform(X)
> ```
> 
> I feel like we are separating the logic while there is no real need for it - in fact they return different shape objects and I would not be expecting that.

Thanks for the thorough review!

Sorry for the lint, I only ran "ruff check" and it passed, so I didn't dig deeper!

Regarding the inverse_transform: I actually meant to remove it from RepeatingBasisFunction before submitting, precisely because it was a big confusing, but it turns out I only removed part of it. There are two reasons why it's more complicated than _RepeatingBasisFunction:

1.  I wanted it to deal with both pandas df and numpy arrays as possible inputs, and if a pandas, I wanted to retrieve the original column name (see this sklearn issue: https://github.com/scikit-learn/scikit-learn/pull/27891). I then removed the pandas part of the code to avoid adding an extra dependency, but it looks like I left the numpy part behind in the end.
2. Unlike _RepeatingBasisFunction which only deals with one column input / n_periods column outputs, RepeatingBasisFunction deals with a whole df/array, selects one of its columns as input, and might pass the other columns as well if `remainder='passthrough'`. Because of the passthrough possibility, the inverse transform should check for the presence of other columns and pass them back. In hindsight, I actually think that this is a bad idea with numpy arrays, because the order is lost (the transform/inverse_transform composition moves the target column first). With pandas it's fine, because the columns are named.

As far as I can see, the best would probably be to remove the inverse_transform from RepeatingBasisFunction entirely for now and wait for https://github.com/scikit-learn/scikit-learn/issues/11463 (inverse_transform for ColumnTransformer) and the other issue mentioned above to be solved, as there would then be a proper way to implement the inverse transform without relying on funny tricks. If a user really needs the inverse transform in the meantime, they can recover the function as I did in the test.

If you agree with this, I'll remove `RepeatingBasisFunction.inverse_transform()` and address the remaining comments.

PR #693, opened 2024-07-28
patch: fix estimator kwargs in signature

FBruzzesi:
# Description

Hotfix to properly cloning kwargs in estimator `__init__`. This is both a bug fix and a breaking change!

Fixes #692 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the style guidelines (ruff)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (also to the readme.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added tests to check whether the new feature adheres to the sklearn convention
- [x] New and existing unit tests pass locally with my changes

PR #694, opened 2024-08-07
feat: bump narwhals and adapt to support pyarrow

FBruzzesi:
# Description

Long story short: I knew that some functionalities would have not worked for pyarrow even if using narwhals and we had to adjust accordingly.

For example we cannot create a series using `native_namespace.Series([...])`, because pyarrow doesn't have a `Series` object, therefore the a workaround was needed:

```diff
- series = native_namespace.Series([...])
+ series = nw.from_dict({"_tmp": [...]})["_tmp"]
```
Namely, create a narhwals dataframe for the given namespace and then select the unique column.

A bump of other changes were needed in the tests to assess with pyarrow tables.

## Type of change

## Checklist:

## Other comments

- The commits are because I branched out from another PR 🙈
- Most likely there are other parts of the codebase in which a change may be required to make sure that pyarrow runs (to be continued)
- Adding @MarcoGorelli to the mix, to make sure there is no better way of going about this


sklego/meta/hierarchical_predictor.py
```diff
@@ -318,11 +321,14 @@ def _predict_estimators(self, X, method_name):
             X = nw.from_native(pd.DataFrame(X))
 
         n_samples = X.shape[0]
-        native_space = nw.get_native_namespace(X)
+        native_namespace = nw.get_native_namespace(X)
+        global_series = nw.from_dict({self._GLOBAL_NAME: [1] * n_samples}, native_namespace=native_namespace)[
```
koaning, line 325: Before putting it in narwhals we make a big list of ones which we put into a dictionary before it goes into narwhals. I wonder ... is there a faster way? My gut feeling is that `np.ones()` is a better idea here but I will gladly hear it if I am missing something.

koaning (commented):
Just one comment. If the comment clears it looks good to me!

koaning:
Oh, and one more thing, must we bump narwhals? I usually prefer not to force the user to use the latest and greatest dependency, but it seems like we need it to support more dataframe types?

sklego/meta/hierarchical_predictor.py
FBruzzesi: Checking if all tests will pass

FBruzzesi:
> Oh, and one more thing, must we bump narwhals? I usually prefer not to force the user to use the latest and greatest dependency, but it seems like we need it to support more dataframe types?

Sadly yes, if we want to have pyarrow support for grouped meta and shift, these functionalities made it only in the latest release (yet narwhals keeps being dependency-free)

FBruzzesi:
> LGTM! Might be good to prep another release?

Sounds good! We went a bit silent, but now have some breaking changes from #693

PR #698, opened 2024-08-21
Fix flaky tests

koaning:
@FBruzzesi the root cause of the flaky test was a broken link with openML. To remedy this, I went ahead and just added the dataset to the repo and added a loading fixture. This should be more stable going forward.


FBruzzesi:
Awesome! Thanks for fixing this so fast 🙌🏼

PR #708, opened 2024-10-29
chore: improve error message for `ZeroDivisionError` in `LowessRegression`

### LINKED ISSUE #707: [BUG] LowessRegressor: ZeroDivisionError: Weights sum to zero, can't be normalized
Using LowessRegressor, I received a "ZeroDivisionError: Weights sum to zero, can't be normalized" error.  A little debugging revealed that the problem is in _calc_wts:

```
    def _calc_wts(self, x_i):
        """Calculate the weights for a single point `x_i` using the training data `self.X_` and the parameters
        `self.sigma` and `self.span`. The weights are calculated as `np.exp(-(distances**2) / self.sigma)`,
        where distances are the distances between `x_i` and all the training samples.

        If `self.span` is not None, then the weights are multiplied by
        `(distances <= np.quantile(distances, q=self.span))`.
        """
        distances = np.linalg.norm(self.X_ - x_i, axis=1)
        weights = np.exp(-(distances**2) / self.sigma)
        if self.span:
            weights = weights * (distances <= np.quantile(distances, q=self.span))
        return weights
```

If sigma is not sized properly, the scaled distances will be large and cause np.exp() to return zeros for the weights, resulting in an exception being raised when the weights are applied.

This seems like the wrong behavior.  Perhaps the weights should be clipped to some minimum non-zero weight so that the predictor can still be used, perhaps with a warning message to explain the problem?

FBruzzesi:
# Description

Adds custom error message for for `ZeroDivisionError` in `LowessRegression`.

Closes #707

## Type of change

## Checklist:


tests/test_estimators/test_lowess.py
```diff
@@ -15,3 +18,15 @@ def test_obvious_usecase():
...
+def test_custom_error_for_zero_division():
+    x = np.arange(0, 100)
+    X = x.reshape(-1, 1)
+    y = np.ones(x.shape)
+    estimator = LowessRegression(sigma=1e-10).fit(X, y)
+
+    with pytest.raises(
+        ValueError, match=re.escape("Weights, resulting from `np.exp(-(distances**2) / self.sigma)`, are all zero.")
+    ):
+        estimator.predict(X[:10] + 0.5)
```
FBruzzesi, line 32: An offset is required, otherwise `X` to predict is same as `X` in `fit` method, leading to weight of 1 for the corresponding value.

koaning: Maybe add that in the codebase as a comment?

koaning (approved):
LGTM!

PR #712, opened 2024-11-06
Revert "Monotonicspline"

koaning:
Reverts koaning/scikit-lego#710


koaning:
Had a doubt last minute but saw that the docs code still worked.

PR #714, opened 2024-11-07
feat: add `handle_zero` option in `ZeroInflatedRegressor` estimator

mkalimeri:
Hi, please review my code for issue 480 - Run with given regressor instead of raising warning in ZeroInflatedRegressor

There are changes in two files
1) zero_inflated_regressor.py: As explained in my comment, I approached this issue in the following way
 - I added a flag 'handle_error', which can take two values: 'error' and 'ignore'. If the user chooses 'error', then, in the case that the train set output only consists of zeros, a ValueError is thrown. If the user chooses 'ignore' and if all the train set outputs are zero, the regressor fits all the train set (the flag handle_error='ignore' is taken into consideration only if there are no non-zero targets). So, when there are only zero targets and the user chooses handle_error='ignore', the regressor will train on the whole dataset, but if there are non-zero targets, the functionality remains as is (ignore lines with zero target)

- The handle_error='error' as default, so the user should actively choose to ignore the fact that the train dataset contains only zero targets.

2) test_zero_inflated_regressor.py: Added unit tests for the new functionality

Please let me know what you think and if you have any suggestions about how to improve the code!


koaning:
Before I dive deeper, could you check and confirm why the tests fail? Just want to make sure it's not an oversight.

mkalimeri:
Hi Vincent,

sorry about that, I was just looking at this. There is a mismatch in the
message of the ValueError thrown (see below).

Best regards,

Maria

[image: Screenshot 2024-11-07 at 14.04.12.png]

mkalimeri:
Hi again, sorry about that... I fixed it locally, shall I send another PR?

BR,

Maria

sklego/meta/zero_inflated_regressor.py
```diff
@@ -79,7 +79,10 @@ def fit(self, X, y, sample_weight=None):
             The target values.
         sample_weight : array-like of shape (n_samples, ), default=None
             Individual weights for each sample.
-
+        handle_zero : string, one of 'error' and 'ignore', default='error
```
FBruzzesi, line 82:
```suggestion
        handle_zero : Literal["error", "ignore"], default='error
```

sklego/meta/zero_inflated_regressor.py
```diff
@@ -79,7 +79,10 @@ def fit(self, X, y, sample_weight=None):
...
+            In the case that all train set output consists of zero values only, whether to ignor and continue ('ignore')
+            or to throw ValueError ('error)
```
FBruzzesi, line 84:
```suggestion
            How to behave in the case that all train set output consists of zero values only.
            
            - `handle_zero = 'error'`: will raise a `ValueError` (default)
            - `handle_zero = 'ignore'`: will continue to train the regressor on the entire dataset
```

sklego/meta/zero_inflated_regressor.py
```diff
@@ -88,7 +91,9 @@ def fit(self, X, y, sample_weight=None):
         Raises
         ------
         ValueError
-            If `classifier` is not a classifier or `regressor` is not a regressor.
+            If `classifier` is not a classifier
+            If`regressor` is not a regressor
```
FBruzzesi, line 95:
```suggestion
            If `regressor` is not a regressor
```

sklego/meta/zero_inflated_regressor.py
```diff
@@ -97,8 +102,14 @@ def fit(self, X, y, sample_weight=None):
...
             )
         if not is_regressor(self.regressor):
-            raise ValueError(f"`regressor` has to be a regressor. Received instance of {type(self.regressor)} instead.")
-
+            raise ValueError(
+                f"`regressor` has to be a regressor. Received instance of {type(self.regressor)} instead."
+            )
+        if not handle_zero in ['ignore', 'error']:
+            raise ValueError(
+                f" 'handle_zero' has to be one of 'ignore' or 'error'. Received {handle_zero} instead."
```
FBruzzesi, line 110: Can we use backticks as well?

```suggestion
                f"`handle_zero` has to be one of {'ignore', 'error'}. Received '{handle_zero}' instead."
```

sklego/meta/zero_inflated_regressor.py
```diff
@@ -88,7 +91,9 @@ def fit(self, X, y, sample_weight=None):
         Raises
         ------
         ValueError
-            If `classifier` is not a classifier or `regressor` is not a regressor.
+            If `classifier` is not a classifier
+            If`regressor` is not a regressor
+            If all train set output is 0 and we do not choose to ignore this
```
FBruzzesi, line 96:
```suggestion
            If all train target entirely consists of zeros and `handle_zero="error"`
```

sklego/meta/zero_inflated_regressor.py
```diff
@@ -112,9 +123,12 @@ def fit(self, X, y, sample_weight=None):
                 logging.warning("Classifier ignores sample_weight.")
                 self.classifier_.fit(X, y != 0)
 
-        non_zero_indices = np.where(y != 0)[0]
+        indices_for_training = np.where(y != 0)[0] # these are the non-zero indices
+        if (handle_zero == 'ignore') & (indices_for_training.size == 0): # if we choose to ignore that all train set output is 0
+            logging.warning("Regressor ignores that all train set output consists of zero values.")
```
FBruzzesi, line 128: I am not fully convinced by this warning message. I think we should be more verbose in stating what's happening.

What do you think of the following:

```suggestion
            logging.warning("Regressor will be training on `y` consisting of zero values only.")
```

FBruzzesi (commented):
Hey @mkalimeri , thanks for the contribution. Code is fine, I left a few comments mostly for documentation and error messages.

Additionally, could you please add a test that is not raising when training with y consisting of zeros only? i.e. with `handle_zero='ignore'`?

sklego/meta/zero_inflated_regressor.py
mkalimeri: Indeed, your edit is clearer, thank you for the suggestion

mkalimeri:
Thank you for the review, I accepted the edits as suggested and will add the unit test shortly

FBruzzesi (approved):
Hey @mkalimeri thanks a ton for the effort on this 🚀

I committed a couple of changes to move `handle_zero` argument to the estimator initialization instead of its `fit` method. Aside that your logic remained untouched!

We should be ready to merge 🙌🏼

mkalimeri:
Hi @FBruzzesi, thank you for improving the code and merging. This was fun! :-)

PR #718, opened 2024-11-14
ci: Add python 3.13 in test.yml

koaning:
Python 3.13, does it blend?


FBruzzesi:
After a quick search: the [issue related to py3.13](https://github.com/numba/numba/issues/9413) was closed last week. I would assume that starting from their next release this should not fail anymore

FBruzzesi (approved):
Added a couple of commits:

- one to exclude windows-py3.13 combo which is failing with a very cryptic error - for now I am quite happy with it running on the other 2 os
- removed trigger on `narwhals-development` branch 🙈

PR #720, opened 2024-11-20
Compatibility with scikit-learn 1.6rc1

### LINKED ISSUE #719: Transition to scikit-learn 1.6
scikit-learn [make transition from 1.5 to 1.6 easier for third-party library using scikit-learn utilities](https://github.com/scikit-learn/scikit-learn/issues/30298) will give us some buffer time to adjust to the new changes before they make it to a release. From other discussions it seems that these changes will appear in 1.7.

Before that happens we should decide how to proceed forward. I see the following options:

- Make a hard cut when scikit-learn 1.7 will come out, and support only version 1.6+
- Implement logic based on scikit-learn version
  - Coming from some narwhals learnings, this might actually be easier than expected and I would like to give it a try before rejecting the idea

FBruzzesi:
# Description

Diff is simpler than expected:

- Fix order of mixin's inheritance
- Added the new [`__sklearn_tags__`](https://scikit-learn.org/1.6/developers/develop.html#estimator-tags) to all the classes that changes the default tags
- Created a utility function in `sklego.common` that perform validation either via `validate_data` (scikit>=1.6) or `check_X_y` and `check_array` (scikit < 1.6). 

There are some checks which are not fully compatible, and one that might be buggy in this RC: `check_sample_weight_equivalence` seems to fail on every of our estimators that support `sample_weight` - it might be worth reporting upstream, but I could not pin the issue

Closes #719

### DIFF
--- a/sklego/__init__.py
+++ b/sklego/__init__.py
@@ -1,3 +1,4 @@
+import re
 import sys
 
 if sys.version_info >= (3, 8):
@@ -5,5 +6,8 @@ if sys.version_info >= (3, 8):
 else:
     import importlib_metadata as metadata
 
+
 __title__ = "sklego"
 __version__ = metadata.version("scikit-lego")
+
+SKLEARN_VERSION = tuple(int(re.sub(r"\D", "", str(v))) for v in metadata.version("scikit-learn").split("."))
--- a/sklego/common.py
+++ b/sklego/common.py
@@ -366,7 +366,7 @@ def validate_data(
         )
 
     else:
-        if y == "no_validation":
-            return check_array(arr=X, estimator=estimator, **check_params)
+        if y is None or (isinstance(y, str) and y == "no_validation"):
+            return check_array(X, estimator=estimator, **check_params)
         else:
             return check_X_y(X=X, y=y, estimator=estimator, **check_params)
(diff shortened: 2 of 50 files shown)


FBruzzesi:
So current tests are all passing, which is definitly a good sign that we did not break anything. I tested with scikit-learn1.6rc1 locally (python3.10, ubuntu) and successfully

koaning (approved):

koaning:
Whoa. So many changes. 

I guess the best check is to merge it now and to see if something breaks in the future, but nothing jumped out of the ordinary.

koaning:
Ah, I did the merge and now stuff seems breaking again 😓

koaning (changes requested):
Changing state to reflect some tests break.

koaning:
This is ... odd. The tests locally in codespaces run just fine but they break in CI.

FBruzzesi:
I will try to take a better look this weekend.. Should we pin scikit-learn version until we fix this?

koaning:
Ah, 1.6 is no longer a release candidate I think. 

Let's delay version pinning unless folks start to submit issues. It could be that we are worrying about details that aren't a huge worry for folks.

PR #723, opened 2024-12-05
Update identitytransformer.py

koaning:
The new scikit-learn version demands this.

### DIFF
--- a/sklego/preprocessing/identitytransformer.py
+++ b/sklego/preprocessing/identitytransformer.py
@@ -3,7 +3,7 @@ from sklearn.utils import check_array
 from sklearn.utils.validation import check_is_fitted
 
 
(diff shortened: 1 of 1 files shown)


koaning:
Yeah, the reason I started picking this up again is because of the influx of emails from Github. I will likely not do it all in one go, but rather trackle bugs one by one. Trying to grab low hanging fruit every morning or so.

FBruzzesi:
For the validation steps, I am waiting for [`sklearn-compat`](https://sklearn-compat.readthedocs.io/en/latest/) to be out for release candidate version. I guess you might have more insights for when that would happen :D

koaning:
I was not aware of this at all actually 😅 will poke around.

PR #725, opened 2024-12-08
chore: class inheritance order

FBruzzesi:
# Description

Follows #724 and #723 to fix order of inheritance.


koaning (approved):
Nice. Eager to get these in.

PR #726, opened 2024-12-16
patch: scikit-learn 1.6 compatibility

FBruzzesi:
# Description

Overseeds #720 by using and adapting [sklearn-compat](https://github.com/sklearn-compat/sklearn-compat)

Tested on python 3.10 and 3.12 with scikit-learn 1.5 and 1.6: all tests are passing these local runs

## Tip for review

I tried to carefully commit each module at the time, hence the best way forward would be to review one commit at the time.

~~With a big disclaimer, at the end I realized that we `validate_data` was not gonna work in all situations, but an adaptation of `check_X_y` would, therefore I rolled back those entirely in another branch which was merged into this one~~

Edit: as suggested, using `validate_data` is the right way to go, and after some deeper dive, I was able to refactor the codebase using such function

### DIFF
--- a/sklego/_sklearn_compat.py
+++ b/sklego/_sklearn_compat.py
@@ -1,0 +1,520 @@
+"""Ease developer experience to support multiple versions of scikit-learn.
+
+This file is intended to be vendored in your project if you do not want to depend on
+`sklearn-compat` as a package. Then, you can import directly from this file.
+
+Be aware that depending on `sklearn-compat` does not add any additional dependencies:
+we are only depending on `scikit-learn`.
+
+Version: 0.1.1
+"""
+
+from __future__ import annotations
+
+import inspect
+import sys
+from dataclasses import dataclass, field
+from typing import Callable, Literal
+
+import sklearn
+from sklearn.utils.fixes import parse_version
+
+sklearn_version = parse_version(parse_version(sklearn.__version__).base_version)
+
+
+########################################################################################
+# The following code does not depend on the sklearn version
+########################################################################################
+
+
+# tags infrastructure
+def _dataclass_args():
+    if sys.version_info < (3, 10):
+        return {}
+    return {"slots": True}
+
+
+def get_tags(estimator):
+    """Get estimator tags in a consistent format across different sklearn versions.
+
+    This function provides compatibility between sklearn versions before and after 1.6.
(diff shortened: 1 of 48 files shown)


sklego/_sklearn_compat.py
```diff
@@ -0,0 +1,572 @@
...
+            ensure_2d=ensure_2d,
+            allow_nd=allow_nd,
+            ensure_min_samples=ensure_min_samples,
+            ensure_min_features=ensure_min_features,
+            estimator=estimator,
+            input_name=input_name,
+            **kwargs,
+        )
+
+    def check_X_y(
```
FBruzzesi, line 220: This is not part of sklearn-compat as of now, and that's why I am using the code snippet instead of depending on it. I will open an issue upstream

sklego/decomposition/pca_reconstruction.py
```diff
@@ -95,6 +97,7 @@ def fit(self, X, y=None):
             If `threshold` is `None`.
         """
         X = check_array(X, estimator=self, dtype=FLOAT_DTYPES)
+        _check_n_features(self, X, reset=True)
```
adrinjalali, line 100: you probably want to call `validate_data` instead of `check_array`, `check_X_y`, and `_check_n_features`. In sklearn itself, we rarely call these functions. It's almost always `validate_data`

FBruzzesi: Thanks @adrinjalali ! I will have a deeper look.

If you look at the first 7 commits of this PR, that was exactly how I approached it - i.e. moving from `check_array` and `check_X_y` to `validate_data`. However, many tests were failing (especially for meta estimators), thus I move forward with the safe path of adjusting what we already have.

adrinjalali: I'd be happy to have a look if you point me to some errors.

FBruzzesi: It seems it was a skill issue of mine! Should be fixed now 😇

koaning (approved):
Not the biggest fan of a big vendor, but I guess it is the most pragmatic way to go about it for now. I guess time will tell if this becomes a burden but I guess by that time we can just assume folks use scikit-learn >> 1.6.

@FBruzzesi well done! This was a lot of work!

PR #730, opened 2025-01-17
feat: raise value error for shape mismatch

FBruzzesi:
# Description

I was the one causing the issue in [narwhals](https://github.com/narwhals-dev/narwhals/pull/1814): we now raise a dedicated error for length mismatch in `with_columns`, while scikit-learn expects a `ValueError` instead. I am adding a preventive check to comply with that.


koaning (approved):
Seems fine on my end!

PR #731, opened 2025-01-17
feat: raise value error for shape mismatch on `HierarchicalPredictor`

FBruzzesi:
# Description

Same as #730, sorry I forgot to add it here as well


koaning (approved):
LGTM

Issue #596, opened 2023-11-09
[DOCS] Example usage in docstring


FBruzzesi:
As [discussed in #586](https://github.com/koaning/scikit-lego/pull/586#discussion_r1385475426), although most of the library features are documented in the user guide, the best way to showcase how to access and use each class/function would be a minimal example usage in the docstrings. And this is currently covered only for a subset of the library features.

Here the list of all the remaining classes that would benefit from it:

- [ ] `linear_model.LowessRegression`
- [ ] `linear_model.ProbWeightRegression`
- [ ] `linear_model.DeadZoneRegressor`
- [ ] `linear_model.DemographicParityClassifier`
- [ ] `linear_model.EqualOpportunityClassifier`
- [ ] `model_selection.TimeGapSplit`
- [ ] `model_selection.GroupTimeSeriesSplit`
- [ ] `model_selection.KlusterFoldValidation`
- [ ] `naive_bayes.GaussianMixtureNB`
- [ ] `naive_bayes.BayesianGaussianMixtureNB`
- [ ] `neighbors.BayesianKernelDensityClassifier`
- [ ] `decomposition.pca_reconstruction.PCAOutlierDetection`
- [ ] `decomposition.umap_reconstruction.UMAPOutlierDetection`
- [ ] `meta.confusion_balancer.ConfusionBalancer`
- [ ] `meta.estimator_transformer.EstimatorTransformer`
- [ ] `meta.grouped_predictor.GroupedPredictor`
- [ ] `meta.grouped_transformer.GroupedTransformer`
- [x] `meta.outlier_classifier.OutlierClassifier`
- [ ] `meta.regression_outlier_detector.RegressionOutlierDetector`
- [ ] `meta.subjective_classifier.SubjectiveClassifier`
- [ ] `meta.thresholder.Thresholder`
- [ ] `mixture.bayesian_gmm_classifier.BayesianGMMClassifier`
- [ ] `mixture.bayesian_gmm_detector.BayesianGMMOutlierDetector`
- [ ] `mixture.gmm_classifier.GMMClassifier`
- [ ] `mixture.gmm_detector.GMMOutlierDetector`
- [x] `preprocessing.dictmapper.DictMapper`
- [ ] `preprocessing.formulaictransformer.FormulaicTransformer`
- [ ] `preprocessing.identitytransformer.IdentityTransformer`
- [ ] `preprocessing.intervalencoder.IntervalEncoder`
[...]

PR #732, opened 2025-01-22
docs: `model_selection` module docstrings

mkalimeri:
# Description

Related to #596

Examples added for

- model_selection.TimeGapSplit
- model_selection.GroupTimeSeriesSplit
- model_selection.ClusterFoldValidation (I see that KlusterFoldValidation has been renamed as ClusterFoldValidation)

## Type of change

## Checklist:

- [NA] I have made corresponding changes to the documentation (also to the readme.md)
- [NA] I have added tests that prove my fix is effective or that my feature works
- [NA] I have added tests to check whether the new feature adheres to the sklearn convention

### DIFF
--- a/sklego/model_selection.py
+++ b/sklego/model_selection.py
@@ -95,24 +95,25 @@ class TimeGapSplit:
 
     # Create dataset
     np.random.seed(1)
-    num_rows = 30
+    num_rows = 50
     df = pd.DataFrame(np.random.randn(num_rows, 4)).rename(columns={0: 'c1', 1: 'c2', 2: 'c3', 3: 'c4'})
-    df['date'] = pd.date_range("2024-01-01", periods=num_rows, freq="s")
+    df['date'] = pd.date_range("2024-01-01", periods=num_rows, freq="h")
 
     # Define parameters
-    td = timedelta(seconds=10)
-    vd = timedelta(seconds=5)
-    gd = timedelta(seconds=2)
+    td = timedelta(hours=15)
+    vd = timedelta(hours=6)
+    gd = timedelta(hours=4)
 
     tgs = TimeGapSplit(date_serie=df['date'], train_duration=td, valid_duration=vd, gap_duration=gd)
     # Print the summary of the first fold
     print(tgs.summary(df).iloc[0])
 
     ### Start date     2024-01-01 00:00:00
-    ### End date       2024-01-01 00:00:09
-    ### Period             0 days 00:00:09
-    ### Unique days                     10
-    ### nbr samples                     10
+    ### End date       2024-01-01 14:00:00
+    ### Period             0 days 14:00:00
+    ### frequency                        h
+    ### Unique days                      1
+    ### nbr samples                     15
     ### Name: (0, train), dtype: object
 
     # Generate the folds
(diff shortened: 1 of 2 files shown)


sklego/model_selection.py
```diff
@@ -84,6 +84,48 @@ class TimeGapSplit:
...
+    gd = timedelta(seconds=2)
+
+    tgs = TimeGapSplit(date_serie=df['date'], train_duration=td, valid_duration=vd, gap_duration=gd)
+    # Print the summary of the first fold
+    print(tgs.summary(df).iloc[0])
+
+    ### Start date     2024-01-01 00:00:00
+    ### End date       2024-01-01 00:00:09
+    ### Period             0 days 00:00:09
+    ### Unique days                     10
```
FBruzzesi, line 114: How is this returning 10 days? 🤔

Edit: Looking into the code, this is fairly random assumption that the number of unique dates are in days.

sklego/model_selection.py
```diff
@@ -84,6 +84,48 @@ class TimeGapSplit:
...
+    ```py
+    from datetime import timedelta
+    import numpy as np
+    import pandas as pd
+    from sklego.model_selection import TimeGapSplit
+
+    # Create dataset
+    np.random.seed(1)
+    num_rows = 30
+    df = pd.DataFrame(np.random.randn(num_rows, 4)).rename(columns={0: 'c1', 1: 'c2', 2: 'c3', 3: 'c4'})
```
FBruzzesi, line 99: You can pass the column names directly:

```suggestion
    df = pd.DataFrame(np.random.randn(num_rows, 4), columns=["c1", "c2", "c3", "c4"])
```

FBruzzesi (commented):
Thanks for the PR @mkalimeri!

The only concern I have is related to the part of code in `TimeGapSplit` summary. @koaning WDYT? 
We can keep it as a follow up, but that's certainly not the right assumption to make in this context

sklego/model_selection.py
mkalimeri: Yes, this should not be unique days but maybe timepoints? Shall we leave this as is for now and open another issue to generate additional/different information in the summary? Happy to pick that up!

koaning: Yeah, this is a nit that would be nice to tackle. I would also totally be open to doing that in this PR, less waiting for merges that way.

mkalimeri: Sure, I can do that. I will come back to you with a plan of what information I would include. Updating this would require changes in tests as well, so it will be more than one file. Is that ok?

mkalimeri: Hi, I got to take a look at this today. We can keep the number of unique days in the summary, but as @FBruzzesi mentioned, we can't assume that the number of days is relevant to all use cases. We could keep the field and correct the calculation, to give an idea of the frequency of data, but it might not be relevant if the granularity of the time series is lower. Another idea is to add the granularity of the time series as a field of the summary. What are your thoughts?

mkalimeri:
I updated the code based on the feedback from the previous review. I kept the summary field 'num of days' and updated the calculation to reflect the actual number of days. I also added a field 'frequency', to inform about the frequency of the data. I updated example/test to reflect these changes. Let me know if you can think of other informative fields!

koaning (approved):
I am approving, partially to trigger the unit tests

FBruzzesi:
@koaning Narwhals CI failed after merging this one (see [scikit-lego job](https://github.com/narwhals-dev/narwhals/actions/runs/30612397190/job/91097902617?pr=3826))

```terminal
FAILED tests/test_model_selection/test_timegapsplit.py::test_timegapsplit_summary - AssertionError: DataFrame.iloc[:, 3] (column name="frequency") are different

DataFrame.iloc[:, 3] (column name="frequency") values are different (100.0 %)
[index]: [(0, train), (0, valid), (1, train), (1, valid), (2, train), (2, valid), (3, train), (3, valid), (4, train), (4, valid), (5, train), (5, valid)]
[left]:  [24h, 24h, 24h, 24h, 24h, 24h, 24h, 24h, 24h, 24h, 24h, 24h]
[right]: [D, D, D, D, D, D, D, D, D, D, D, D]
At positional index 0, first diff: 24h != D
```

koaning:
GAh! And here I was trying to save time 😅

FBruzzesi:
I just had some time to double check what's happening:

1. As we have and use uv.lock, our dependencies are not getting updated. That's _kind of_ an issue as we are not testing with latest
2. In this specific case: we derive the frequency column of `TimeGapSplit.summary()` via `pd.tseries.frequencies.to_offset(delta).freqstr`. In pandas 3.0 the Day offset became a calendar day (a non-fixed duration, DST-aware). An exact `Timedelta` is a fixed duration, so `to_offset` can no longer map it onto `Day` and falls back to hours:

delta | pandas 2.3.3 | pandas 3.0.5
-- | -- | --
1 day | D | 24h
2 days | 2D | 48h
7 days | 7D | 168h
36 hours | 36h | 36h
90 minutes | 90min | 90min

3. If I try to first bump all version in the uv lock `uv sync --all-extras --all-groups --upgrade`, we get a bunch more tests failing.
4. I can try to make some time next week to fix them :)

PR #760, opened 2025-11-11
tests: Skip more `RandomAdder` and `RandomRegressor` compatibility checks

FBruzzesi:
# Description

Started to fail since pytest 9.0


FBruzzesi:
@koaning FYI there is still one workflow leftover from the old "batch" that appears as mandatory

PR #784, opened 2026-04-16
fix: Add `date_series` as `TimeGapSplit` parameter and deprecate `date_serie`

### LINKED ISSUE #761: [BUG] Typo in method parameter
In order to help us fix a potential bug we need: 

In the `TimeGapSplit` cross validation, the first parameter is `date_serie`, but that should be `date_series=df["date"],`. I believe this is a typo because the [docstring here](https://github.com/koaning/scikit-lego/blob/2472152e3adadac901737c0f4f7be3fde10f7159/sklego/model_selection.py#L127) says "series"

[Update here](https://github.com/koaning/scikit-lego/blob/2472152e3adadac901737c0f4f7be3fde10f7159/sklego/model_selection.py#L114) (and other lines)

I should note that it's less a "bug" because it doesn't break anything and more an "error"

- [x] a clear and concise description of what the bug is.
- [ ] code that helps us reproduce this 
- [ ] **optional** a unit test that helps us catch the bug in the future

RedZapdos123:
## Description:

Issue #761 reported a typo in the `TimeGapSplit` constructor argument name, where `date_serie` was used instead of `date_series`.

This PR adds a backward compatible fix:

- Supports the correct argument name `date_series`
- Keeps support for the legacy name `date_serie`
- Validates that exactly one of `date_series` or `date_serie` is provided
- Keeps internal compatibility by aliasing `date_serie` to `date_series`
- Updates `TimeGapSplit` docs examples to use `date_series`

It also adds regression tests that cover:

- `date_series` works correctly
- Passing both `date_series` and `date_serie` raises a clear `ValueError`

Closes #761.

## Validation:

- `ruff check sklego/model_selection.py tests/test_model_selection/test_timegapsplit.py`
- `pytest -q tests/test_model_selection/test_timegapsplit.py`
- `pytest -q`

Note: `pytest -q` reports one unrelated pre-existing failure in `tests/test_meta/test_grouped_predictor.py::test_custom_shrinkage`. (See #785 )

## Type of change

## Checklist

- [x] I have commented my code, particularly in hard to understand areas

#### The validation screenshots of the tests run, locally on WSL:

image

Screenshot 2026-04-16 223559

### DIFF
--- a/sklego/model_selection.py
+++ b/sklego/model_selection.py
@@ -50,8 +50,6 @@ class TimeGapSplit:
         Series with the date, that should have all the indices of X used in the split() method.
         If the Series is not pandas-like (for example, if it's a Polars Series, which does not have
         an index) then it must the same length as the `X` and `y` objects passed to `split`.
-    date_serie : Series | None, default=None
-        Backward-compatible alias for `date_series`.
     valid_duration : datetime.timedelta
         Retraining period.
     stride_duration : datetime.timedelta | None
@@ -69,6 +67,8 @@ class TimeGapSplit:
 
         - `"rolling"` window has fixed size and is shifted entirely.
         - `"expanding"` left side of window is fixed, right border increases each fold.
+    date_serie : Series | None, default=None
+        Backward-compatible alias for `date_series`.
 
     Notes
     -----
@@ -88,26 +88,34 @@ class TimeGapSplit:
 
     def __init__(
         self,
-        date_serie=None,
+        date_series=None,
         valid_duration=None,
         stride_duration=None,
         train_duration=None,
         gap_duration=timedelta(0),
         n_splits=None,
         window="rolling",
-        date_series=None,
+        date_serie=None,
     ):
         if valid_duration is None:
             raise ValueError("`valid_duration` has to be defined")
 
-        if (date_serie is None) and (date_series is None):
-            raise ValueError("Either `date_series` or `date_serie` has to be defined")
-
-        if (date_serie is not None) and (date_series is not None):
-            raise ValueError("Only one of `date_series` and `date_serie` can be provided")
-
-        if date_series is not None:
-            date_serie = date_series
+        match (date_series, date_serie):
+            case (None, None):
+                msg = "`date_series` cannot be None"
+                raise ValueError(msg)
+            case (value, None):
+                self.date_series = nw.from_native(value, series_only=True).alias("__date__")
+            case (None, value):
+                msg = (
+                    "Please use `date_series` instead of `date_serie`, "
+                    "`date_serie` will be deprecated in future versions"
+                )
+                warn(msg, DeprecationWarning)
+                self.date_series = nw.from_native(value, series_only=True).alias("__date__")
+            case (_, _):
+                msg = "Cannot provide both `date_series` and `date_serie`"
+                raise ValueError(msg)
 
         # If stride length is not defined, set it equal to the length validation set
         if stride_duration is None:
@@ -126,7 +134,6 @@ class TimeGapSplit:
         if (train_duration is not None) and (train_duration <= stride_duration):
             raise ValueError("stride_duration is longer than train_duration, it should be shorter.")
 
-        self.date_series = nw.from_native(date_serie, series_only=True).alias("__date__")
         self.date_serie = self.date_series
         self.train_duration = train_duration
         self.valid_duration = valid_duration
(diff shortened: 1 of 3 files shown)


RedZapdos123:
@FBruzzesi I request a review for this PR (bug fix).

sklego/model_selection.py
```diff
@@ -46,10 +46,12 @@ class TimeGapSplit:
 
     Parameters
     ----------
-    date_serie : Series
+    date_series : Series
         Series with the date, that should have all the indices of X used in the split() method.
         If the Series is not pandas-like (for example, if it's a Polars Series, which does not have
         an index) then it must the same length as the `X` and `y` objects passed to `split`.
+    date_serie : Series | None, default=None
+        Backward-compatible alias for `date_series`.
```
FBruzzesi: I would move that to the end of the parameters list

sklego/model_selection.py
```diff
@@ -86,14 +88,27 @@ class TimeGapSplit:
...
-        date_serie,
-        valid_duration,
+        date_serie=None,
+        valid_duration=None,
         stride_duration=None,
         train_duration=None,
         gap_duration=timedelta(0),
         n_splits=None,
         window="rolling",
+        date_series=None,
```
FBruzzesi: Since we are changing it, let's re-order the parameters so that `date_series` appears first:

```suggestion
        date_series=None,
        valid_duration=None,
		stride_duration=None,
		train_duration=None,
		gap_duration=timedelta(0),
		n_splits=None,
		window="rolling",
        date_serie=None,
```

sklego/model_selection.py
```diff
@@ -86,14 +88,27 @@ class TimeGapSplit:
...
+            raise ValueError("`valid_duration` has to be defined")
+
+        if (date_serie is None) and (date_series is None):
+            raise ValueError("Either `date_series` or `date_serie` has to be defined")
+
+        if (date_serie is not None) and (date_series is not None):
+            raise ValueError("Only one of `date_series` and `date_serie` can be provided")
+
+        if date_series is not None:
+            date_serie = date_series
```
FBruzzesi: I think this might be a good case to use `match`.

In this way we have case by case control and we can raise better errors and/or custom warnings.

For example:

* if they are both None, we should not mention the existence of `date_serie`, and only mention `date_series`
* In the case of `date_series=None` and `date_serie not None`, we should move forward but raise a DeprecationWarning. 

Here is the suggestion:

```suggestion
        match (date_series, date_serie):
            case (None, None):
                msg = "`date_series` cannot be None"
                raise ValueError(msg)

            case (value, None):
                self.date_series = nw.from_native(value, series_only=True).alias("__date__")

            case (None, value):
                msg = (
                    "Please use `date_series` instead of `date_serie` arguments. "
                    "`date_serie` will be deprecated in future versions"
                )
                warn(msg, DeprecationWarning)

                self.date_series = nw.from_native(value, series_only=True).alias("__date__")

            case (_, _):
                msg = "Cannot provide both `date_series` and `date_serie`"
                raise ValueError(msg)
```

Then line 129 can be removed

---

See [tutorial](https://benhoyt.com/writings/python-pattern-matching/) (one of many) for more in depth explanation of Structural pattern matching

FBruzzesi (commented):
Hey @RedZapdos123 👋🏼 thanks for your contribution. I added a couple of comments/suggestions to hopefully have more insightful messages for users

sklego/model_selection.py
RedZapdos123: Moved the alias parameter to the end.

sklego/model_selection.py
RedZapdos123: The constructor parameters are now reordered so that date_series is teh first, and date_serie is the last.

sklego/model_selection.py
RedZapdos123: Thank you for the advice. I refactored this to the structural pattern matching with case handling:

- Both missing: raises error mentioning only date_series.
- Current only: proceeds normally.
- Legacy only: proceeds with DeprecationWarning.
- Both provided: raises conflict error.

I also removed the now redundant downstream check, and added focused regression coverage for warning and error paths.

FBruzzesi: Thanks for addressing it!

FBruzzesi (approved):
Thanks @RedZapdos123 for the contribution and quick adjustment! It's a +1 from me 😉

PR #786, opened 2026-04-16
fix(test): Do not compare floats directly

### LINKED ISSUE #785: test_custom_shrinkage fails on the clean main due to strict float equality
`tests/test_meta/test_grouped_predictor.py::test_custom_shrinkage` fails on a clean clone of `main` due to strict equality comparison between floating point values that differ only at machine precision.

### Reproduciblity:

- Environment: WSL (Linux), Python 3.11.15, pytest 9.0.3

### Minimal Reproducible Example (MRE):

Run only the failing test:

```bash
pytest -q tests/test_meta/test_grouped_predictor.py::test_custom_shrinkage -vv
```

Failing assertion (from clean run):

```text
E       AssertionError: assert [1.8333333333...6666666666665] == [1.8333333333...6666666666665]
E       At index 0 diff: 1.833333333333333 != 1.8333333333333333
```

The test location is:

- `tests/test_meta/test_grouped_predictor.py` line ~370
- Current assertion: `assert expected_prediction == shrink_est.predict(X).tolist()`

### Expected behavior:

The test should pass when predictions are numerically equivalent within floating point tolerance.

### Actual behavior:

The test fails because it requires exact list equality on floating point outputs.

### Why this seems brittle?

This appears to be a precision artifact rather than a semantic model error. A tolerance based comparison (for example `np.testing.assert_allclose`) would likely make this test robust across environments.

### Additional context:

On the same test run, full test execution also reports this as the only failure:

```bash
pytest -q
# 1 failed, 6756 passed, 190 skipped, 3641 warnings
# FAILED tests/test_meta/test_grouped_predictor.py::test_custom_shrinkage
```

#### The screenshot of the test failure:

Image

FBruzzesi:
# Description

Closes #785

I couldn't replicate it, but that's a very good point. We should not compare floats directly. This PR moves the comparisons to `np.allclose`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
     -   id: check-merge-conflict
     -   id: check-added-large-files
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.8
+    rev: v0.15.11
     hooks:
     -   id: ruff  # Run the linter.
         args: [--fix, sklego, tests]
--- a/tests/test_meta/test_grouped_predictor.py
+++ b/tests/test_meta/test_grouped_predictor.py
@@ -266,9 +266,7 @@ def test_constant_shrinkage(shrinkage_data):
         np.array([means["Earth"], means["BE"], means["Antwerp"]]) @ shrinkage_factors,
         np.array([means["Earth"], means["BE"], means["Brussels"]]) @ shrinkage_factors,
     ]
-
-    for exp, pred in zip(expected_prediction, shrink_est.predict(X).tolist()):
-        assert pytest.approx(exp) == pred
+    assert np.allclose(shrink_est.predict(X), expected_prediction)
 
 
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ repos:
     -   id: check-merge-conflict
     -   id: check-added-large-files
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.8
+    rev: v0.15.11
```
koaning, line 12: Why was the ruff version incremented?

koaning (approved):
Just one oddity, agree with the fix.

.pre-commit-config.yaml
FBruzzesi: I sporadically run `pre-commit autoupdate` 😇

koaning (approved):

koaning:
@copilot resolve the merge conflicts in this pull request

koaning (approved):

PR #790, opened 2026-05-03
deps: Bump jupyterlab, nbconvert, notebook due to sec vulnerabilities

FBruzzesi:
# Description

As per title, bump dependencies due to recent security vulnerabilities

Closes #787, #788, #789, #792, #793

Additionally adds uv-lock pre-commit hook to lower effort to remember of sync-ing


FBruzzesi:
@koaning there is no day that goes by without new security vulnerability these days! 😭

PR #805, opened 2026-07-05
Clarify equal opportunity warning message

DresdenGman:
Clarifies the warning emitted by `equal_opportunity_score` so it includes both the `y_true` and `y_hat` positive-target conditions that trigger the zero-return path.

This only updates the warning text and its regression coverage; it does not change the metric logic.


koaning (approved):

koaning:
Will fix the test in another PR. 

Thanks!

PR #812, opened 2026-07-24
feat: Enable `EstimatorTransformer.get_feature_names_out` via `ClassNamePrefixFeaturesOutMixin` mixin

### LINKED ISSUE #533: [BUG]  'EstimatorTransformer' object has no attribute 'get_feature_names_out'

DMZ22:
# Description

`EstimatorTransformer` inherits `TransformerMixin`, which does not supply `get_feature_names_out`, so the method simply isn't there:

```python
>>> EstimatorTransformer(LinearRegression()).get_feature_names_out(None)
AttributeError: 'EstimatorTransformer' object has no attribute 'get_feature_names_out'
```

The practical symptom is the pipeline case, which fails even after fitting:

```python
>>> make_pipeline(StandardScaler(), EstimatorTransformer(LinearRegression())).fit(X, y).get_feature_names_out()
AttributeError: Estimator estimatortransformer does not provide get_feature_names_out.
```

The transformer's output is **not** one-to-one with the input columns — `transform` returns one column per target for a multi-output estimator and reshapes to a single column otherwise — so `OneToOneFeatureMixin` would be wrong here. The sklearn convention for this shape is `ClassNamePrefixFeaturesOutMixin` plus a `_n_features_out`, which is what this adds; `n_outputs_` is recorded during `fit`.

```python
single output   -> ['estimatortransformer0']
two targets     -> ['estimatortransformer0', 'estimatortransformer1']
in a pipeline   -> ['estimatortransformer0']
before fit      -> NotFittedError   (rather than AttributeError)
```

In every case the number of names equals `transform(X).shape[1]`, which is asserted in the tests rather than assumed. A side benefit is that `set_output(transform="pandas")` now yields a properly-named DataFrame instead of failing.

@koaning — this is the change you green-lit on #533 back in 2022 ("Yes please!"); @CarloLepelaars offered then but no PR followed, and nothing open touches this file, so I hope it's helpful rather than stepping on anyone.

Fixes #533

## Type of change

## Checklist:

- [x] My code follows the style guidelines (ruff) — `ruff check` and `ruff format --check` both clean
- [ ] I have made corresponding changes to the documentation — the method is inherited API, so nothing in the docs asserts its absence; happy to add a line if you'd like one
- [x] I have added tests to check whether the new feature adheres to the sklearn convention — the existing `parametrize_with_checks` suite still passes, and the names/width equality is asserted

`tests/test_meta/` goes from **1073 passed** to **1110 passed** with the same 15 pre-existing collection errors (missing `pyarrow` locally). Reverting only `estimator_transformer.py` while keeping the tests fails 37 of them, so they do guard the change.

---

One thing I found while measuring the output width, which I have **not** touched because it is a separate bug and a behaviour change:

`transform` does `output.reshape(-1, 1)` whenever `multi_output_` is `False`. With `predict_func="predict_proba"` and a 1-D `y`, `predict_proba` returns `(n_samples, n_classes)`, so the reshape multiplies the rows:

```python
X has 60 rows, 3-class y
  estimator_.predict_proba(X) -> (60, 3)
  transform(X)                -> (180, 1)      # 60 samples in, 180 rows out
```

The class docstring records this too — its example is 150 samples with `predict_func="predict_proba"` and states `Shape of transformed data: (300, 1)`. Happy to open a separate issue, or a PR, if you'd like it fixed.


CarloLepelaars:
Appreciate you picking this up. I forgot about it

FBruzzesi (commented):
Thanks for the contribution @DMZ22 - I think the changes are totally fine.

The only issue I see is that `ClassNamePrefixFeaturesOutMixin` was introduced in scikit-learn 1.2, and we currently allow scikit-learn 1.0+ . @koaning what's your take in bumping the minimum scikit-learn version dependency?

koaning:
I've been behind on reviews, sorry about that. 

I think I am cool with the version bump. It's been a few years at this point and the main thing that's important is that it doesn't increase the maintainance burden for us.

koaning (approved):

